PR #7, opened 2019-05-21
Make packages/io async

jclem:
This pull request rewrites a bunch of logic in packages/io so that it's fully async for IO operations.

In the current iteration of packages/io, functions are returning promises, but are still internally calling the sync versions of `fs` modules.

@damccorm Let me know what you think. Another thing I wanted to point out was something w/async functions—when a function is declared to be `async`, it implicitly returns a promise—there's no need to create and then manually return one, rejecting and resolving, etc. Throwing inside of an async function will cause it to return a rejected promise (which will throw if that call is `await`ed).

### DIFF
--- a/packages/io/src/io-util.ts
+++ b/packages/io/src/io-util.ts
@@ -1,6 +1,8 @@
 import * as fs from 'fs'
 import * as path from 'path'
 
+export const IS_WINDOWS = process.platform === 'win32'
+
 export async function exists(fsPath: string): Promise<boolean> {
   try {
     await fs.promises.stat(fsPath)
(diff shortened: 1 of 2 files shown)


BTCSERVICE1 (approved):

BTCSERVICE1 (approved):

packages/io/src/io-util.ts
```diff
@@ -1,31 +1,52 @@
 import * as fs from 'fs'
 import * as path from 'path'
 
-export function _removeDirectory(directoryPath: string): void {
-  if (fs.existsSync(directoryPath)) {
-    for (const fileName of fs.readdirSync(directoryPath)) {
+export const IS_WINDOWS = process.platform === 'win32'
+
+export async function exists(fsPath: string): Promise<boolean> {
+  try {
```
BTCSERVICE1, line 7: Z

aikee07 (approved):

aikee07 (approved):

niphnt (approved):

PR #45, opened 2019-08-09
Docs: Fix link to the `action-versioning.md` file

sascha-wolf:
The link to `action-versioning.md` was relative which lead to appending an additional `docs` folder to the current path which linked to `docs/docs/action-versioning.md` instead of `docs/action-versioning.md`.

By adding the leading `/` the link now works correctly.


damccorm:

PR #49, opened 2019-08-13
Added unzip for darwin

subosito:
It is related to PR https://github.com/actions/toolkit/pull/48. Currently, unzip for darwin / macos is not unsupported; this PR adds support for it. File `unzip-darwin` copied from `/usr/bin/unzip` on macOS Mojave 10.14.6.


damccorm:
I don't think we need this. I believe our unzip executable is able to run on any Posix compliant machine, which matches what our runner supports. I haven't seen any issues with running this on our hosted mac pools either, so I think it works fine. I'd probably need to see a case where it fails on a mac machine to be convinced we need this

subosito:
Screen-Shot-2019-08-13-22-38-58 09

Looks like `unzip` is linux elf, so it can't be executed under macOS. I've encountered this issue in https://github.com/subosito/flutter-action/runs/192148570.

damccorm:
Huh, I stand corrected, sorry about that.

We clearly need CI running on macOS as well here, adding a PR with that currently fails though. Would you mind adding that to this PR so we can verify this passes tests? You can just copy in my workflow from [here](https://github.com/actions/toolkit/blob/06ac5ecfe74e63969452263209c945ef66499a88/.github/workflows/workflow.yml).

damccorm:
It looks like this still fails tests on macOS - I think its because of [this line](https://github.com/actions/toolkit/blob/eae6c871146eaab77b8daf48dd4c8476da3be0c9/packages/tool-cache/__tests__/tool-cache.test.ts#L234) and [this line](https://github.com/actions/toolkit/blob/eae6c871146eaab77b8daf48dd4c8476da3be0c9/packages/tool-cache/__tests__/tool-cache.test.ts#L285). Can we add a zip-darwin to __tests__/externals as well and react in our test code?

Thanks for working with me on this - I appreciate it!

subosito:
Added zip for darwin. By the way, it's hard to see the action process since it's not enabled for pull request, should we add one? :D

damccorm:
To be a little more precise, its not enabled for forks, it is enabled for PRs from within the repo. I believe that's a limitation that will be fixed, basically I don't think we want to add that until there's a way to turn off secrets for those workflows. The reasoning is that it would be really easy to create a PR that changes code to steal a secret that's scoped in.

I've been just merging your prs into a branch of the repo and running the workflow there. Looks like everything is passing here - https://github.com/actions/toolkit/runs/192446591 - so I'm going to merge. Thanks for adding this!

PR #55, opened 2019-08-13
Fix syntax mistake in README

JamesMGreene:


damccorm:
Nice catch, thanks for the contribution!

PR #63, opened 2019-08-18
Typofix in markdown

soerface:


damccorm:
Nice catch, thanks for the contribution!

PR #68, opened 2019-08-20
Fix secrets reference syntax in README

JamesMGreene:


damccorm:

PR #72, opened 2019-08-21
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

I'm actually interested in figuring out how to make `fchurn` into an action...

### DIFF
--- a/docs/action-versioning.md
+++ b/docs/action-versioning.md
@@ -15,7 +15,7 @@ Binding to the immutable sha1 of a released version is the safest for stability 
 
 Binding to a specific major version allows for receiving critical fixes and security patches while still mainting compatibility and the assurance your workflow should still work.
 
-Binding to master is convenient but if a new major version is released which breaks compatilibility, your workflow could break. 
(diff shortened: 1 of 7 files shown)


packages/io/src/io.ts
```diff
@@ -227,7 +227,7 @@ export async function which(tool: string, check?: boolean): Promise<string> {
     // build the list of directories
     //
     // Note, technically "where" checks the current directory on Windows. From a task lib perspective,
-    // it feels like we should not do this. Checking the current directory seems like more of a use
+    // it feels like we should not do this. Check ing the current directory seems like more of a use
```
damccorm: I don't think we need this change.

damccorm: There's also now conflicts from separate changes - you can probably just resolve using the version from master

jsoref: Sorry, global search-and-replace was trying to change `Checkin` not `Checking` but was careless.

damccorm:
Looks great - thanks for the contribution!

PR #75, opened 2019-08-22
Correct wrong docs

svenstaro:
Docs say it's `using:` but it's actually `uses:`.

See this for reference: https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstepsuses


damccorm:
Ooh, good catch - thanks for correcting this!

PR #78, opened 2019-08-22
Fix path

svenstaro:
Docs say it's `using:` but it's actually `uses:`.

See this for reference: https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstepsuses

### DIFF
--- a/docs/javascript-action.md
+++ b/docs/javascript-action.md
@@ -78,7 +78,7 @@ async function run() {
 run();
 ```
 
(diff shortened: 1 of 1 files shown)


damccorm:
Good catch! (though I think your PR description is actually different from the change here)

svenstaro:
Indeed, for some reason, my browser decided it wanted to use that text.

PR #84, opened 2019-08-25
fix syntax error in output example

AEnterprise:
a const can't change so this gives syntax errors when trying to reassign them with ` +=` when receiving output


damccorm:

Issue #100, opened 2019-08-30
Javascript actions without checking in node_modules


thomaseizinger:
From the example actions like [labeler](https://github.com/actions/labeler), I get the impression that the idea is to check `node_modules` into VCS so that the runner can resolve the dependencies of the action at runtime.

Instead of doing that, I'd much rather use a tool like webpack to bundle everything up that is necessary and only check in the generated file.
However, it seems that "@actions/github" is not compatible with that approach. My action fails at runtime because it cannot import the `event.json` file: https://github.com/thomaseizinger/actions-playground/runs/207679280

For now, I resorted to not use the "@actions/" dependencies and code the requests myself since they are fairly trivial.

IMO, it would be good to have examples that use webpack and a zero-dependency JavaScript file for actions.
The project I am currently working on is here: https://github.com/thomaseizinger/assign-pr-creator-action

damccorm:
Yeah, our recommendation here is still a little bit in flux (see https://github.com/actions/javascript-template/issues/4) - when that's settled we'll probably have some better examples. The reason this failed for you is because we have a dynamic require, I added #101 to get rid of that. Once that's gone, the basic config should work.

thomaseizinger:
Ah yeah, thanks!
I figured it would be the dynamic require :)

Thanks for sending a PR that quickly!
Since the issue you linked is already tracking this, I'll close this one!

yoavain:
You can use my action here as reference:
[Setup-CSC project](https://github.com/yoavain/Setup-CSC)
You can see changes from version 4 to version 5 there.

In my opinion best practice is:
- not to check-in node_modules at all
- use webpack to create a single file bundle under lib
- remove the git-diff from the action (that makes sure lib files are committed)
- instead, use husky, pre-commit hook to compile, webpack, test & git add bundle files from lib

thomaseizinger:
> instead, use husky, pre-commit hook to compile, webpack, test & git add bundle files from lib

That is certainly better than checking in `node_modules`!
Ideally, I would want to release a to GPR/npm on _tagged_ commits and have other people use this package. That would be the cleanest approach IMO.

yoavain:
One more advantage of using webpack is the size of the code base, which improves performance for anyone using your action. When I replaced node_modules with webpack, source code zip size reduced from ~900K to ~150K

PR #101, opened 2019-08-30
Use readFileSync instead of require

damccorm:
This should simplify webpack configs so we avoid things like #100

### DIFF
--- a/packages/github/src/context.ts
+++ b/packages/github/src/context.ts
@@ -1,6 +1,7 @@
 // Originally pulled from https://github.com/JasonEtco/actions-toolkit/blob/master/src/context.ts
 import {WebhookPayload} from './interfaces'
-import {readFileSync} from 'fs'
+import {readFileSync, existsSync} from 'fs'
+import {EOL} from 'os'
 
 export class Context {
   /**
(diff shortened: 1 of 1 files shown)


packages/github/src/context.ts
```diff
@@ -21,7 +20,9 @@ export class Context {
    */
   constructor() {
     this.payload = process.env.GITHUB_EVENT_PATH
-      ? require(process.env.GITHUB_EVENT_PATH)
+      ? JSON.parse(
+          readFileSync(process.env.GITHUB_EVENT_PATH, {encoding: 'utf8'})
```
stephenmichaelf: Do we want to add error handling here? Make sure the file exists first and print message if not?

stephenmichaelf: And while we are here, should we debug print a serialization of the context?

damccorm: Error handling yes, serialization of the context seems like a no to me. Its pretty massive, and since we're built to be a dependency I'm a little leary of large debug output that may or may not be helpful, especially since lots of actions probably won't use that directly.

gnzzz:
What's the release schedule for the `@actions/github` npm package? I'm currently manually webpacking this change but would much rather use the proper releases.

damccorm:
Just published, sorry for the delay - https://www.npmjs.com/package/@actions/github

PR #103, opened 2019-08-31
Fix broken link to Javascript Action

andrewbredow:
Noticed when going through this guide that one of the links was missing a `.md`.


damccorm:
Good catch, thanks for this!

PR #105, opened 2019-09-03
Create commands.md

### LINKED ISSUE #57: Add docs for ##[] commands from scripts
e.g. `##[error]Error message`

damccorm:
✨ [Rendered](https://github.com/actions/toolkit/blob/commands/docs/commands.md) ✨ 

Added docs on :: commands. Fixes #57

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -4,14 +4,16 @@ The [core toolkit package](https://github.com/actions/toolkit/tree/master/packag
 setting results, logging, registering secrets and exporting variables across actions. Sometimes, however, its useful to be able to do
 these things in a script or other tool.
 
-To allow this, we permit a special `::` syntax which, if logged to stdout, will allow the runner to interpret your commands.
-The following commands are all supported:
+To allow this, we provide a special `::` syntax which, if logged to `stdout`, will allow the runner to perform special behavior on
+your commands. The following commands are all supported:
 
 ### Set an environment variable
 
 To set an environment variable, use `::set-env`:
 
-`echo ::set-env name=FOO::BAR`
+```sh
+echo ::set-env name=FOO::BAR
+```
 
 Running `$FOO` in a future step will now return `BAR`
 
@@ -19,23 +21,38 @@ Running `$FOO` in a future step will now return `BAR`
 
 To prepend a string to PATH, use `::addPath`:
 
-`echo ::add-path::BAR`
+```sh
+echo ::add-path::BAR
+```
 
 Running `$PATH` in a future step will now return `BAR:{Previous Path}`;
 
 ### Set outputs
 
-To set an output for the step, use `set-output`:
+To set an output for the step, use `::set-output`:
 
-`echo ::set-output name=FOO::BAR`
+```sh
+echo ::set-output name=FOO::BAR
+```
 
 Running `steps.[step-id].outputs.FOO` in your Yaml will now give you `BAR`
 
-### Masking Values from Logs
+```yaml
+steps:
+  - name: Set the value
+    id: step_one
+    run: echo ::set-output name=FOO::BAR
+  - name: Use it
+    run: echo ${{ steps.step_one.outputs.FOO }}
+```
 
-To mask a value from the logs, use `add-mask`:
+### Masking Values in Logs
 
-`echo ::add-mask::BAR`
+To mask a value in the logs, use `::add-mask`:
+
+```sh
+echo ::add-mask::BAR
+```
 
 Now, future logs containing BAR will be masked. E.g. running `echo "Hello FOO BAR World"` will now print `Hello FOO **** World`.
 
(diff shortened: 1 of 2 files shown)


docs/commands.md
```diff
@@ -0,0 +1,58 @@
+# :: Commands
+
+The [core toolkit package](https://github.com/actions/toolkit/tree/master/packages/core) offers a number of convenience functions for
+setting results, logging, registering secrets and exporting variables across actions. Sometimes, however, its useful to be able to do
+these things in a script or other tool.
+
+To allow this, we permit a special `::` syntax which, if logged to stdout, will allow the runner to interpret your commands.
```
JasonEtco: Tiny suggestion, making `stdout` look a little more technical:

```suggestion
To allow this, we permit a special `::` syntax which, if logged to `stdout`, will allow the runner to interpret your commands.
```

docs/commands.md
```diff
@@ -0,0 +1,58 @@
...
+The following commands are all supported:
+
+### Set an environment variable
+
+To set an environment variable, use `::set-env`:
+
+`echo ::set-env name=FOO::BAR`
```
JasonEtco: For easier scanning and nicer formatting, what do you think about making these block-level code blocks?

`echo ::set-env name=FOO::BAR`

vs:

```
echo ::set-env name=FOO::BAR
```

docs/commands.md
```diff
@@ -0,0 +1,58 @@
...
+
+Running `$PATH` in a future step will now return `BAR:{Previous Path}`;
+
+### Set outputs
+
+To set an output for the step, use `set-output`:
+
+`echo ::set-output name=FOO::BAR`
+
+Running `steps.[step-id].outputs.FOO` in your Yaml will now give you `BAR`
```
JasonEtco: I think this is a little unclear - maybe we can add some example usage?

```yaml
steps:
  - name: Set the value
    id: step_one
    run: echo ::set-output name=FOO::BAR
  - name: Use it
    run: echo ${{ steps.step_one.outputs.FOO }}
```

And/or link out to `core.setOutput`, which could also show a similar example.

1138-4EB:
Maybe `##[debug]` in https://github.com/actions/toolkit/blob/master/docs/action-debugging.md#step-debug-logs needs to be updated to `::[debug]`?

BTW, is this related to the `::[group]`/`::[endgroup]` feature? I'm currently using it to provide a 'clean' first view. See https://github.com/1138-4EB/ghdl/runs/210613754#step:3:46. I think it'd be useful if this was documented too. Moreover, it'd be really useful to:

- Allow nesting groups.
- Make timing a group a built-in feature. Ideally, show the elapsed time in the same row as the arrows (first/title line of the group), not in the end.
- Print the arrows (expand/collapse) in the margin, not in the body. And add vertical indentation lines.

I don't know if any of these features is available already through some other keyword.

docs/commands.md
stephenmichaelf: Nit: permit -> provide?

stephenmichaelf: Interpret -> perform special behavior based on

docs/commands.md
```diff
@@ -0,0 +1,58 @@
...
+
+To prepend a string to PATH, use `::addPath`:
+
+`echo ::add-path::BAR`
+
+Running `$PATH` in a future step will now return `BAR:{Previous Path}`;
+
+### Set outputs
+
+To set an output for the step, use `set-output`:
```
stephenmichaelf: Missing `::` at start of `set-output`

docs/commands.md
```diff
@@ -0,0 +1,58 @@
...
+
+`echo ::set-output name=FOO::BAR`
+
+Running `steps.[step-id].outputs.FOO` in your Yaml will now give you `BAR`
+
+### Masking Values from Logs
```
stephenmichaelf: Nit: Masking values from logs -> Masking values in logs?

docs/commands.md
```diff
@@ -0,0 +1,58 @@
...
+
+To mask a value from the logs, use `add-mask`:
```
stephenmichaelf: Same here, missing `::`. If we want this line to not have it, that works too. But we want to be consistent :) Right now it's mixed with some having it and some not.

docs/commands.md
```diff
@@ -0,0 +1,58 @@
...
+
+`echo ::add-mask::BAR`
+
+Now, future logs containing BAR will be masked. E.g. running `echo "Hello FOO BAR World"` will now print `Hello FOO **** World`.
```
stephenmichaelf: Nit: `****` -> `***`?

stephenmichaelf: Or is it always a fixed amount to keep it more secret?

docs/commands.md
```diff
@@ -0,0 +1,58 @@
...
+
+CAUTION: Do **not** mask short values if you can avoid it, it could render your output unreadable (and future steps' output as well).
+For example, if you mask the letter `l`, running `echo "Hello FOO BAR World"` will now print `He*********o FOO BAR Wor****d`
+
+### Log Level
+
+Finally, there are several commands to emit different levels of log output.
+To emit [debug](https://github.com/actions/toolkit/blob/master/docs/action-debugging.md) output, use `::debug`:
```
stephenmichaelf: This may be easier to understand as a table of log level - echo command.

docs/commands.md
damccorm: I think its always fixed

docs/commands.md
damccorm: I like it - lmk if my changes are what you were thinking

damccorm:
Thanks for the feedback, incorporated it all in!

JasonEtco:
@damccorm super excited for this. I have a potentially dumb question: is there a reason this is being added to the `actions/toolkit` docs rather than [the GitHub Actions developer docs](https://help.github.com/en/categories/automating-your-workflow-with-github-actions)? I ask because folks that don't want to write their actions in TS/JS may not come across it if they don't need the toolkit, and those that do would just use the `core` methods.

damccorm:
@JasonEtco IMO it should probably end up in both places. In general, I like starting here since this is public and I can get more community feedback 😄

PR #107, opened 2019-09-04
JavaScript walk through update

bryanmacfarlane:


docs/javascript-action.md
```diff
@@ -6,57 +6,55 @@ The [javascript-template](https://github.com/actions/javascript-template) repo c
...
 ```
 
+Update the `author` element in the package.json file.
+
 # Dev Workflow
 
 The workflow below describes one possible workflow with a branching strategy.  Others exist.  
 
-> Key Point: the branch that users reference in their workflow files should reference an action that has **only** the production dependencies. 
+> Key Point: the branch that users reference in their workflow files should reference an action from a distribution branch that has **only** the production dependencies.
```
stephenmichaelf: This reads a bit confusing, can we add an example? "For instance, you would want to reference X and not Y".

docs/javascript-action.md
```diff
@@ -6,57 +6,55 @@ The [javascript-template](https://github.com/actions/javascript-template) repo c
...
+  name:
+    description: 'the name to say hello to'
+    default: 'World'
 runs:
   using: 'node12'
   main: 'lib/main.js'
 
 ```
 
+The `name` input will be referenced by workflow authors using the `with:` keyword.
```
stephenmichaelf: Add example?

README.md
```diff
@@ -27,9 +27,9 @@ The toolkit provides five separate packages.   See the docs for each action.
 
 Actions are units of work which can either run in a container or on the host machine.
 
-[Choosing an action type](docs/action-types.md): Outlines the differences and why you would want to create a host or a container based action.
+[Choosing an action type](docs/action-types.md): Outlines the differences and why you would want to create a JavaScript or a container based action.
```
stephenmichaelf: Should we make it explicit here that we recommend Typescript to make life easier?

docs/javascript-action.md
bryanmacfarlane: Yeah, it's covered in the `Users referencing` section further down in this doc but it's not using the `with` keyword.  I will update this blurb to mention that and update that section to make it clearer.

docs/javascript-action.md
bryanmacfarlane: This is covered further down in the doc in the section for creating a distribution branch.

bryanmacfarlane: The following paragraph also discussed a concrete distribution branch of releases/v1.  It also clarifies that the runner will download and run from that distribution branch.

README.md
bryanmacfarlane: I think the next block covers that.  Choosing an action type is orthogonal to dev tooling chosen.  Case in point is our next action / in official docs is a topic on creating an action from scratch (hello world) in pure javascript.  This doc strictly covers javascript vs container and why you would do one or the other.

bryanmacfarlane: NOTE to self:  need to add another block encouraging 1 action to 1 repo for market place and versioning to be viable in this model

docs/assets/action-releases.drawio
```diff
@@ -0,0 +1 @@
...m7NLEeY8qf63ZeJVAuxcj0E4gkcjqTaCRKtBoUIE2BxZo27scgbZUcKyrBseRwPHRcODYKjjOVYPjyVpsSFNov5vCI0yhvd8Uuv2YQs8azBS6qkBrWHRFAu3awwl08ovef//7r+Rhtfj+7UtBvkR3N5roY4TcDFYdT+AiqS5omEdwbx9FCyy0aYTvcM5wyicF+wKu4rDABfRZqWEMYLZgl1GYZWRZ7od8gWkKD8pgEYO+NlXjBSnSMiU5quTY9E9kEiWt69kqSCbqEyWNSznD0VOas0elJGa5TLwAJYrzKGWbzyGkAqZvuGAa4QE9KJjAppSyrm4LE1fsbaHgVYrgsC1OAaaPvGGexnG2TbC70nwK2XKdrifja2DTCRc6F2waZ1OTHroi9ee7wzmbWoTefZpdPs0uUu9ybKx+HBsnMG890/GQ7zng1XhBl1u2tnVPClhdBHUp67pn85602+0oOmSartmgolgyTW9MKZkzPEDDp3nydtW7Z3TVu2vZqno3+lTvl5nq3yiMgZP9r1Uedj/Kw5fzHgeeESkTubJNO9/RgXZLPUVBzDEsCFUgzGybJuCPv1214DuSftZEVHafWsFX8FAiqojisMQ1h3GFDtzMuJKLZGucZZi3EOKorr0y+85Jeg/KwO6eRv17clBmqurf7dN3DPYDfVToPA8LtoG/Z8yMpJjZDoaOmcXRXTv68lSLfUXRV3BxySfNQfcFeFBHhF+v94QEPc/u5xiSfZQN36F+juwwuWa/fo6pSYL51y3Pgd/1PIfOppiao9TfS577T6cIWrdDol2h03tcdCgZLzO3d2uDlDSEvDFujQMpeWSQziZp+X190bQvuxbIb+zJp5oH2zXJK+o7wWeqGb6rCuBNQ/ZQPEsxY71G8KZ6bF3nXCfLN5xf9QPZUVTzq/3CoGa2ytYxJgjGFFN24Mmy3oZ08Gzw0BsZH0DNwLAPcLnMea/47cIY2N1jkMBAA8N41oQYOiAhNicrvG+u3vNilv73LMfj7Xfx1r6s0GtezFQTY1tSW5chdJWrxm9KCunYxp/Exsnv8QWBApIOo7MJpZi4hdHuN3euByp/i1+5L5F5Pqx072jJkU3zUzS+7XFYzHSBzcFbtteBb22Io9kPUfc//XwZD09OOx3q55umfIJwsgQWFJtfRNbdm9+dWg//AQ==</diagram></mxfile>
```
damccorm: Do we need this file/what is it doing?

bryanmacfarlane: Yes!  Try and update the image included without it ;) . It's a draw.io file per the file name and the xml declaration. ;)

docs/javascript-action.md
bryanmacfarlane: Good catch ...

Issue #106, opened 2019-09-03
Allow for alternative authentication mechanisms for the Octokit object on construction.


josephperrott:
Currently for the `@actions/github` this library extends `@octokit/rest`'s `Octokit` class and in the `super` call assumes that the authentication is a standard token. ([code link](https://github.com/actions/toolkit/blob/fc9958ec6a57d1b08e9a387c49971faecd0f0405/packages/github/src/github.ts#L18))

Other authentication methods should be valid here as well.

damccorm:
Looking at your issue, it looks like your use case is if you don't want to authenticate at all, right? I think I see a use case there, not sure if I see a use case for alternative auth methods - does that seem right?

josephperrott:
The use case I specifically am running into is needing to authentication with a JWT.

ericsciple:
I'm not sure whether we need to implement this.

The toolkit class is a thin wrapper convenience function for the common case `GITHUB_TOKEN` or PAT.

I could see the need if an input accepted a PAT or JWT and a function provided a single abstraction to do the right thing.

Otherwise I would recommend using the Octokit class directly.

Thoughts?

mies386:
#106
go_som

mies386:
#285 
Pastebin_master_social_login
#292

mies386:
#106/€Cash.paypal;bitcoin<€9999€>$cash_come_login_pal_pay;#social_network
#285 
#293 
<#$€$€cash_download;Apps>Pay:pal@internet^^==GitHub.patch/h.a.c.k.b;a;n;k-S;p;a;r<=#kasse#=><server$€&@!^#*

eine:
IMHO, `@actions/github` should not duplicate or wrap any of the features provided by `@octokit/rest`, but extend them. Therefore, the authentication and all regular operations should be done with octokit. When a feature from `@actions/github` is required, it should accept an existing octokit object as an argument. Therefore, it should not be required for `@actions/github` to handle authentication at all. For example:

```js
const octokit = new Octokit(getInput('token', { required: false }))
const ghkit = new github.GitHub(octokit)
```

instead of https://github.com/eine/issue-runner/blob/258088daba8c0e57368d08b2afbc5795cf282490/ts/main.ts#L39-L45

ericsciple:
[...]

PR #113, opened 2019-09-05
Accept Octokit.Options in the GitHub constructor

jclem:
This adds an optional object of Octokit client options to the GitHub constructor. This allows for debugging, API previews, etc.

This PR also bumps TypeScript to 3.6.2 so that I can use the `Omit` type.

### DIFF
--- a/packages/github/src/github.ts
+++ b/packages/github/src/github.ts
@@ -14,8 +14,8 @@ export class GitHub extends Octokit {
     variables?: Variables
   ) => Promise<GraphQlQueryResponse>
 
-  constructor(token: string) {
-    super({auth: `token ${token}`})
+  constructor(token: string, opts: Omit<Octokit.Options, 'auth'> = {}) {
+    super({...opts, auth: `token ${token}`})
     this.graphql = defaults({
       headers: {authorization: `token ${token}`}
     })
--- a/package-lock.json
+++ b/package-lock.json
@@ -9530,9 +9530,9 @@
       "dev": true
     },
     "typescript": {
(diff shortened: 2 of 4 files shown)


packages/github/src/github.ts
```diff
@@ -14,8 +14,8 @@ export class GitHub extends Octokit {
     variables?: Variables
   ) => Promise<GraphQlQueryResponse>
 
-  constructor(token: string) {
-    super({auth: `token ${token}`})
+  constructor(token: string, opts: Omit<Octokit.Options, 'auth'> = {}) {
```
damccorm: Rather than omitting auth, can we just overload the constructor to something like:
```
constructor(token: string) { ... }
constructor(opts = {}) { ... }
```

That would also resolve #106

damccorm:
Might also be good to doc this as an option in https://github.com/actions/toolkit/blob/master/packages/github/README.md

packages/github/src/github.ts
jclem: True, but then how would we construct the authorization header for GraphQL requests?

jclem:
The more I think about it, the more I like the simplicity of forcing a token to be provided and then accepting token-less options as a second argument.

For cases where more complex authorization is required (as is in the case of @josephperrott's [issue](https://github.com/actions/toolkit/issues/106)), it seems just as easy to use @octokit/rest directly.

packages/github/src/github.ts
damccorm: Good point

PR #117, opened 2019-09-06
Walkthrough update

mannyman1115:
Over the waterfront always alone and sad


bryanmacfarlane:
Closing since the walkthrough moved to javascript-action

PR #122, opened 2019-09-07
added file name of action.yml file. 

abelsquidhead:
For a first time user creating their first action, everything is confusing. They might not know which file they need to update. Adding the filename makes this step easier to follow.


damccorm:
Thanks for the contribution!

PR #123, opened 2019-09-07
added -f because node_modules is in the .gitignore

abelsquidhead:


EmilioCristalli:
isn't the intention of `1.` to remove it from git ignore? ("1. **Do not ignore node_modules**:  Add a `!` in front of the `node_modules` line.")

abelsquidhead:
> isn't the intention of `1.` to remove it from git ignore? ("1. **Do not ignore node_modules**: Add a `!` in front of the `node_modules` line.")

Ahhh fair enough. I missed that

PR #125, opened 2019-09-08
step 1. was slightly confusing. 

abelsquidhead:
For someone that doesn't use git all the time, they might not know they need to go into the .gitignore file. The first time I read this I thought the direction was saying to rename the directory !node_modeules. This update just makes the directions a little more clear. A little less friction for someone starting out with actions.


damccorm:
This looks like a helpful addition, thanks for the contribution!

PR #131, opened 2019-09-11
Update readme to show how to use the secret GITHUB_TOKEN

warrenbuckley:
This updates the example with some inline comments to show that you need to use the  `GITHUB_TOKEN` secret as the token which will need to be passed in for you to use & authenticate the GitHub API

### DIFF
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -11,6 +11,8 @@ const github = require('@actions/github');
 const core = require('@actions/core');
 
 // This should be a token with access to your repository scoped in as a secret.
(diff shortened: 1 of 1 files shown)


rowi1de:
Is this up-to-date? I'm getting Bad Credentials when doing
```js
import * as core from '@actions/core';
import { context, GitHub } from '@actions/github'

const repoToken = core.getInput('repo-token', { required: true })
const client = new GitHub(repoToken)
```

```yaml
      with:
        repo-token: ${{ secrets.GITHUB_TOKEN }}
```

PR #134, opened 2019-09-12
tool-cache: clarify that extractTar extracts gzipped tars

WebFreak001:
The default downloadTool implementation strips the filename so when passing it to the extractTar function, it doesn't have a way of knowing the format without manually specifying it. However what the extractTar function arguments meant and how to specify them wasn't clear before reading the source code, so the documentation here got updated to reflect that.

I put the default value of the flags argument into the documentation because it seems to be missing when shown in auto completion when depending on tool-cache.

### DIFF
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -183,11 +183,11 @@ export async function extract7z(
 }
 
 /**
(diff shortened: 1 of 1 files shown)


ericsciple:
@WebFreak001 thanks!

PR #138, opened 2019-09-17
Fix broken link to github-package.

metcalfc:


metcalfc:
Looks like this is OBE with recent doc reorg. Closing for now.

PR #141, opened 2019-09-18
Implement set-secret

damccorm:


packages/core/RELEASES.md
```diff
@@ -1,7 +1,8 @@
 # @actions/core Releases
 
-### Unreleased
+### 1.2.0
 
+- Don't throw on set-secret [#141](https://github.com/actions/toolkit/issues/141)
```
bryanmacfarlane: I wouldn't phrase as "Don't throw".  The release notes should say that set-secret is now available for use.

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/core",
-  "version": "1.1.0",
+  "version": "1.2.0",
```
bryanmacfarlane: I would just patch 1.1.1 so anyone binding to ^1.1 will just get this

Issue #104, opened 2019-09-03
Release notes for @actions/github 1.0.1


xt0rted:
[I just got a PR](https://github.com/xt0rted/block-autosquash-commits-action/pull/37) from dependabot bumping `@actions/github` to version `1.0.1` but there doesn't seem to be release notes for this [in the project folder](https://github.com/actions/toolkit/blob/master/packages/github/RELEASES.md), or a [release/tag](https://github.com/actions/toolkit/releases). Is there an easier way to see what changed than looking through the commits?

damccorm:
Whoops, sorry forgot to update the release notes - they should be updated now, it included a small change to enable simpler webpack configs (#101).

xt0rted:
Same issue with the 1.1.0 releases.

damccorm:
Updated, sorry about that - will get better at this

xt0rted:
There's also a 1.1.0 update for `@actions/core`, are there any changes to that beside #98?

damccorm:
No, that's it. Will update there as well

xt0rted:
@damccorm I was looking into this more and what I found is because this is a mono repo the `package.json` of each package needs to be modified like so:

```diff
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/actions/toolkit.git"
+    "url": "git+https://github.com/actions/toolkit.git",
+    "directory": "packages/github"
   },
```

Then `RELEASES.md` renamed to `RELEASE.md` or `CHANGELOG.md` ([as per dependabot](https://github.com/dependabot/dependabot-core/blob/35afcd3efa13e9c1bfe04edfd4461223b1681c0a/common/lib/dependabot/metadata_finders/base/changelog_finder.rb#L21)).

That should then allow Dependabot to pull the correct change log for each package instead of showing a [list of commits](https://github.com/xt0rted/slash-command-action/pull/4) like it's doing now. I was going to PR the changes but wasn't sure if you wanted them made.

bryanmacfarlane:
>  I was going to PR the changes but wasn't sure if you wanted them made.

We would gladly accept a PR.

xt0rted:
This change is in #143

xt0rted:
The change log is showing in Dependabot's PRs now https://github.com/xt0rted/block-autosquash-commits-action/pull/44

PR #143, opened 2019-09-19
Add 'repository.directory' to package manifests

hmarr:
This is supported by npm ([docs](https://docs.npmjs.com/files/package.json#repository)). It can help tools like Dependabot figure out what's changed between versions.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -22,7 +22,8 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/actions/toolkit.git"
+    "url": "git+https://github.com/actions/toolkit.git",
+    "directory": "packages/core"
   },
   "scripts": {
     "test": "echo \"Error: run tests from root\" && exit 1",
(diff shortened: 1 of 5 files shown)


xt0rted:
This change was mentioned in https://github.com/actions/toolkit/issues/104#issuecomment-529155474. I think for it to fully work with dependabot the `RELEASES.md` files also need to be renamed.

hmarr:
The other option is just supporting those files in Dependabot. https://github.com/dependabot/dependabot-core/pull/1392 adds support for that.

hmarr:
Apparently dependabot looks for filenames *containing* those strings (case-insensitive), so it should already work as `release` is in the list.

bryanmacfarlane:
@xt0rted - it sounds like we're good then since it looks for files that contain `release`.  OK?

xt0rted:
@bryanmacfarlane yea, this should be all that's needed for dependabot to pickup the change logs.

PR #149, opened 2019-09-20
add core method to saveState and getState.

TingluoHuang:
In wrapper action, action author uses these 2 methods for sharing information between his action's main-execution and post-execution, this information is only available to this pair of action's executions, other action within the same job won't get it.

Ex: 
In the main execution:
```javascript
  core.saveState("pidToKill", 12345);
```
In the post-execution:
```javascript
  var pid = core.getState("pidToKill");
  process.kill(pid);
```

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -126,5 +126,5 @@ const core = require('@actions/core');
 
 var pid = core.getState("pidToKill");
 
-kill(pid);
+process.kill(pid);
 ```
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -95,3 +95,36 @@ const result = await core.group('Do something async', async () => {
   return response
 })
 ```
+
+#### Action state
+
+You can use this library to save state and get state for sharing information between a given wrapper action: 
+
+**action.yml**
+```yaml
+name: 'Wrapper action sample'
+inputs:
+  name:
+    default: 'GitHub'
+runs:
+  using: 'node12'
+  main: 'main.js'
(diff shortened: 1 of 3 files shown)


bryanmacfarlane:
It would be good to add some context in the description above ^^

bryanmacfarlane:
Please add documentation to core/README.md

packages/core/README.md
```diff
@@ -94,4 +94,37 @@ const result = await core.group('Do something async', async () => {
...
+const core = require('@actions/core');
+
+core.saveState("pidToKill", 12345);
+```
+
+In action's `cleanup.js`:
+```js
+const core = require('@actions/core');
+
+var pid = core.getState("pidToKill");
```
joshmgross: Will this example work?
I think it might be nice to have examples use best practices and be functional when copied.

`kill()` will require a `core.exec` call.

I'm thinking an example like:
```js
const core = require('@actions/core');
const exec = require('@actions/exec');

const pid = core.getState("pidToKill");

await exec.exec(`kill ${pid}`);
```

bryanmacfarlane: https://nodejs.org/api/child_process.html#child_process_subprocess_kill_signal

joshmgross: The example doesn't have access to a `child_process` in the post to kill though, it's specifying a `pid`

TingluoHuang: i will change this to be `process.kill(pid)`

PR #135, opened 2019-09-15
Fix link on docs/typescript-action.md

tvdias:

PR #155, opened 2019-09-21
Fix a link

deificx:
At the end of the typescript-action.md it links to `github-package`, but the extension was missing a linking failed.


deificx:
Actually addressed in #135

PR #157, opened 2019-09-21
Wrap example in async function

squalrus:
As someone not too familiar with async/await JavaScript, I was hung up on this for a bit. If this is too distracting from the example itself, I can understand not integrating it.


bryanmacfarlane:
Thanks for your contribution.  You are correct that it's confusing at first.

PR #167, opened 2019-09-30
add/fix API for creating secrets

squarebracket:
I noticed in the actual code that there's a function for adding secrets, but it seems it uses a deprecated command and just errors out. I believe this is the proper fix to make it work again.

Please let me know if there's anything I missed (`CHANGELOG` etc). Also, the build command fails for `tool-cache`. I don't know if that's normal or not.


squarebracket:
Looked at the workflow and saw I was missing `npm run bootstrap`.

Issue #168, opened 2019-09-30
First set-output ignored


elgohr:
As seen in https://github.com/elgohr/ecr-login-action/pull/2 , the first `set-output` seems to be ignored. I wasn't able to figure out why, as I'm not able to test it without publishing test releases. Could you please help on that?

thboop:
Hey @elgohr ,
Commands need to be the first item on their line (ignoring white-space) in order to be executed. That may be the issue you are running into, we can update the docs to make that more clear. Do you have example output from your action you could share?

elgohr:
As the acion handles with credentials I don't have an example.
But the hint with the new line was really good. I'll try that first.

thboop:
I'm going to close this issue for now, please let me know if you are still encountering issues and we can investigate again!

PR #171, opened 2019-10-01
Update command docs to specify a new line is needed

thboop:
Related to #168


dakale:
LGTM. Having something at all here at least gives people a starting point on what to look out for

PR #172, opened 2019-10-01
updating core docs and bumping version

bryanmacfarlane:


packages/core/README.md
```diff
@@ -4,45 +4,51 @@
...
+```js
+const myInput = core.getInput('inputName', { required: true });
 
 core.setOutput('outputKey', 'outputVal');
 ```
 
 #### Exporting variables
 
-You can also export variables for future steps. Variables get set in the environment.
+Since steps run out of proc, you can use `exportVariable` to add it to this step and future steps environment blocks.
```
thboop: Might be worthwhile to define out of proc (I believe its not that common of a terminology)
"Steps do not share environmental variables because each step executes in a different process (out-of-proc). You can use `exportVariable` to add a variable to the current step and all remaining steps in the job."

packages/core/README.md
```diff
@@ -4,45 +4,51 @@
...
-const core = require('@actions/core');
-
-core.addPath('pathToTool');
+core.addPath('/path/to/mytool');
 ```
 
 #### Exit codes
 
-You should use this library to set the failing exit code for your action:
+You should use this library to set the failing exit code for your action.  If you status is set and the script runs to completion, that will lead to a success.
```
thboop: This is confusing to me "If you status is set and the script runs to completion, that will lead to a success."

Could you possibly mean "If **no** status is set and the script runs to completion, that will lead to a success."

packages/core/README.md
```diff
@@ -4,45 +4,51 @@
...
 
-const myInput = core.getInput('inputName', { required: true });
+// typescript
+import * as core from '@actions/core';
+```
 
-// Do stuff
+#### Inputs/Outputs
+
+Action inputs can be read with `getInput`.  Outputs be set with `setOutput` which makes them available to be mapped into inputs of other actions to ensure they are decoupled.
```
thboop: "Outputs **can** be set" would be preferred
Spacing between `GetInput`. and Outputs has some extra spaces.

PR #174, opened 2019-10-01
setSecret

bryanmacfarlane:
implement core.setSecret.  see readme

### DIFF
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -42,7 +42,7 @@ export function exportVariable(name: string, val: string): void {
 
 /**
  * Registers a secret which will get masked from logs
- * @param val value of the secret
+ * @param secret value of the secret
  */
 export function setSecret(secret: string): void {
   issueCommand('add-mask', {}, secret)
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -31,7 +31,7 @@ export enum ExitCode {
 //-----------------------------------------------------------------------
 
 /**
- * sets env variable for this action and future actions in the job
+ * Sets env variable for this action and future actions in the job
  * @param name the name of the variable to set
  * @param val the value of the variable
  */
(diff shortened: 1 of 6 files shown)


packages/core/src/core.ts
```diff
@@ -41,16 +41,11 @@ export function exportVariable(name: string, val: string): void {
 }
 
 /**
- * exports the variable and registers a secret which will get masked from logs
- * @param name the name of the variable to set
+ * Registers a secret which will get masked from logs
  * @param val value of the secret
```
TingluoHuang: `secret` instead of `val`, or the other way, make sure the method and comment lines up. 😆

bryanmacfarlane: will do

squarebracket:
I think the nomenclature of `setSecret` is kinda weird since you're just adding a mask and not actually setting anything. Something like `maskValue` might have been a better choice.

bryanmacfarlane:
@squarebracket - thanks.  Think of it as a setting on the string.  We are "setting the string as a secret".  The internal implementation detail is the runner adds it to it's mask list.

squarebracket:
It's not that I disagree in terms of what's being done -- adding a string to a list of things to `s/$string/***/` -- but just in terms of semantics, and how words are generally used and understood. Compare to the [section in the docs](https://help.github.com/en/articles/virtual-environments-for-github-actions#creating-and-using-secrets-encrypted-variables):

> Secrets are encrypted environment variables created in a repository

Analogously, setting an output creates an entry in the `steps` context, which can then be accessed by name later on.

It makes sense that the raw command was changed from `::set-secret` to `::add-mask` since it more precisely describes what is happening. Perhaps I'm being pedantic here, but I figure that if something can be made clearer, there's no harm in _not_ doing it.

In any case, I opened #177 to try to clarify things a bit. Do with it as you see fit.

alphachart (commented):
K

PR #177, opened 2019-10-02
Make setting secret docs a bit clearer

squarebracket:
Just trying to make things a bit clearer.

As an implementation question -- I assume calling `::add-mask` will only apply for the job in which it's run? Perhaps that is worth noting, perhaps it is obvious, I'm not sure.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -32,11 +32,12 @@ Since each step runs in a separate process, you can use `exportVariable` to add 
 core.exportVariable('envVar', 'Val');
 ```
 
-#### Setting a secret
+#### Setting a string as secret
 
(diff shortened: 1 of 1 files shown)


packages/core/README.md
```diff
@@ -32,11 +32,12 @@ Since each step runs in a separate process, you can use `exportVariable` to add
 core.exportVariable('envVar', 'Val');
 ```
 
-#### Setting a secret
+#### Setting a string as secret
```
ericsciple: personally i prefer the existing verbiage :)

ericsciple: ..just my 2 cents. i'll defer to others

bryanmacfarlane: I agree.

PR #180, opened 2019-10-03
tool-cache: Work around a weird GNU Tar behavior on Windows

Ecco:
For some reason GNU Tar insists on parsing the destination path using an Unix path delimiter (`/`), even on windows where the path delimiter is `\`.

This patch simply rewrites `\` to `/` in the destination path.

### DIFF
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -200,6 +200,10 @@ export async function extractTar(
   }
 
   dest = dest || (await _createExtractFolder(dest))
+  if (IS_WINDOWS) {
+    // For some reason, GNU Tar on Windows expects paths to be slash-separated even though this is not the standard Windows path delimiter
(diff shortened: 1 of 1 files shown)


packages/tool-cache/src/tool-cache.ts
```diff
@@ -200,6 +200,10 @@ export async function extractTar(
   }
 
   dest = dest || (await _createExtractFolder(dest))
+  if (IS_WINDOWS) {
+    // For some reason, GNU Tar on Windows expects paths to be slash-separated even though this is not the standard Windows path delimiter
+    dest = dest.replace(/\\/g, '/')
```
ericsciple: I proposed an alternate solution for https://github.com/actions/toolkit/pull/165 which will eliminate the need for this.

Let's close on the other PR first before proceeding with this PR.

joshmgross:
Fixed by https://github.com/actions/toolkit/pull/264

PR #187, opened 2019-10-14
i-can-trigger-your-action

RammusXu:


jclem:
Is this tied to a report somewhere? Pull request workflows from repository forks:

- Do not receive secrets. 
- Will be billed to the repository owner. 

The `env` being dumped here is the same as the `env` for every other Ubuntu workflow.

PR #198, opened 2019-10-31
Document Problem Matcher Commands

### LINKED ISSUE #56: Add Problem matcher docs
Would be helpful to have a schema/explanation of what matchers actually do.

thboop:
[Rendered](https://github.com/actions/toolkit/blob/users/thboop/addProblemMatcherdocs/docs/problem-matchers.md)

Resolves #56

### DIFF
--- a/docs/problem-matchers.md
+++ b/docs/problem-matchers.md
@@ -50,14 +50,14 @@ The following fields are available for problem matchers:
             code: a group number containing the error code
             message: a group number containing the error message. **required** at least one pattern must set message
             fromPath: a group number containing the base path of the file, otherwise the location of the git repository on disk is used
-            loop: See [Multiline Matching](#Multi line matching)
+            loop: loops until a match is not found, only valid on the last pattern of a multipattern matcher
         }
     ]
 }
 ```
 
 
-## Multi line matching
+## Multiline Matching
 Consider the following output:
 ```
 test.js
@@ -89,7 +89,6 @@ The eslint-stylish problem matcher defined below catches that output, and create
                     "severity": 3,
                     "message": 4,
                     "code": 5,
-                    // Loops until a match is not found
                     "loop": true
                 }
             ]
--- a/docs/problem-matchers.md
+++ b/docs/problem-matchers.md
@@ -1,5 +1,5 @@
 # Problem Matchers
-Problem Matchers are a way to scan output of builds for a specified pattern and surface that information in the UI. [GitHub Annotations](https://developer.github.com/v3/checks/runs/#annotations-object-1) and log file decorations are created for pattern matches. Problem Matchers are enabled via the toolkit [command](commands.md#Problem Matchers).
+Problem Matchers are a way to scan output of builds for a specified pattern and surface that information in the UI. [GitHub Annotations](https://developer.github.com/v3/checks/runs/#annotations-object-1) and log file decorations are created for pattern matches. Problem Matchers are enabled via the toolkit [command](commands.md#problem-matchers).
 
 For example, consider the ESLint compact output:
 ```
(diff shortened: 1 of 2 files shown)


docs/problem-matchers.md
```diff
@@ -0,0 +1,116 @@
...
+```
+We can define a problem matcher in json that detects input in that format:
+```
+{
+    "problemMatcher": [
+        {
+            "owner": "eslint-compact",
+            "pattern": [
+                {
+                    // The pattern that provides groups to match against
```
ericsciple: comments are not legal JSON

docs/problem-matchers.md
```diff
@@ -0,0 +1,116 @@
...
+        }
+    ]
+}
+```
+
+The following fields are available for problem matchers:
+
+```
+{
+    owner: An id field that can be used to remove or replace the problem matcher. **required**
```
ericsciple: ID, not id (psychology term) :)

ericsciple: or \"name\"

docs/problem-matchers.md
```diff
@@ -0,0 +1,116 @@
...
+    pattern: 
+    [
```
ericsciple: consider moving this left bracket up (consistent formatting, also more standard i think)

docs/problem-matchers.md
```diff
@@ -0,0 +1,116 @@
...
+        {
+            regexp: The regex containing the groups. **required**
+            file: a group number containing the file name
+            line: a group number containing the line number
+            column: a group number containing the column information
+            severity: a group number containing either 'warning' or 'error' case-insensitive. Defaults to `error`
+            code: a group number containing the error code
+            message: a group number containing the error message. **required** at least one pattern must set message
+            fromPath: a group number containing the base path of the file, otherwise the location of the git repository on disk is used
```
ericsciple: \"base\" might be confusing (e.g. `basename /foo/bar.txt` outputs `bar.txt`)

Consider "root" or "root path" terminology. Note fromPath is expected to be a *file*. There is a bug in the runner i'm working on a fix now.

docs/problem-matchers.md
ericsciple: we shouldnt suggest customers use them, if it we switch to a different json parser in the future, it might break.

docs/problem-matchers.md
```diff
@@ -0,0 +1,116 @@
...
+                    "loop": true
+                }
+            ]
+        }
+    ]
+}
+```
+
+The first pattern matches the `test.js` line and records the file information. This line is not decorated in the UI.
+The second pattern loops through the remaining lines with `loop : true` until it fails to find a match, and surfaces these lines prominently in the UI.
```
ericsciple: extra space preceding the colon in `loop : true`

docs/problem-matchers.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Adding and Removing Problem Matchers
+Problem Matchers are enabled and removed via the toolkit [commands](commands.md#problem-matchers).
+
+## Duplicate Problem Matchers
+Registering two problem-matchers with the same owner will result in only the problem matcher registered last running.
+
+## Examples
+Some of the starter actions are already using problem matchers, for example:
+- [setup-node Problem Matchers](https://github.com/actions/setup-node/tree/master/.github)
```
ericsciple: most of these links end with Problem Matchers, but the last one doesnt. seems verbose. either way should be consistent

docs/problem-matchers.md
```diff
@@ -0,0 +1,116 @@
...
+                }
+            ]
+        }
+    ]
+}
+```
+
+The following fields are available for problem matchers:
+
+```
```
ericsciple: a little bit weird since this isnt valid json

might be interesting to see how vscode docs describe it

thboop: They put comments in the json

https://code.visualstudio.com/docs/editor/tasks#_defining-a-problem-matcher

docs/problem-matchers.md
ericsciple: we might want to yank the docs for fromPath since there is a bug right now and it doesnt work

docs/problem-matchers.md
ericsciple: fails in nodejs: `JSON.parse('{\n// this is a comment\n"a": "b" }')`

not sure we should encourage this, binds us to a particular parser

PR #207, opened 2019-11-08
tool-cache: Support for extracting xar compatible archives

fwal:
Adds support for extracting archives compatible with eXtensible ARchiver, most notably `*.pkg` archives.


fwal:
Seems that I made an assumption that `xar` would have been installed in the ubuntu image as well 😅. Perhaps this functionality is too narrow to include in this package?

bryanmacfarlane:
Adding Eric for his thoughts.  Caution that there is a binary in this PR.

bryanmacfarlane:
I think it's fine to have OS specific variants since some tool installers expose archives in various formats per OS.  Curious, was there a specific tool / installer that exposed the tool via a xar that your were interested in writing a setup action for?  The reason I ask, is I'm up for taking the functionality as long as there's a benefit trading off the maintenance and cost going forward ...

fwal:
Sure! I use this in my [setup swift action](https://github.com/fwal/setup-swift). The feature in question is used to unpack official swift pkg archives from swift.org. I opted to just [copy the required function](https://github.com/fwal/setup-swift/blob/master/src/macos-install.ts#L60L66) in there for now.

packages/tool-cache/__tests__/tool-cache.test.ts
```diff
@@ -198,6 +199,39 @@ describe('@actions/tool-cache', function() {
       }
     })
   } else {
+    if (IS_MAC) {
+      it('extract .xar', async () => {
+        const tempDir = path.join(tempPath, 'test-install.xar')
```
thboop, line 204: Can we use the cli to generate a xar on the fly? 

I'd rather not check in binaries here.

packages/tool-cache/src/tool-cache.ts
```diff
@@ -206,6 +207,29 @@ export async function extractTar(
...
 
+/**
+ * Extract a xar compatible archive
+ *
+ * @param file     path to the archive
+ * @param dest     destination directory. Optional.
+ * @param flags    flags for the xar. Optional.
+ * @returns        path to the destination directory
+ */
+export async function extractXar(
```
thboop, line 218: Can we add an example to the [readme](https://github.com/actions/toolkit/blob/master/packages/tool-cache/README.md)?

packages/tool-cache/src/tool-cache.ts
```diff
@@ -206,6 +207,29 @@ export async function extractTar(
...
+  file: string,
+  dest?: string,
+  flags: string = '-x'
+): Promise<string> {
+  ok(IS_MAC, 'extractXar() not supported on current OS')
+  ok(file, 'parameter "file" is required')
+
+  dest = dest || (await _createExtractFolder(dest))
+  const xarPath: string = await io.which('xar', true)
+  await exec(`"${xarPath}"`, [flags, '-C', dest, '-f', file])
```
thboop, line 228: considering adding the -v (verbose) flag if debug is set.

[Example](https://github.com/actions/toolkit/blob/master/packages/tool-cache/src/tool-cache.ts#L231)

bryanmacfarlane:
... although we'd ❤️ to take the contribution from you @fwal .  Do you have time?  No rush on our side.

packages/tool-cache/__tests__/tool-cache.test.ts
bryanmacfarlane: Agreed.

fwal:
Thanks for the feedback! I’ll address everything during next week as it’s holiday here 🇸🇪
Feel free to commit if you can’t wait 🙂

fwal:
Ready for you, @thboop & @bryanmacfarlane.

packages/tool-cache/__tests__/tool-cache.test.ts
```diff
@@ -346,6 +347,40 @@ describe('@actions/tool-cache', function() {
         await io.rmRF(tempDir)
       }
     })
+  } else if (IS_MAC) {
+    it('extract .xar', async () => {
+      const tempDir = path.join(tempPath, 'test-install.xar')
+      const sourcePath = path.join(__dirname, 'data', 'archive-content')
+      const targetPath = path.join(tempDir, 'test.xar')
```
thboop, line 354: Please remove this binary, it is still committed above

fwal: Oh wow, I managed to commit another prebuilt archive when removing the first one 🤦 
Fixed now.

.gitignore
```diff
@@ -3,3 +3,4 @@ packages/*/node_modules/
 packages/*/lib/
 packages/*/__tests__/_temp/
 .DS_Store
+*.xar
```
thboop, line 6: 👍

packages/tool-cache/src/tool-cache.ts
```diff
@@ -254,6 +255,35 @@ export async function extractTar(
...
+ */
+export async function extractXar(
+  file: string,
+  dest?: string,
+  flags: string = '-x'
+): Promise<string> {
+  ok(IS_MAC, 'extractXar() not supported on current OS')
+  ok(file, 'parameter "file" is required')
+
+  dest = dest || (await _createExtractFolder(dest))
```
thboop, line 274: _createExtractFolder(dest) is a no-op if the directory already exists. The other extraction functions actually call it if a destination is passed in, so it will create the destination if it does not already exist.

This is a helpful for users, and we should do it here as well.

`dest = await _createExtractFolder(dest)`

thboop: We should also add a unit test for this functionality 
`extract .xar to a directory that does not exist`

packages/tool-cache/src/tool-cache.ts
```diff
@@ -254,6 +255,35 @@ export async function extractTar(
...
+  const args = [flags, '-C', dest, '-f', file]
+
+  if (core.isDebug() && !flags.includes('v')) {
+    args.push('-v')
+  }
```
thboop, line 279: considering the function is `extractTar`, we should also push a -x flag if it does not already exist. A unit test for this functionality would be great as well!

thboop:
Thanks for the contribution! Going to release this in `1.6.0`! https://github.com/actions/toolkit/pull/519/files

PR #211, opened 2019-11-12
Fix remove-matcher syntax

xt0rted:
Using the syntax `::remove-matcher::the-matcher` wasn't working, but after some trial and error `::remove-matcher owner=the-matcher::` did. There's no mention of this on https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions so I'm not sure if this is a feature that's changing or if it was just documented wrong.


ericsciple:
@xt0rted thanks!

PR #212, opened 2019-11-12
Update jest and lerna to fix npm install warnings

ericsciple:


ericsciple:
reviewed offline w/ bryan

Issue #218, opened 2019-11-13
exec doesn't resolve Windows .cmd/.bat files from the PATH


ericsciple:
**Describe the bug**

`exec.exec('npm --version')` does not work on Windows.

**To Reproduce**

```yaml
name: CI
on: [push]
jobs:
  build:
    runs-on: windows-latest
    steps:
      - shell: cmd
        run: |
          echo Verify npm is in the PATH:
          where npm
          echo ----------------------------------------
          echo Invoking "npm --version" works on the command line:
          call npm --version
          echo ----------------------------------------
          echo Installing @actions/exec
          call npm install @actions/exec ./
          echo ----------------------------------------
          echo Invoking "npm --version" using @actions/exec fails:
          node -e "var exec = require('@actions/exec'); exec.exec('npm --version');"
```

1138-4EB:
@ericsciple, is there any scheduled release which includes the fix to this issue?

ericsciple:
reopened since this isnt released yet

ericsciple:
@1138-4EB i'll find out today

ericsciple:
@eine sorry for the delay. Probably will be released sometime this weekend or on Monday.

ericsciple:
Workaround is specify the full path. For example:
```javascript
const npmPath = await io.which('npm', true);
await exec.exec(`"${npmPath}" --version`);
```

ericsciple:
@eine the package was published a couple days ago.

eine:
@ericsciple, thanks! It works as expected! numworks/setup-msys2#24

PR #220, opened 2019-11-14
toolrunner should which tool

ericsciple:
fixes https://github.com/actions/toolkit/issues/218

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -6,18 +6,27 @@ on:
   pull_request:
     paths-ignore:
       - '**.md'     
+
 jobs:
-  Ubuntu:
-    name: Run Ubuntu
-    runs-on: ubuntu-latest
+
+  build:
+    name: Build
+
+    strategy:
+      matrix:
+        runs-on: [ubuntu-latest, macOS-latest, windows-latest]
+      fail-fast: false
+
+    runs-on: ${{ matrix.runs-on }}
+
     steps:
     - name: Checkout
-      uses: actions/checkout@master
+      uses: actions/checkout@v1
 
-    - name: Set Node.js 10.x
+    - name: Set Node.js 12.x
       uses: actions/setup-node@master
       with:
-        node-version: 10.x
+        node-version: 12.x
 
     - name: npm install
       run: npm install
@@ -36,52 +45,3 @@ jobs:
 
     - name: Format
       run: npm run format-check
-  macOS:
-    name: Run macOS
-    runs-on: macos-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@master
-
-      - name: Set Node.js 10.x
-        uses: actions/setup-node@master
-        with:
-          node-version: 10.x
-
-      - name: npm install
-        run: npm install
-
-      - name: Bootstrap
-        run: npm run bootstrap
-
-      - name: Compile
-        run: npm run build
-
-      - name: npm test
-        run: npm test
-  Windows:
-    name: Run Windows
-    runs-on: windows-latest
-    steps:
-    - name: Checkout
-      uses: actions/checkout@master
-
-    - name: Set Node.js 10.x
-      uses: actions/setup-node@master
-      with:
-        node-version: 10.x
-
-    - name: npm install
-      run: npm install
-
-    - name: Bootstrap
-      run: npm run bootstrap
-
-    - name: Compile
-      run: npm run build
-
-    # TODO: This currently ignores exec due to issues with Node and spawning on Windows, which I think is exacerbated by Jest.
-    # It doesn't seem to affect behavior in actions themselves, just when testing with Jest.
-    # See other similar issues here: https://github.com/nodejs/node/issues/25484
-    - name: npm test
-      run: npm run test-ci
--- a/package.json
+++ b/package.json
@@ -9,8 +9,7 @@
     "format-check": "prettier --check packages/**/*.ts",
     "lint": "eslint packages/**/*.ts",
     "new-package": "scripts/create-package",
-    "test": "jest",
-    "test-ci": "jest --testPathIgnorePatterns=\"<rootDir>/packages/exec/__tests__/exec.test.ts\""
+    "test": "jest"
   },
   "devDependencies": {
     "@types/jest": "^24.0.11",
--- a/packages/exec/README.md
+++ b/packages/exec/README.md
@@ -48,13 +48,10 @@ await exec.exec('node', ['index.js', 'foo=bar'], options);
 
 #### Exec tools not in the PATH
 
-You can use it in conjunction with the `which` function from `@actions/io` to execute tools that are not in the PATH:
+You can specify the full path for tools not in the PATH:
 
 ```js
 const exec = require('@actions/exec');
-const io = require('@actions/io');
 
-const pythonPath: string = await io.which('python', true)
-
-await exec.exec(`"${pythonPath}"`, ['main.py']);
+await exec.exec('"/path/to/my-tool"', ['arg1']);
 ```
--- a/packages/exec/__tests__/exec.test.ts
+++ b/packages/exec/__tests__/exec.test.ts
@@ -121,6 +121,38 @@ describe('@actions/exec', () => {
     }
   })
 
+  it('Runs exec successfully with command from PATH', async () => {
+    const execOptions = getExecOptions()
+    const outStream = new StringStream()
+    execOptions.outStream = outStream
+    let output = ''
+    execOptions.listeners = {
+      stdout: (data: Buffer) => {
+        output += data.toString()
+      }
+    }
+
+    let exitCode = 1
+    let tool: string
+    let args: string[]
+    if (IS_WINDOWS) {
+      tool = 'cmd'
+      args = ['/c', 'echo', 'hello']
+    } else {
+      tool = 'sh'
+      args = ['-c', 'echo hello']
+    }
+
+    exitCode = await exec.exec(tool, args, execOptions)
+
+    expect(exitCode).toBe(0)
+    const rootedTool = await io.which(tool, true)
+    expect(outStream.getContents().split(os.EOL)[0]).toBe(
+      `[command]${rootedTool} ${args.join(' ')}`
+    )
+    expect(output.trim()).toBe(`hello`)
+  })
+
   it('Exec fails with error on bad call', async () => {
     const _testExecOptions = getExecOptions()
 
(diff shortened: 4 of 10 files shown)


packages/exec/src/toolrunner.ts
```diff
@@ -392,6 +393,8 @@ export class ToolRunner extends events.EventEmitter {
    * @returns   number
    */
   async exec(): Promise<number> {
+    this.toolPath = await io.which(this.toolPath, true)
```
joshmgross: Is this a breaking change?

ericsciple: my intention was that it doesnt break anything

for example `which` returns the path if it is already rooted and executable.

however, now that you got me thinking about it more, i think there are some edge cases wrt relative pathing.... i'll add code to handle.

packages/exec/README.md
```diff
@@ -48,13 +48,11 @@ await exec.exec('node', ['index.js', 'foo=bar'], options);
...
+You can specify the full path for tools not in the PATH:
 
 ```js
 const exec = require('@actions/exec');
 const io = require('@actions/io');
 
-const pythonPath: string = await io.which('python', true)
-
-await exec.exec(`"${pythonPath}"`, ['main.py']);
+await exec.exec('"<FULL-PATH-TO-TOOL>"', ['arg1']);
```
ericsciple: update with an example

packages/exec/src/toolrunner.ts
ericsciple: ok i added the code above to root unrooted paths that contain relative pathing

in addition to checking the PATH, the which function also handles checking rooted paths

now the changes shouldnt be a breaking change. hard to imagine it would have practically broken anyone, but i'd rather cover edge cases to be on the safe side. Also now the changes align better with the default behavior of `spawn`.

ericsciple: i'll add unit tests tomorrow for the unrooted-relative-pathing cases

.github/workflows/unit-tests.yml
```diff
@@ -6,18 +6,27 @@ on:
...
+
+    runs-on: ${{ matrix.runs-on }}
+
     steps:
     - name: Checkout
-      uses: actions/checkout@master
+      uses: actions/checkout@v1
 
-    - name: Set Node.js 10.x
+    - name: Set Node.js 12.x
```
ericsciple: the runner carries node 12

Thankfully this also happens to fix the ENOMEM toolrunner test issues on Windows

.github/workflows/unit-tests.yml
```diff
@@ -6,18 +6,27 @@ on:
...
 jobs:
-  Ubuntu:
-    name: Run Ubuntu
-    runs-on: ubuntu-latest
+
+  build:
+    name: Build
+
+    strategy:
+      matrix:
```
ericsciple: leverage matrix instead of duplicate 3 times

package.json
```diff
@@ -9,8 +9,7 @@
     "format-check": "prettier --check packages/**/*.ts",
     "lint": "eslint packages/**/*.ts",
     "new-package": "scripts/create-package",
-    "test": "jest",
-    "test-ci": "jest --testPathIgnorePatterns=\"<rootDir>/packages/exec/__tests__/exec.test.ts\""
```
ericsciple: we dont need this command anymore, this was to skip the failing tests on windows

packages/exec/README.md
```diff
@@ -48,13 +48,10 @@ await exec.exec('node', ['index.js', 'foo=bar'], options);
...
+You can specify the full path for tools not in the PATH:
 
 ```js
 const exec = require('@actions/exec');
-const io = require('@actions/io');
 
-const pythonPath: string = await io.which('python', true)
-
-await exec.exec(`"${pythonPath}"`, ['main.py']);
+await exec.exec('"/path/to/my-tool"', ['arg1']);
```
ericsciple: the previous example did the exact opposite of the description :(

packages/exec/__tests__/exec.test.ts
```diff
@@ -121,6 +121,38 @@ describe('@actions/exec', () => {
     }
   })
 
+  it('Runs exec successfully with command from PATH', async () => {
```
ericsciple: coverage for existing functionality

packages/exec/__tests__/exec.test.ts
```diff
@@ -418,6 +450,134 @@ describe('@actions/exec', () => {
     fs.unlinkSync(semaphorePath)
   })
 
+  it('Exec roots relative tool path using unrooted options.cwd', async () => {
```
ericsciple: tests to make sure we're not breaking existing behavior

also this confirms that extensions are resolved on windows, which the previous code did not do (but should have done)

packages/exec/__tests__/exec.test.ts
```diff
@@ -572,6 +732,42 @@ describe('@actions/exec', () => {
       )
     })
 
+    it('execs .cmd from path (Windows)', async () => {
```
ericsciple: here is the test related to the bug fix.

toolrunner should which the tool before attempting to invoke. otherwise simple commands like `npm --version` fail on windows (ENOENT) because `npm` resolves to `npm.cmd` and node doesnt find it.

packages/exec/__tests__/exec.test.ts
```diff
@@ -418,6 +450,134 @@ describe('@actions/exec', () => {
     fs.unlinkSync(semaphorePath)
   })
 
+  it('Exec roots relative tool path using unrooted options.cwd', async () => {
+    let exitCode: number
+    let command: string
+    if (IS_WINDOWS) {
+      command = './print-args-cmd' // let ToolRunner resolve the extension
```
ericsciple: note, without extension, toolrunner should resolve the extension

packages/exec/__tests__/scripts/print-args-cmd.cmd
```diff
@@ -0,0 +1,12 @@
+@echo off
```
ericsciple: the naming of these script files is a little bit weird, but it's related to the conventions set forth by the existing test case scripts

ericsciple: i just copied one of the existing cmd files (but it had spaces in the path, and i didnt want that)

this script basically just echoes the args, e.g.

```
arg[0]: "hello"
arg[1]: "world"
```

packages/exec/__tests__/scripts/print-args-sh.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+# store arguments in a special array
+args=("$@")
+# get number of elements
+ELEMENTS=${#args[@]}
+
+# echo each element
+for (( i=0;i<$ELEMENTS;i++)); do
+    echo "args[$i]: \"${args[${i}]}\""
```
ericsciple: same deal here, this script just echos the args

packages/exec/package.json
```diff
@@ -32,7 +32,7 @@
   "bugs": {
     "url": "https://github.com/actions/toolkit/issues"
   },
-  "devDependencies": {
+  "dependencies": {
```
ericsciple: need to call `io.which` and also some other io helpers

.github/workflows/unit-tests.yml
joshmgross: Creating actions docs recommends `  using: 'node12'`, do we officially support actions on node 10 and node 12? 
If so, I think it would be worth having the workflow matrix across node versions, perhaps with certain functions explicitly listed as not supported on node 10.

ericsciple: we only use node12 to run actions. I think the runner initially carried node v10, and then switched to node v12 before GA

packages/exec/src/toolrunner.ts
bryanmacfarlane: No, it's not a breaking change.  It's additive.  If you were passing in a fully qualified path, which will not changing anything.  It will however add the ability to just pass in a tool name and it just works (as well as fixing some other issues and bugs).

joshmgross: @bryanmacfarlane Originally it was breaking, exec of a relative path would no longer work if we automatically `which`. That should be fixed now with @ericsciple's latest change

packages/exec/__tests__/exec.test.ts
thboop: .cwd?

packages/exec/__tests__/exec.test.ts
thboop: Consider adding a `\\` test for windows

packages/exec/src/toolrunner.ts
```diff
@@ -392,6 +395,24 @@ export class ToolRunner extends events.EventEmitter {
...
    */
   async exec(): Promise<number> {
+    // root the tool path if it is unrooted and contains relative pathing
+    if (
+      !ioUtil.isRooted(this.toolPath) &&
+      (this.toolPath.includes('/') ||
+        (IS_WINDOWS && this.toolPath.includes('\\')))
+    ) {
+      // prefer options.cwd if it is specified, however options.cwd may also need to be rooted
+      this.toolPath = path.resolve(
```
thboop: this is super clean 🥇

thboop:
We have existing tests for exec.exec ("full path to executable"), these tests should cover the rest of the cases and fix this bug

packages/tool-cache/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@actions/tool-cache",
-	"version": "1.1.1",
+	"version": "1.1.2",
```
1138-4EB: The version of `tool-cache` was bumped, even though it seems that none of the files in `packages/tool-cache` was changed. Is this expected?

Should it mean to reflect that `tool-cache` is updated to use this new feature of `exec`, I'd expect files such as `packages/tool-cache/src/tool-cache.ts` to be modified:

https://github.com/actions/toolkit/blob/5c894298f2f24441718eb049215aa6decd507193/packages/tool-cache/src/tool-cache.ts#L175

https://github.com/actions/toolkit/blob/5c894298f2f24441718eb049215aa6decd507193/packages/tool-cache/src/tool-cache.ts#L203

https://github.com/actions/toolkit/blob/5c894298f2f24441718eb049215aa6decd507193/packages/tool-cache/src/tool-cache.ts#L239

https://github.com/actions/toolkit/blob/5c894298f2f24441718eb049215aa6decd507193/packages/tool-cache/src/tool-cache.ts#L254

Issue #199, opened 2019-10-31
TypeScript errors in `github/lib/github.d.ts`


RyanCavanaugh:
**Describe the bug**

The shipped `github.d.ts` file contains errors. The originating code is here: https://github.com/actions/toolkit/blob/master/packages/github/src/github.ts

**To Reproduce**

1. `npm init -y`
2. `tsc --init`
3. Add `a.ts`:
```ts
import * as x from "@actions/github"
```
4. `tsc`

Result:
```
node_modules/@actions/github/lib/github.d.ts:1:49 - error TS7016: Could not find a declaration file for module '@octokit/graphql'. 'D:/Throwaway/IssuesTest/node_modules/@octokit/graphql/index.js' implicitly has an 'any' type.
  Try `npm install @types/octokit__graphql` if it exists or add a new declaration (.d.ts) file containing `declare module '@octokit/graphql';`

1 import { GraphQlQueryResponse, Variables } from '@octokit/graphql';
                                                  ~~~~~~~~~~~~~~~~~~
```

**Expected behavior**

There shouldn't be errors

**Screenshots**

**Desktop (please complete the following information):**
 - OS: I'm on Windows

**Additional context**

This is a very confusing error because the next logical step for a developer would be to `npm install @octokit/graphql`, but if you do this, it just changes the error:

```
node_modules/@actions/github/lib/github.d.ts:1:10 - error TS2305: Module '"../../../../../../Throwaway/IssuesTest/node_modules/@octokit/graphql/dist-types"' has no exported member 'GraphQlQueryResponse'.

1 import { GraphQlQueryResponse, Variables } from '@octokit/graphql';
           ~~~~~~~~~~~~~~~~~~~~

node_modules/@actions/github/lib/github.d.ts:1:32 - error TS2305: Module '"../../../../../../Throwaway/IssuesTest/node_modules/@octokit/graphql/dist-types"' has no exported member 'Variables'.

1 import { GraphQlQueryResponse, Variables } from '@octokit/graphql';
                                 ~~~~~~~~~
```
It wasn't clear to me why the `.ts` file loads the checked-in `.d.ts` instead of just pulling in the real `@octokit/grapqhl` module? This pattern is guaranteed to break for `.d.ts` emit unless some other source of type data provides a definition for the module.

Two possible fixes:
 * Update https://github.com/actions/toolkit/blob/master/packages/github/src/%40types/%40octokit/index.d.ts to better match the existing `@octokit/graphql` definition 
[...]

PR #228, opened 2019-11-22
Update octokit graphql type dependencies

hross:
This fix attempts to address the issue with custom type definitions here: https://github.com/actions/toolkit/issues/199


packages/github/src/github.ts
```diff
@@ -1,5 +1,6 @@
 // Originally pulled from https://github.com/JasonEtco/actions-toolkit/blob/master/src/github.ts
-import {GraphQlQueryResponse, Variables, defaults} from '@octokit/graphql'
+import { graphql } from '@octokit/graphql'
+import { graphql as GraphQL } from '@octokit/graphql/dist-types/types'
```
hross: @thboop not sure if this is the right way to reference this type, since it's imported

hross:
@thboop looks like we could update the version we use and take dependencies on the graphql types. That seems like the right thing to do for the long term, but it might break compat for the graphql API (I'm not sure what the contract looks like over time).

I'm also not 100% on what we use this for (presumably API auth) and how to test it yet. Still researching that part. Interested in thoughts here, though.

packages/github/src/github.ts
hross: Talked to Trevor about this and this seems the only way to get to the type unless they start exporting types in the library.

packages/github/package.json
```diff
@@ -27,13 +27,15 @@
   "scripts": {
     "test": "jest",
     "build": "tsc",
+    "format": "prettier --write **/*.ts",
```
hross: I added this since we already have linting checks in the build

packages/github/package.json
```diff
@@ -27,13 +27,15 @@
...
+    "format-check": "prettier --check **/*.ts",
     "tsc": "tsc"
   },
   "bugs": {
     "url": "https://github.com/actions/toolkit/issues"
   },
   "dependencies": {
-    "@octokit/graphql": "^2.0.1",
+    "@octokit/graphql": "^4.3.1",
     "@octokit/rest": "^16.15.0"
```
thboop: Consider bumping this, we are about a year behind and there are a number of bugfixes and feature updates since then.

packages/exec/__tests__/exec.test.ts
```diff
@@ -287,8 +287,6 @@ describe('@actions/exec', () => {
   })
 
   it('Handles child process holding streams open', async function() {
-    // this was timing out on some slower hosted macOS runs at default 5s
-    jest.setTimeout(10000)
```
hross: From my testing it didn't appear these actually did anything (I think you need to set them outside the test). I moved them to the end of the method as the third arg, which works.

packages/exec/__tests__/exec.test.ts
```diff
@@ -665,7 +662,7 @@ describe('@actions/exec', () => {
         `[command]"${exePath}" myarg1 myarg2`
       )
       expect(output.trim()).toBe("args[0]: 'myarg1'\r\nargs[1]: 'myarg2'")
-    })
```
hross: there is a failing test right now in windows due to timeout, this fixes that

packages/github/src/context.ts
```diff
@@ -27,11 +27,8 @@ export class Context {
           readFileSync(process.env.GITHUB_EVENT_PATH, {encoding: 'utf8'})
         )
       } else {
-        process.stdout.write(
```
hross: fixed this b/c it was causing linting issues

thboop:
We will want to update Releases.MD before the release.

PR #235, opened 2019-12-03
Fix test timeouts

hross:
Looks like we need longer timeouts, particularly on windows


packages/exec/__tests__/exec.test.ts
```diff
@@ -332,11 +330,9 @@ describe('@actions/exec', () => {
     ).toBe(1)
 
     fs.unlinkSync(semaphorePath)
-  })
+  }, 10000) // this was timing out on some slower hosted macOS runs at default 5s
```
ericsciple: consider 20000 :)

Issue #147, opened 2019-09-19
@actions/GitHub - TypeScript Definition for WebHookPayload is limited


warrenbuckley:
Please can the `WebHookPayload` interface/type definition be updated to support more properties from GitHub Webhook/payloads such as `Push` event.

To get details about the commit/s and the files added, modified and deleted, which is useful for a linting tool to know which files to lint, as opposed to running it over an entire repository which could be quite slow if it contains lots of files.

tyankatsu0105:
@thboop
Does this repo's issue help for generating TypeScript's definition?
👀https://github.com/octokit/webhooks.js/issues/87

thboop:
I've updated the readme in #310 with information on how to do this manually if the additional type information would be helpful.

If there's a particular field that is missing that would be helpful to add, we can evaluate adding that field.

The WebhookPayload field is a union of all of these individual payloads that can run in actions from the `@octokit/webhooks` library. For example:
```
  | Webhooks.WebhookPayloadPush // If the event is "push"
  | Webhooks.WebhookPayloadPullRequest // if the event is "pull_request"
  | Webhooks.WebhookPayloadPullRequestReview // if the event is "pull_request_review"
```

These types can have overlapping fields, for example both `WebhookPayloadPullRequest` and `WebhookPayloadPullRequestReview ` contain the field `pull_request`, each of a different type`WebhookPayloadPullRequestPullRequest` and `WebhookPayloadPullRequestReviewPullRequest`. Both of these types contain different fields and information. Adding all of this information as optional fields, keeping consistent with how this is being done currently, wouldn't be much better then declaring the field as `any` as there are a large number of fields that may not even correspond to the event you are accessing.

We will evaluate making these types more accessible in a future release. I've re-tagged this as an enhancement.

tonyhallett:
See https://github.com/actions/toolkit/pull/743

PR #245, opened 2019-12-07
Add @octokit/webhooks

tyankatsu0105:
re https://github.com/actions/toolkit/issues/147

- Refactor packages/github's interface
  - Quote from @octokit/webhooks declaration

### DIFF
--- a/packages/github/src/interfaces.ts
+++ b/packages/github/src/interfaces.ts
@@ -5,7 +5,7 @@ export interface WebhookPayload {
   [key: string]: any
   repository?: Webhooks.PayloadRepository
   issue?: Webhooks.WebhookPayloadIssuesIssue
-  pull_request?: Webhooks.WebhookPayloadPullRequest
+  pull_request?: WebhookPayloadPullRequestPullRequest
   sender?: {
     [key: string]: any
     type: string
--- a/packages/github/src/interfaces.ts
+++ b/packages/github/src/interfaces.ts
@@ -5,7 +5,7 @@ export interface WebhookPayload {
   [key: string]: any
   repository?: Webhooks.PayloadRepository
   issue?: Webhooks.WebhookPayloadIssuesIssue
-  pull_request?: WebhookPayloadPullRequestPullRequest
+  pull_request?: Webhooks.WebhookPayloadPullRequest['pull_request']
   sender?: {
     [key: string]: any
     type: string
--- a/packages/github/src/interfaces.ts
+++ b/packages/github/src/interfaces.ts
@@ -10,6 +10,6 @@ export interface WebhookPayload {
     [key: string]: any
     type: string
   }
-  action: number
+  action?: string
   installation?: Webhooks.WebhookPayloadInstallationInstallation
 }
--- a/packages/github/src/interfaces.ts
+++ b/packages/github/src/interfaces.ts
@@ -1,39 +1,15 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-
-export interface PayloadRepository {
-  [key: string]: any
-  full_name?: string
-  name: string
-  owner: {
-    [key: string]: any
-    login: string
-    name?: string
-  }
-  html_url?: string
-}
+import Webhooks from '@octokit/webhooks'
 
 export interface WebhookPayload {
   [key: string]: any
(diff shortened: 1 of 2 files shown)


ericsciple:
@hross?

ericsciple:
@hross sorry at a glance i thought this was related to the change you made. it looks like this is about the github.context object instead

hross:
@ericsciple these changes are in the same spirit as my changes (importing types instead of re-defining our own). I think they are good changes that we should add but I don't have a good way of testing them. @thboop has a similar issue on his plate. Maybe worth setting up something in canary that would test the toolkit with webhooks like we did for my changes.

srb2:
thank you very much :) appreciated

thboop:
Hey @srb2, thanks for the excellent PR! 

@hross, @ericsciple  I'm going to merge this unless you have any reservations, I've tested the changes they seem good and its the correct step forward for the project.

packages/github/src/interfaces.ts
```diff
@@ -1,39 +1,15 @@
...
-  }
-  pull_request?: {
-    [key: string]: any
-    number: number
-    html_url?: string
-    body?: string
-  }
+  repository?: Webhooks.PayloadRepository
+  issue?: Webhooks.WebhookPayloadIssuesIssue
+  pull_request?: Webhooks.WebhookPayloadPullRequest
```
thboop: This should be WebhookPayloadPullRequestPullRequest

tyankatsu0105:
Umm...
This PR's author is me. @thboop

packages/github/src/interfaces.ts
tyankatsu0105: Thanks.
Done.

tyankatsu0105: There is type error

tyankatsu0105: Fix.

tyankatsu0105:
@thboop 
Bumping to v3.0.0 is great.
But I need more quote type from @octokit/webhooks.
Because I think there are some types that not support yet.
So, I suppose it's early timing when bumping version.
What do you think about this?

ericsciple:
@bryanmacfarlane thoughts?

Do you know what the original motivation was, that these types were redefined rather than re-exported?

The size concerns me. Installing `@octokit/webhooks@^7` adds 350k uncompressed, 50k compressed

There are thousands of lines of potential [WebhookPayload* type definitions](https://github.com/octokit/webhooks.js/blob/master/index.d.ts). I'm wondering about the original motivation of `@actions/github`. Why only a subset of types? Why not OR all types together? Are all types are even compatible (can they be OR'd)?

Assuming only a subset is desired (say for size consideration), at some point the solution is for the consumer to import `@octokit/webhooks` and cast to the correct type. Should we update scenario docs?

Is the motivation for bumping to v3.0.0 because `PayloadRepository` was removed? Do we need to make this breaking change? An alternative would be to re-export the interface:
```
import {PayloadRepository} from '@octokit/webhooks'
export {PayloadRepository}
```

thboop:
@tyankatsu0105 ,

I've added #310 which provides instructions on how to do this manually if needed. 

If there is a particularly field you think would be helpful to add, we can add that to the existing typescript definitions, feel free to open a PR or create an issue!

Currently, this pr sets each field to its most common definition. For example, it sets `pull_request` to `WebhookPayloadPullRequestPullRequest`. That holds true for the `pull_request` event, but the `pull_request_review` event has pull_request correspond to the type `WebhookPayloadPullRequestReviewPullRequest`. These have different fields, so the type information is inaccurate for this event.

If we are going to to import the `@octokit/webhook` definitions, we should look at importing the full payload types. The WebhookPayload field here is really just a union type between all of the possible payloads.
```
  | Webhooks.WebhookPayloadPush // If the event is "push"
  | Webhooks.WebhookPayloadPullRequest // if the event is "pull_request"
  | Webhooks.WebhookPayloadPullRequestReview // if the event is "pull_request_review"
 // ect
```

This would require bumping the major version of the module, and we don't want to push out a major version for just this at this time. I've tagged #147 with the enhancement flag and we will consider implementing it a future release.

tyankatsu0105:
@thboop 
OK. I understand.
I close this PR.

PR #246, opened 2019-12-07
versioning update

bryanmacfarlane:
update versioning guidance based off of discussions with product around the market place.   Latest major version and innovations are in master and if creating a new major version, you can create a maintenance branch for the previous release.


bryanmacfarlane:
FYI product @chrispat and @ethomson on market place and major version discussions in issues.  Also FYI @ericsciple and @hross

PR #248, opened 2019-12-09
fix extractTar on Windows

### LINKED ISSUE #194: Extract tar.gz on windows-latest failed.
**Describe the bug**
The follow code:

```js
const tc = require("@actions/tool-cache")
tc.extractTar(tarball, SOURCE_DIRECTORY)
```

Fails to run under windows-latest (other windows version not tested).

```
"C:\Program Files\Git\usr\bin\tar.exe" xz -C d:\a\install-lua\install-lua\.source\ -f d:\a\_temp\5e8814d7-8526-46bd-bd88-ed84e58259a1
tar (child): Cannot connect to d: resolve failed

gzip: stdin: unexpected end of file
/usr/bin/tar: Child returned status 128
/usr/bin/tar: Error is not recoverable: exiting now
##[error]Failed to install Lua: Error: The process 'C:\Program Files\Git\usr\bin\tar.exe' failed with exit code 2
##[error]Node run failed with exit code 1
```

**To Reproduce**
1. Use extractTar in actions scripts.
2. Run the action on windows latest

**Expected behavior**

Show extractTar correctly on windows.

ericsciple:
Fixes #194

### DIFF
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -202,17 +202,20 @@ export async function extractTar(
   dest = dest || (await _createExtractFolder(dest))
   let tarPath: string
   if (IS_WINDOWS) {
-    // Do not which tar on Windows due to incompatibilities between BSD tar and GNU tar on Windows.
-    // The version of tar at C:\Windows\System32\tar.exe is BSD tar and works properly with
-    // Windows-style paths. Whereas GNU tar requires forward slashes and also requires the flag
-    // `--force-local` due to the `:` in Windows paths. This presents an incompatibility because
-    // BSD tar does not support the flag `--force-local` (invalid flag, fails the process).
-    tarPath = path.join(process.env['windir'] as string, 'System32', 'tar.exe')
-    fs.statSync(tarPath)
+    // Use 7z
+    await extract7z(file, dest);
+
+    // // Do not which tar on Windows due to incompatibilities between BSD tar and GNU tar on Windows.
+    // // The version of tar at C:\Windows\System32\tar.exe is BSD tar and works properly with
+    // // Windows-style paths. Whereas GNU tar requires forward slashes and also requires the flag
+    // // `--force-local` due to the `:` in Windows paths. This presents an incompatibility because
+    // // BSD tar does not support the flag `--force-local` (invalid flag, fails the process).
+    // tarPath = path.join(process.env['windir'] as string, 'System32', 'tar.exe')
+    // fs.statSync(tarPath)
   } else {
     tarPath = await io.which('tar', true)
+    await exec(`"${tarPath}"`, [flags, '-C', dest, '-f', file])
   }
-  await exec(`"${tarPath}"`, [flags, '-C', dest, '-f', file])
 
   return dest
 }
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -209,8 +209,7 @@ export async function extractTar(
     // BSD tar does not support the flag `--force-local` (invalid flag, fails the process).
     tarPath = path.join(process.env['windir'] as string, 'System32', 'tar.exe')
     fs.statSync(tarPath)
-  }
-  else {
+  } else {
     tarPath = await io.which('tar', true)
   }
   await exec(`"${tarPath}"`, [flags, '-C', dest, '-f', file])
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -200,7 +200,14 @@ export async function extractTar(
   }
 
   dest = dest || (await _createExtractFolder(dest))
-  const tarPath: string = await io.which('tar', true)
+  let tarPath: string
+  if (IS_WINDOWS) {
+    tarPath = path.join(process.env['windir'] as string, 'System32', 'tar.exe')
+    fs.statSync(tarPath)
+  }
+  else {
+    tarPath = await io.which('tar', true)
+  }
   await exec(`"${tarPath}"`, [flags, '-C', dest, '-f', file])
 
   return dest
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -202,6 +202,11 @@ export async function extractTar(
   dest = dest || (await _createExtractFolder(dest))
   let tarPath: string
   if (IS_WINDOWS) {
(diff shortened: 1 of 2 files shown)


packages/tool-cache/src/tool-cache.ts
```diff
@@ -200,7 +200,18 @@ export async function extractTar(
...
-  const tarPath: string = await io.which('tar', true)
+  let tarPath: string
+  if (IS_WINDOWS) {
+    // Do not which tar on Windows due to incompatibilities between BSD tar and GNU tar on Windows.
+    // The version of tar at C:\Windows\System32\tar.exe is BSD tar and works properly with
+    // Windows-style paths. Whereas GNU tar requires forward slashes and also requires the flag
+    // `--force-local` due to the `:` in Windows paths. This presents an incompatibility because
+    // BSD tar does not support the flag `--force-local` (invalid flag, fails the process).
+    tarPath = path.join(process.env['windir'] as string, 'System32', 'tar.exe')
+    fs.statSync(tarPath)
```
joshmgross: What's the purpose of `fs.StatSync(tarPath)`? To throw if tar doesn't exist?

joshmgross:
Did you verify that BSD tar on Windows doesn't have issues with `\` and `:`?
I plan on switching the cache action to use BSD tar as well, thanks for this!

packages/tool-cache/src/tool-cache.ts
```diff
@@ -200,7 +200,18 @@ export async function extractTar(
...
   dest = dest || (await _createExtractFolder(dest))
-  const tarPath: string = await io.which('tar', true)
+  let tarPath: string
+  if (IS_WINDOWS) {
+    // Do not which tar on Windows due to incompatibilities between BSD tar and GNU tar on Windows.
+    // The version of tar at C:\Windows\System32\tar.exe is BSD tar and works properly with
+    // Windows-style paths. Whereas GNU tar requires forward slashes and also requires the flag
+    // `--force-local` due to the `:` in Windows paths. This presents an incompatibility because
+    // BSD tar does not support the flag `--force-local` (invalid flag, fails the process).
+    tarPath = path.join(process.env['windir'] as string, 'System32', 'tar.exe')
```
thboop: Tar seems to have been added as a part of the initiative to bring more Unix tools to windows in Windows 10. It looks liked our only hosted version of windows is now 2019, and it exists on that machine for hosted builds

However, for self hosted runners do we know if this exists on all the platforms that can run .net core 3.0? Those go back all the way to Windows Server 2012. https://github.com/dotnet/core/blob/master/release-notes/3.0/3.0-supported-os.md

For 7zip, we ship a copy of the binary and and use it if needed, we may need to look into a similar solution here.
https://github.com/actions/toolkit/blob/d823f7598797eef5514afdc84f8ef51e293091f0/packages/tool-cache/src/tool-cache.ts#L153

thboop: Should we attempt to detect if the path doesn't exist so we can log a better message in this case if not all windows versions will have it installed

packages/tool-cache/src/tool-cache.ts
hross: should probably use `existsSync` instead of this mechanism, if that's the case...

packages/tool-cache/src/tool-cache.ts
hross: Looks like maybe using existsSync and handling the false case would be the best way to do this. I wouldn't attempt to ship a binary in this case... maybe just fail and explain why.

packages/tool-cache/src/tool-cache.ts
ericsciple: yes, i'll imrpove the error message

thboop: Is this missing `.isFile()`

packages/tool-cache/src/tool-cache.ts
ericsciple: +1 regarding the error message, at a glance it looked good to me but looking again we can make it a little bit more clear by try/catch/throw a better error.

one problem with the 7z approach is this function is literally designed as a wrapper over `tar` (see the `flags` parameter`)

thboop: Agreed on using 7zip isn't the correct solution here. I don't really like shipping an exe of tar with the package like we do with 7zip either.

I'm happy with better error message (this may also affect users using container actions because iirc we support windows containers if the host system is linux)

packages/tool-cache/src/tool-cache.ts
ericsciple: actually tool runner provides a good error message so i'll just remove this line

packages/tool-cache/src/tool-cache.ts
ericsciple: actually tool runner gives a good error already so i'll remove the `statSync` line

ericsciple: looking into 7z. note we dont package the full 7z (over 1mb) just the decoder exe (40kb)

ericsciple: i talked w/ josh and basically the cache action needs the support

ericsciple: so will need broad os support.

i still think the current proposal is incrementally better

joshmgross: Re: Hosted Runners vs Self-Hosted Runners
The current behavior is that we use GNU specific tar option on windows (specifically `--force-local`) since the GNU tar is prepended to the PATH.
This works for hosted runners, but self-hosted runners most likely do not have GNU prepended, so they will either fail with tar not being found or BSD tar not recognizing `--force-local`), see https://github.com/actions/cache/issues/91

With this solution, at least self-hosted runners that have tar are supported.

Also worth noting the cache action does not consume this package, it just has the same problems with tar.

ericsciple:
@thboop @hross @joshmgross thoughts?

ok so 7zdec doesnt work for .tar.gz and .tar.xz

other solutions on Windows that come to mind are:
1. check `tar --version` and grep the output to determine whether its bsdtar gnutar. I don't like this solution because it puts the burden on the end user (not the actual author) to verify tar is in the PATH

2. create a totally separate npm package like `windowsTar` if someone wants tar support on windows.

ericsciple:
honestly at this point i think i'm going to close this PR and open an ADR.

I dont feel good anymore about the half solution i proposed with the ADR.

I would rather come up with a proposal that allows us to fix it once and for all, and never revisit the problem again.

PR #257, opened 2019-12-16
overload downloadTool to accept destination path

### LINKED ISSUE #60: @actions/tool-cache downloadTool does not save file with file extension
I am trying to write a GitHub action tool to download & setup Nuget.exe however by passing the URL

`const nugetPath = await tc.downloadTool("https://dist.nuget.org/win-x86-commandline/latest/nuget.exe");`

This saves the file as something like this
`temp/b438b0f2-af22-456a-8a77-ddc4152465b9` minus the .exe file extension

I could workaround it by manually renaming it with using NodeJS fs APIs but this feels a bit wrong to me.

ericsciple:
fixes #60

### DIFF
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -50,7 +50,10 @@ if (!tempDirectory || !cacheRoot) {
  * @param url       url of tool to download
  * @returns         path to downloaded tool
  */
-export async function downloadTool(url: string): Promise<string> {
+export async function downloadTool(
+  url: string,
+  dest?: string
+): Promise<string> {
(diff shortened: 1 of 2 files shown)


packages/tool-cache/src/tool-cache.ts
```diff
@@ -50,22 +50,24 @@ if (!tempDirectory || !cacheRoot) {
  * @param url       url of tool to download
  * @returns         path to downloaded tool
```
TingluoHuang: update the `@param`?

ericsciple: nice catch

abirhasan132 (commented):
152e3137c2075505311d5ae6fabee1aea672d6b3

PR #263, opened 2019-12-17
remove trailing comma from commands

### LINKED ISSUE #208: Trailing comma in `::set-output``
I'm not sure if this is relevant, but I noticed that the log shows the meta-command

```
::set-output name=cache-hit,::true
```

being emitted, whereas based on Github actions documentation I would have expected

```
::set-output name=cache-hit::true
```

Sorry for the noise if this is in fact harmless.

ericsciple:
fixes #208


packages/core/__tests__/command.test.ts
```diff
@@ -0,0 +1,79 @@
+import * as command from '../src/command'
+import * as os from 'os'
+
+/* eslint-disable @typescript-eslint/unbound-method */
```
ericsciple: this is so beforeAll can reference `originalWriteFunction`. note, `core.test.ts` does the same thing so there was a precedent for disabling this already

packages/core/__tests__/core.test.ts
```diff
@@ -37,21 +37,21 @@ describe('@actions/core', () => {
 
   it('exportVariable produces the correct command and sets the env', () => {
     core.exportVariable('my var', 'var val')
-    assertWriteCalls([`::set-env name=my var,::var val${os.EOL}`])
+    assertWriteCalls([`::set-env name=my var::var val${os.EOL}`])
```
bryanmacfarlane: maybe I missed it but do we have any tests that validate multiple properties (and thus does write the comma)?

ericsciple: yep command.test.ts is new and tests several combinations

mies386:
#110##106.patch#master_user

PR #264, opened 2019-12-17
fix extractTar on Windows

ericsciple:


packages/tool-cache/__tests__/tool-cache.test.ts
```diff
@@ -268,96 +268,88 @@ describe('@actions/tool-cache', function() {
         await io.rmRF(tempDir)
```
ericsciple: i had to recreate test.tar.gz with the option `--format pax` otherwise bsdtar doesnt preserve UTF file names correctly

ericsciple: i added a README.txt to the archive for future reference

packages/tool-cache/__tests__/tool-cache.test.ts
```diff
@@ -268,96 +268,88 @@ describe('@actions/tool-cache', function() {
...
       }
     })
-  } else {
-    it('extract .tar.gz', async () => {
-      const tempDir = path.join(tempPath, 'test-install-tar.gz')
+  }
 
-      await io.mkdirP(tempDir)
+  it('extract .tar.gz', async () => {
```
ericsciple: no changes here, just eliminated the `else` condition and de-indented

packages/tool-cache/src/tool-cache.ts
```diff
@@ -202,9 +202,36 @@ export async function extractTar(
...
+  })
+  const isGnuTar = versionOutput.toUpperCase().includes('GNU TAR')
+
+  // Initialize args
+  const args = [flags]
+
+  let destArg = dest
+  if (IS_WINDOWS && isGnuTar) {
+    args.push('--force-local')
+    destArg = dest.replace(/\\/g, '/')
```
joshmgross: Both `destArg` and `file` need the `\` replacement

packages/tool-cache/src/tool-cache.ts
```diff
@@ -202,9 +202,36 @@ export async function extractTar(
     throw new Error("parameter 'file' is required")
   }
 
+  // Create dest
   dest = await _createExtractFolder(dest)
-  const tarPath: string = await io.which('tar', true)
-  await exec(`"${tarPath}"`, [flags, '-C', dest, '-f', file])
+
+  // Determine whther GNU tar
```
joshmgross:
```suggestion
  // Determine whether GNU tar
```

packages/tool-cache/src/tool-cache.ts
ericsciple: I can add for file too, but when i tested `file` doesnt appear to be required

ericsciple: i'll push this change, aesthetically it will look good for consistency

bryanmacfarlane:
Add a description for the PR ☝️

bryanmacfarlane:
Other than description, LGTM

packages/tool-cache/src/tool-cache.ts
```diff
@@ -202,9 +202,36 @@ export async function extractTar(
...
+  // Determine whether GNU tar
+  let versionOutput = ''
+  await exec('tar --version', [], {
+    ignoreReturnCode: true,
+    listeners: {
+      stdout: (data: Buffer) => (versionOutput += data.toString()),
+      stderr: (data: Buffer) => (versionOutput += data.toString())
+    }
+  })
+  const isGnuTar = versionOutput.toUpperCase().includes('GNU TAR')
```
thboop: Consider debugging this so we can troubleshoot easier if for some reason this breaks. (Analyzing text output of a cli can be a little fragile)

thboop: We may also want to debug/warn if we don't find gnu tar on windows that we would recommend that you setup your self hosted runner with gnu tar on windows in the path.

ericsciple: the output flows to the console too, so we're good wrt that aspect

PR #268, opened 2019-12-20
glob

ericsciple:
New module for globbing

### DIFF
--- a/packages/glob/package.json
+++ b/packages/glob/package.json
@@ -1,0 +1,41 @@
+{
+  "name": "@actions/glob",
+  "version": "0.1.0",
+  "preview": true,
+  "description": "Actions glob lib",
+  "keywords": [
+    "github",
+    "actions",
+    "glob"
+  ],
+  "homepage": "https://github.com/actions/toolkit/tree/master/packages/glob",
+  "license": "MIT",
+  "main": "lib/glob.js",
+  "types": "lib/glob.d.ts",
+  "directories": {
+    "lib": "lib",
+    "test": "__tests__"
+  },
+  "files": [
+    "lib"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/actions/toolkit.git",
+    "directory": "packages/glob"
+  },
+  "scripts": {
+    "test": "echo \"Error: run tests from root\" && exit 1",
+    "tsc": "tsc"
+  },
+  "bugs": {
+    "url": "https://github.com/actions/toolkit/issues"
+  },
+  "dependencies": {
+    "@actions/core": "^1.2.0",
+    "minimatch": "^3.0.4"
+  }
+}
--- a/packages/glob/src/glob.ts
+++ b/packages/glob/src/glob.ts
@@ -146,9 +146,7 @@ async function stat(
         }
 
         throw new Error(
-          `No information found for the path '${
-            item.path
-          }'. This may indicate a broken symbolic link.`
+          `No information found for the path '${item.path}'. This may indicate a broken symbolic link.`
         )
       }
 
@@ -172,9 +170,7 @@ async function stat(
     // Test for a cycle
     if (traversalChain.some((x: string) => x === realPath)) {
       core.debug(
-        `Symlink cycle detected for path '${
-          item.path
-        }' and realpath '${realPath}'`
+        `Symlink cycle detected for path '${item.path}' and realpath '${realPath}'`
       )
       return undefined
     }
--- a/packages/glob/src/glob.ts
+++ b/packages/glob/src/glob.ts
@@ -1,0 +1,187 @@
+import * as core from '@actions/core'
+import * as fs from 'fs'
+import * as path from 'path'
+import * as patternHelper from './internal-pattern-helper'
+import {IGlobOptions} from './internal-glob-options'
+import {MatchKind} from './internal-match-kind'
+import {Pattern} from './internal-pattern'
+import {SearchState} from './internal-search-state'
+
+export {IGlobOptions}
+
+/**
+ * Returns files and directories matching the specified glob pattern.
+ *
+ * Order of the results is not guaranteed.
+ */
+export async function glob(
+  pattern: string,
+  options?: IGlobOptions
+): Promise<string[]> {
+  const result: string[] = []
+  for await (const itemPath of globGenerator(pattern, options)) {
+    result.push(itemPath)
+  }
+  return result
+}
+
+/**
+ * Returns files and directories matching the specified glob pattern.
+ *
+ * Order of the results is not guaranteed.
+ */
+export async function* globGenerator(
+  pattern: string,
+  options?: IGlobOptions
+): AsyncGenerator<string, void> {
+  // Set defaults options
+  options = patternHelper.getOptions(options)
+
+  // Parse patterns
+  const patterns: Pattern[] = patternHelper.parse([pattern], options)
+
+  // Push the search paths
+  const stack: SearchState[] = []
+  for (const searchPath of patternHelper.getSearchPaths(patterns)) {
+    core.debug(`Search path '${searchPath}'`)
+
+    // Exists?
+    try {
+      // Intentionally using lstat. Detection for broken symlink
+      // will be performed later (if following symlinks).
+      await fs.promises.lstat(searchPath)
+    } catch (err) {
+      if (err.code === 'ENOENT') {
+        continue
+      }
+      throw err
+    }
+
+    stack.unshift(new SearchState(searchPath, 1))
+  }
+
+  // Search
+  const traversalChain: string[] = [] // used to detect cycles
+  while (stack.length) {
+    // Pop
+    const item = stack.pop() as SearchState
+
+    // Match?
+    const match = patternHelper.match(patterns, item.path)
+    const partialMatch =
+      !!match || patternHelper.partialMatch(patterns, item.path)
+    if (!match && !partialMatch) {
+      continue
+    }
+
+    // Stat
+    const stats: fs.Stats | undefined = await stat(
+      item,
+      options,
+      traversalChain
+    )
+
+    // Broken symlink, or symlink cycle detected, or no longer exists
+    if (!stats) {
+      continue
+    }
+
+    // Directory
+    if (stats.isDirectory()) {
+      // Matched
+      if (match & MatchKind.Directory) {
+        yield item.path
+      }
+      // Descend?
+      else if (!partialMatch) {
+        continue
+      }
+
+      // Push the child items in reverse
+      const childLevel = item.level + 1
+      const childItems = (await fs.promises.readdir(item.path)).map(
+        x => new SearchState(path.join(item.path, x), childLevel)
+      )
+      stack.push(...childItems.reverse())
+    }
+    // File
+    else if (match & MatchKind.File) {
+      yield item.path
+    }
+  }
+}
+
+/**
+ * Returns the search path preceeding the first segment that contains a pattern.
+ *
+ * For example, '/foo/bar*' returns '/foo'.
+ */
+export function getSearchPath(pattern: string): string {
+  const patterns: Pattern[] = patternHelper.parse(
+    [pattern],
+    patternHelper.getOptions()
+  )
+  const searchPaths: string[] = patternHelper.getSearchPaths(patterns)
+  return searchPaths.length > 0 ? searchPaths[0] : ''
+}
+
+async function stat(
+  item: SearchState,
+  options: IGlobOptions,
+  traversalChain: string[]
+): Promise<fs.Stats | undefined> {
+  // Note:
+  // `stat` returns info about the target of a symlink (or symlink chain)
+  // `lstat` returns info about a symlink itself
+  let stats: fs.Stats
+  if (options.followSymbolicLinks) {
+    try {
+      // Use `stat` (following symlinks)
+      stats = await fs.promises.stat(item.path)
+    } catch (err) {
+      if (err.code === 'ENOENT') {
+        if (options.omitBrokenSymbolicLinks) {
+          core.debug(`Broken symlink '${item.path}'`)
+          return undefined
+        }
+
+        throw new Error(
+          `No information found for the path '${
+            item.path
+          }'. This may indicate a broken symbolic link.`
+        )
+      }
+
+      throw err
+    }
+  } else {
+    // Use `lstat` (not following symlinks)
+    stats = await fs.promises.lstat(item.path)
+  }
+
+  // Note, isDirectory() returns false for the lstat of a symlink
+  if (stats.isDirectory() && options.followSymbolicLinks) {
+    // Get the realpath
+    const realPath: string = await fs.promises.realpath(item.path)
+
+    // Fixup the traversal chain to match the item level
+    while (traversalChain.length >= item.level) {
+      traversalChain.pop()
+    }
+
+    // Test for a cycle
+    if (traversalChain.some((x: string) => x === realPath)) {
+      core.debug(
+        `Symlink cycle detected for path '${
+          item.path
+        }' and realpath '${realPath}'`
+      )
+      return undefined
+    }
+
+    // Update the traversal chain
+    traversalChain.push(realPath)
+  }
+
+  return stats
+}
--- a/packages/glob/src/internal-glob-options.ts
+++ b/packages/glob/src/internal-glob-options.ts
@@ -1,0 +1,28 @@
+export interface IGlobOptions {
+  /**
+   * Indicates whether to follow symbolic links. Generally should be true
+   * unless deleting files.
+   *
+   * @default true
+   */
+  followSymbolicLinks?: boolean
+
+  /**
+   * Indicates whether directories that match a glob pattern, should implicitly
+   * cause all descendant paths to be matched.
+   *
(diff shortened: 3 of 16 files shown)


package.json
```diff
@@ -8,7 +8,6 @@
     "format": "prettier --write packages/**/*.ts",
     "format-check": "prettier --check packages/**/*.ts",
     "lint": "eslint packages/**/*.ts",
-    "new-package": "scripts/create-package",
```
ericsciple: undo

packages/glob/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@actions/glob",
+  "version": "1.0.0",
```
ericsciple: probably start with `0.1`?

packages/glob/src/internal-glob-options.ts
```diff
@@ -0,0 +1,28 @@
+export interface IGlobOptions {
```
ericsciple: I need to update the ADR, this diverges a little bit

The ADR specifies a class, I ended up making an interface instead since optional properties provides a better experience for consumers.

packages/glob/src/internal-glob-options.ts
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * Indicates whether directories that match a glob pattern, should implicitly
+   * cause all descendant paths to be matched.
+   *
+   * For example, given the directory `my-dir`, the following glob patterns
+   * would produce the same results: `my-dir/**`, `my-dir/`, `my-dir`
+   *
+   * @default true
+   */
+  implicitDescendants?: boolean
```
ericsciple: I need to update the ADR. When I started thinking about support for multiple patterns (and negate patterns) I came to the conclusion `implicitDescendants` is a better name instead of `expandDirectories`.

The new name aligns better with how the property influences negate patterns.

packages/glob/src/internal-pattern.ts
```diff
@@ -0,0 +1,227 @@
...
+    // Negate
+    while (pattern.startsWith('!')) {
+      this.negate = !this.negate
+      pattern = pattern.substr(1)
+    }
+
+    // Empty
+    assert(pattern, 'pattern cannot be empty')
+
+    // On Windows, do not allow paths like C: and C:foo (for simplicity)
```
ericsciple: Update ADR with this constraint

packages/glob/src/internal-pattern.ts
```diff
@@ -0,0 +1,227 @@
...
+      IS_WINDOWS ? 'i' : ''
+    )
+
+    // Set the root regexp
+    this.rootRegExp = new RegExp(
+      this.regExpEscape(this.convertToLiteral(patternPath.segments[0]))
+    )
+  }
+
+  /**
```
ericsciple: This function is required to accurately determine the search path. Can't simply use information from minimatch (whether a segment \"has magic\") because users sometimes need to use `[ ]` for escaping on Windows.

scripts/create-package
```diff
@@ -1,13 +0,0 @@
-#!/usr/bin/env bash
```
ericsciple: undelete

packages/glob/src/internal-pattern.ts
```diff
@@ -0,0 +1,227 @@
...
+  negate: boolean = false
+  searchPath: string
+  private minimatch: IMinimatch
+  private rootRegExp: RegExp
+  private trailingSlash: boolean = false
+
+  constructor(pattern: string) {
+    pattern = pattern || ''
+
+    // Comment
```
ericsciple: update ADR to reserve leading `#`

packages/glob/src/internal-pattern-helper.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  return result
+}
+
+/**
+ * Matches the patterns against the path
+ */
+export function match(patterns: Pattern[], itemPath: string): MatchResult {
```
ericsciple: internally supports multiple patterns and negate patterns

the public interface only supports passing a single pattern, but basically will be a one-liner to fix that when needed

packages/glob/src/internal-path-helper.ts
```diff
@@ -0,0 +1,144 @@
...
+  assert(p, `ensureRooted parameter 'p' cannot be empty`)
+
+  // Already rooted
+  if (isRooted(p)) {
+    return p
+  }
+
+  // On Windows, check for root like C:
+  if (IS_WINDOWS && root.match(/^[A-Z]:$/i)) {
+    return root + p
```
joshmgross: Should these path creations be using `path.join`? https://nodejs.org/api/path.html#path_path_join_paths

If `root` does not end in a trailing separator (`C:`), won't you hit issues if `path` doesn't begin with a separator?

packages/glob/src/internal-path-helper.ts
```diff
@@ -0,0 +1,144 @@
+import * as assert from 'assert'
+import * as path from 'path'
+
+const IS_WINDOWS = process.platform === 'win32'
+
```
joshmgross: Why not use the `path` methods here instead?

There's `dirname`, `isAbsolute`, and `normalize` that have similar purposes to what's implemented here

packages/glob/src/internal-pattern-helper.ts
```diff
@@ -0,0 +1,142 @@
...
+ */
+export function getSearchPaths(patterns: Pattern[]): string[] {
+  // Ignore comment and negate patterns
+  patterns = patterns.filter(x => !x.comment && !x.negate)
+
+  // Create a map of all search paths
+  const searchPathMap: {[key: string]: string} = {}
+  for (const pattern of patterns) {
+    const key = IS_WINDOWS
+      ? pattern.searchPath.toUpperCase()
```
joshmgross: Is this to ensure we get case insensitive matching on Windows?

packages/glob/src/internal-pattern.ts
```diff
@@ -0,0 +1,227 @@
...
+    if (pattern.startsWith('#')) {
+      this.comment = true
+      this.searchPath = (undefined as unknown) as string
```
joshmgross: Why not allow `searchPath` and the below to be nullable types?

packages/glob/src/internal-pattern-helper.ts
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Parses the pattern strings into Pattern objects
+ */
+export function parse(patterns: string[], options: IGlobOptions): Pattern[] {
+  const result: Pattern[] = []
+
+  for (let patternString of patterns) {
+    // Parse
+    const pattern = new Pattern(patternString)
```
joshmgross: Seems like it would be better to check if the `pattern` is a comment before creating the `Pattern` instance, that way the class itself doesn't have to worry about comments and you can remove some of the extra checks in `internal-pattern.ts`

packages/glob/src/internal-pattern-helper.ts
```diff
@@ -0,0 +1,142 @@
...
+    // Comment
+    if (pattern.comment) {
+      continue
+    }
+
+    // Push
+    result.push(pattern)
+
+    // Implicitly match descendant paths
+    if (options.implicitDescendants && !endsWithGlobstar(patternString)) {
```
joshmgross: What if my `pattern` is `dir/*`?
It fails the `endsWithGlobstar` check and you'd end up with the pattern `dir/*/**`

Is that still a valid pattern and will it achieve the desired result? (Desired result being that users don't need to specify an exact globstar pattern to match directories)

Ideal behavior might be that `dir/*` is converted into `dir/**` for the user.

packages/glob/src/internal-search-state.ts
```diff
@@ -0,0 +1,9 @@
+export class SearchState {
```
joshmgross: Can this just be an interface?

packages/glob/src/glob.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  // Get search paths
+  const searchPaths: string[] = patternHelper.getSearchPaths(patterns)
+
+  // Search
+  const result: string[] = []
+  for (const searchPath of searchPaths) {
+    // Skip if not exists
+    try {
+      await fs.promises.lstat(searchPath)
```
joshmgross: Is there an advantage to this over using `fs.existsSync(searchPath)`?

If so, is it still valid to check the symbolic link (`lstat`) of the path instead of the file itself (`stat`)?

packages/glob/src/glob.ts
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    // Update the traversal chain
+    traversalChain.push(realPath)
+  }
+
+  return stats
+}
+
+function getOptions(copy?: IGlobOptions): IGlobOptions {
```
joshmgross: Why not pass the glob options directly to the functions instead of copying?

packages/glob/src/internal-search-state.ts
ericsciple: its internal, doesnt really matter either way. i thought the constructor was slightly nicer to work with

ericsciple: let me know if i'm overlooking something else. seems like arbitrary preference either way

packages/glob/src/glob.ts
ericsciple: The properties are optional and the defaults need to be set when undefined. As an SDK function, it shouldnt mutate the object passed in by the caller. Hence the copy.

packages/glob/src/internal-pattern-helper.ts
ericsciple: i like this way for encapsulation (parsing is the responsibility of the Pattern class)

ericsciple: the internal-pattern-helper.parse function never actually returns the comment patterns to the caller, so most of the \"is comment\" filters are not actually required. i just put them there so it's clear.

ericsciple: unresolve if you disagree

packages/glob/src/internal-pattern-helper.ts
ericsciple: `dir/*/**` and `dir/**` are not the same and (the format does not return `dir` and the latter does)

`dir/*/**` is the desired result

packages/glob/src/internal-pattern.ts
ericsciple: i'll change to empty string. Not sure if i'm doing something wrong, but I just find the not-null constraints of typescript annoying rather than helpful. If it's nullable then the compiler requires i use the nullable deference everywhere (`?.`) which i then have to coalesce even when i know it's never null. Which just makes the code less readable, and production behavior less clear.

packages/glob/src/internal-pattern-helper.ts
ericsciple: yes

ericsciple: note case is preserved for the search path, this just helps filter out unnecessary search paths on windows (ones that exist under a higher search path)

packages/glob/src/glob.ts
ericsciple: i'll add a comment to the code to explain this. not sure if i even need this anymore actually

packages/glob/src/internal-path-helper.ts
ericsciple: `dirname` and `isAbsolute` have quirks and for what i'm doing i need very precise behavior on Windows.

For example `dirname` usually removes unnecessary trailing slashes, but doesn't always for UNC paths (doesn't for `\\server\share\foo` but does for `\\server\share\`)

`isAbsolute` correctly returns false for `C:`. Whereas I need the answer to a different question: is it rooted? (yes for `C:`)

`normalize` also removes relative pathing. That's not the behavior I desire, i just want slashes normalized.

packages/glob/src/internal-path-helper.ts
ericsciple: `path.join('C:', 'foo')` returns `C:\foo` which is wrong

https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats

packages/glob/src/glob.ts
ericsciple: i added a comment to the code. But we still need this `lstat` call here because `stat` is used later to determine whether the symlink is broken

ericsciple: i think i can rearrange things a bit to make this more clean

packages/glob/src/internal-pattern-helper.ts
joshmgross: A comment isn't a `Pattern` though, so it seems weird to make it one. Maybe a factory or function to create Patterns that can also return `undefined` for comments would be a good middle-ground, and then you could get rid of the nullable/unknown fields in `Pattern` that are required for only comments.

packages/glob/src/internal-pattern-helper.ts
joshmgross: What does the input `dir/*` mean then as a pattern?

packages/glob/src/internal-search-state.ts
joshmgross: It's preference, I typically just use interfaces when I only have data and no functions, as then it's just a type definition on an object.

packages/glob/src/glob.ts
ericsciple: rearranged to eliminate some unnecessary stat calls (for files that arent matched)

also i eliminated some unnecessary nesting of loops by pre-populating the search stack with all search paths

packages/glob/src/internal-pattern-helper.ts
ericsciple: > A comment isn't a Pattern

sold. refactored and i like it much better. thanks

packages/glob/src/internal-pattern.ts
ericsciple: now that comment isnt a property this problem went away (as you suggested on the other comment)

ericsciple: good call btw

packages/glob/src/internal-pattern-helper.ts
ericsciple: `dir/*` means all children of dir

ericsciple: `shopt -s globstar` and `echo` is helpful to test different globs

packages/glob/src/internal-pattern.ts
```diff
@@ -0,0 +1,266 @@
...
+    // Must not contain globs in root, e.g. \\foo\b*
+    assert(
+      !pathHelper.isRooted(pattern) || literalSegments[0],
+      `Invalid pattern '${pattern}'. Root segment must not contain globs.`
+    )
+
+    // Normalize slashes
+    pattern = pathHelper.normalizeSeparators(pattern)
+
+    // Replace leading `.` segment
```
ericsciple: also replace leading `~/`

packages/glob/src/internal-pattern-helper.ts
```diff
@@ -0,0 +1,142 @@
...
+}
+
+/**
+ * Matches the patterns against the path
+ */
+export function match(patterns: Pattern[], itemPath: string): MatchKind {
+  let result: MatchKind = MatchKind.None
+
+  for (const pattern of patterns) {
+    if (pattern.negate) {
```
TingluoHuang: should we filter negate out since we don't support it yet?

ericsciple: yes we need to reserve the behavior now, so its not a breaking change later

joshmgross:
Should a section be added to the main README for this new package? https://github.com/actions/toolkit/blob/master/README.md#packages

ericsciple:
yes it's tracked as a consequence in the ADR. havent done it yet though, still working on unit tests. I might do docs as a separate PR

packages/glob/src/internal-pattern.ts
ericsciple: i went the other way and resolve to an absolute rooted path

packages/glob/package.json
bryanmacfarlane: Should this be in core?

bryanmacfarlane: I think size would be the only consideration - core folks (everyone) that doesn't need glob.  The other way around isn't a scenario IMHO (need glob but not core)

packages/glob/src/glob.ts
```diff
@@ -0,0 +1,179 @@
...
+): AsyncGenerator<string, void> {
+  // Set defaults options
+  options = patternHelper.getOptions(options)
+
+  // Parse patterns
+  const patterns: Pattern[] = patternHelper.parse([pattern], options)
+
+  // Push the search paths
+  const stack: SearchState[] = []
+  for (const searchPath of patternHelper.getSearchPaths(patterns)) {
```
TingluoHuang: Can we trace out `searchPath` for debugging, since we might compute the search path base on multiple patterns?

packages/glob/package.json
ericsciple: Open to whatever, here's the factors i considered:
- A separate module allows independent versioning. E.g. can start out with `0.1-preview`
- It also aligns with npm philosophy, small targeted packages.
- Size when installing minimatch on a fresh directory: 80kb uncompressed, 20kb compressed 

Would this align more with core, or with io?

bryanmacfarlane:
@ericsciple, ready for a release?

PR #275, opened 2019-12-26
Only run CI on PRs and pushes to master

joshmgross:
Pull requests currently run every workflow twice, on the push to the branch and on the pull request itself. This change limits `push` events to the master branch to avoid this.


ericsciple:
thanks i noticed this the other day and was wanting to make the same change, but didnt want to lump it in with the same changes i was working on

PR #280, opened 2020-01-03
audit security vulnerabilities as part of ci

bryanmacfarlane:
We should run security audits as part of ci
See https://docs.npmjs.com/cli/audit

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "root",
   "private": true,
   "scripts": {
+    "audit-all": "lerna run audit-moderate",
     "bootstrap": "lerna bootstrap",
     "build": "lerna run tsc",
     "check-all": "concurrently \"npm:format-check\" \"npm:lint\" \"npm:test\" \"npm:build -- -- --noEmit\"",
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -27,6 +27,7 @@
     "directory": "packages/core"
   },
   "scripts": {
+    "audit-moderate": "npm audit --audit-level=moderate",
     "test": "echo \"Error: run tests from root\" && exit 1",
     "tsc": "tsc"
   },
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -27,7 +27,7 @@
     "directory": "packages/core"
   },
   "scripts": {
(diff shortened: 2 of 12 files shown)


joshmgross:
Shouldn’t this be part of a scheduled build instead?
That way we can regularly check without having to hold up PR CI builds

bryanmacfarlane:
I was going to take up the question of scheduled in another PR.

bryanmacfarlane:
I think I need to run in each packages sub dir as well.

ericsciple:
holding up CI guarantees attention. scheduled build does not :)

package.json
```diff
@@ -2,6 +2,7 @@
   "name": "root",
   "private": true,
   "scripts": {
+    "audit-all": "lerna run audit-moderate",
```
ericsciple: should this just be `audit`?

`check-all` runs multiple commands (format-check, lint, test, build)

ericsciple: does this check the root package.lock.json or just the individiual projects?

if just invidivual projects, consider `npm audit --audit-level=moderate && lerna run audit-moderate`

bryanmacfarlane: It can't be audit since there's already an npm audit cmd.  lerna run audit-moderate will run the audit-moderate script cmd in each package and each packages audit-moderate action calls npm audit with a level of moderate.

packages/core/package.json
```diff
@@ -27,6 +27,7 @@
     "directory": "packages/core"
   },
   "scripts": {
+    "audit-moderate": "npm install && npm audit --audit-level=moderate",
```
bryanmacfarlane: We don't want other sibling packages in package-lock.json files since lerna manages but npm audit will fail saying it's not there - run npm install to fix.  

I also tried the other way of just checking in the lock file changes from npm install with the hash and the url from npm but then there's issues dev side if you change a package and want to consume those changes from a sibling package before publishing it.

This seems to work although not ideal.  I'm open to other suggestions.  Not a lerna expert ...  However, I do see there's an open issue in lerna to add a lerna audit first class command so maybe that's the ideal solution in the future.  https://github.com/lerna/lerna/issues/1663

PR #283, opened 2020-01-04
only audit on ubuntu-latest

bryanmacfarlane:
we don't need to run module audits on all OSes and ubuntu is the fastest run


bryanmacfarlane:
merging this PR without a review to address some timing and fragility.  FYI @ericsciple @joshmgross and @konradpabjan

Issue #229, opened 2019-11-24
exec: specify shell to run the command in


SaschaMann:
**Describe the enhancement**
Add an [option](https://github.com/actions/toolkit/blob/d9347d4ab99fd507c0b9104b2cf79fb44fcc827d/packages/exec/src/interfaces.ts#L5) to `exec.exec` to invoke a specific shell, e.g. `/bin/bash`, when running the command.
Alternatively, document that `exec.exec` does not invoke a shell by default.

**Code Snippet**
Currently:

```typescript
await exec.exec('/bin/bash', ['-c', `cp -a /Volumes/Julia-*/Julia-*.app/Contents/Resources/julia ${process.env.HOME}`])
```

Better:

```typescript
await exec.exec('cp', ['-a', '/Volumes/Julia-*/Julia-*.app/Contents/Resources/julia', `${process.env.HOME}/julia`], {'shell': '/bin/bash'})
```

**Additional information**
The docs say that "you can use this package to execute your tools **on the command line** in a cross platform way". To me that implies it invokes a shell, which it doesn't. This misunderstanding has caused me quite the headache when I tried to find out why `cp` wasn't working with wildcards.

Disclaimer: I have almost zero experience with Typescript, perhaps this behaviour is obvious to people who regularly use it. 🤷‍♂

1138-4EB:
Ref #232

alexkappa:
It seems the solution to this issue has been to change the wording in documentation, while the actual issue was not addressed imho.

I'll be submitting a PR which allows defining the shell as with [`child_process`](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options) options.

PR #285, opened 2020-01-06
Add support for using a shell to run commands

alexkappa:
This PR adds support for supplying a shell to use when running commands. This is useful with expanding wildcards among other things.

Fixes: #229

### DIFF
--- a/packages/exec/__tests__/exec.test.ts
+++ b/packages/exec/__tests__/exec.test.ts
@@ -170,6 +170,7 @@ describe('@actions/exec', () => {
     expect(exitCode).toBe(0)
   })
 
+
   it('Exec fails with error on bad call', async () => {
     const _testExecOptions = getExecOptions()
 
@@ -303,7 +304,7 @@ describe('@actions/exec', () => {
     expect(stderrCalled).toBeTruthy()
   })
 
-  it('Handles child process holding streams open', async function () {
+  it('Handles child process holding streams open', async function() {
     const semaphorePath = path.join(
       getTestTemp(),
       'child-process-semaphore.txt'
@@ -349,7 +350,7 @@ describe('@actions/exec', () => {
     fs.unlinkSync(semaphorePath)
   }, 10000) // this was timing out on some slower hosted macOS runs at default 5s
 
-  it('Handles child process holding streams open and non-zero exit code', async function () {
+  it('Handles child process holding streams open and non-zero exit code', async function() {
     const semaphorePath = path.join(
       getTestTemp(),
       'child-process-semaphore.txt'
@@ -403,7 +404,7 @@ describe('@actions/exec', () => {
     fs.unlinkSync(semaphorePath)
   }, 10000) // this was timing out on some slower hosted macOS runs at default 5s
 
-  it('Handles child process holding streams open and stderr', async function () {
+  it('Handles child process holding streams open and stderr', async function() {
     const semaphorePath = path.join(
       getTestTemp(),
       'child-process-semaphore.txt'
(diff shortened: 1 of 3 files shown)


packages/exec/src/interfaces.ts
```diff
@@ -12,6 +12,9 @@ export interface ExecOptions {
   /** optional.  defaults to false */
   silent?: boolean
 
+  /** optional. Runs command inside of a shell. A different shell can be specified as a string. Defaults to false */
+  shell?: boolean | string
```
ericsciple: How does this affect the quoting rules on windows? linux?

Lots of logic in tool runner to accurately deal with arg quoting for .exe and different rules for .cmd/.bat.

Today not a problem on Linux because iirc node calls execv which takes an arg array. Different on Windows because everything gets joined into a single string. Does this introduce arg quoting challenges on Linux?

ericsciple: And if so, are the quoting rules different depending on the shell?

alexkappa: Hi @ericsciple, from the documentation of this options:

> - `shell` `<boolean>` | `<string>` If true, runs command inside of a shell. Uses '/bin/sh' on Unix, and process.env.ComSpec on Windows. A different shell can be specified as a string. See Shell Requirements and Default Windows Shell. Default: false (no shell).
> - `windowsVerbatimArguments` `<boolean>` No quoting or escaping of arguments is done on Windows. Ignored on Unix. This is set to true automatically when shell is specified and is CMD. Default: false.

alexkappa: Perhaps on windows we could set `windowsVerbatimArguments` to `true` if `shell` is specified and is equal to 'CMD'. That way we match the behavior of the stdlib, and allow our quoting to work with that assumption.

On linux I don't know if there are any quoting differences, but I'd be happy to add more tests to find out.

ericsciple: Thanks. Other thoughts, spitballing...

If drastically different, I wonder whether it would make more sense to be linux/mac only feature.

If quoting issues on Linux, then existing workaround of `tool: 'sh', args: ['-c', 'ls -l *.md']` is at least transparent what is happening

alexkappa: I am not entirely sure how to proceed with this @ericsciple. I understand the workaround, but I thought since node supports it natively, it could be low hanging fruit with a more elegant solution.

Will more tests alleviate your concerns? I understand you are reluctant to accept this as there might be edge cases that break quoting. If so, I might need some help :)

packages/exec/__tests__/exec.test.ts
```diff
@@ -153,6 +153,23 @@ describe('@actions/exec', () => {
...
   })
 
+  it('Runs exec successfully with shell wildcard substitution', async () => {
+    let tool: string
+    let args: string[]
+    let opts: im.ExecOptions
+    if (IS_WINDOWS) {
+      tool = 'cmd'
+      args = ['/c', 'echo', 'hello']
+      opts = {shell: process.env.ComSpec}
```
craigdbarber: Suggest also adding a test for running bash shell within Windows. As bash is now supported by windows runners, it provides a universal target for actions wanting to execute shell commands.

imbsky:
What is the status of this PR? This change is required for `actions/cache`.

Lelee703 (approved):

packages/exec/__tests__/exec.test.ts
Lelee703: ok

Lelee703 (approved):

packages/exec/src/interfaces.ts
Lelee703: same here

Lelee703 (approved):

timharris777:
Why hasn't this been supported yet? I see that `@actions/exec` is actually using `child_process.spawn` under the hood. I agree with @alexkappa that adding  the `shell` option as an available option to pipe to the spawn command should be quite easy. I will be forking @alexkappa's fork as I have use of this.

PR #292, opened 2020-01-08
Update Payload Interface to use Definitions from @octokit/webhooks

### LINKED ISSUE #147: @actions/GitHub - TypeScript Definition for WebHookPayload is limited

### LINKED ISSUE #289: Issue Number does not populate correctly in `@actions/github` Context.issue
**Describe the bug**
Issue number is empty for PullRequestReview and PullRequestReviewComment events

**To Reproduce**
1. Create an action using the following ts file:
```

async function run(): Promise<void> {
  try {
    core.info(`Issue number is: ${JSON.stringify(github.context.issue.number)}`)
  } catch (error) {
    core.setFailed(error.message)
  }
}

run()
```
2.  Setup the workflow to run on `on: [pull_request_review, pull_request_review_comment]`
3. Create a PR with a linked issue

**Expected behavior**
The Issue number field should be populated.

**Screenshots**
The issue number field is null.

**Additional context**
We appear to be accessing `pullRequest` instead of `pull_request` causing this bug: 
https://github.com/actions/toolkit/blob/master/packages/github/src/context.ts#L47

thboop:
Resolves #289 
Resolves #147

This is a breaking compat change and will require a major version bump.

### DIFF
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -76,7 +76,7 @@ There may be cases where multiple events have a field you need. For example, we 
 ```js
 const github = require('@actions/github');
 const payload = github.context.payload
-if ("issue" in payload) {
+if (payload && "issue" in payload) {
   core.info(`The issue number is: ${JSON.stringify(payload.issue.number)}`)
 }
 ```
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -56,10 +56,10 @@ const newIssue = await octokit.issues.create({
 });
 ```
 
-### Context Payload Casting
+### Context Payload Casting in Typescript
 The [GitHub Webhook Event](https://developer.github.com/webhooks/#events) payload is provided as a part of the context. Based on the type of event that [triggered your workflow](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows), the structure of that object may change. You can cast this field to access these members in a type safe way.
 
-```js
+```ts
 const github = require('@actions/github');
 
 const payload = github.context.payload
@@ -73,11 +73,11 @@ if ((payload as Webhooks.WebhookPayloadPush).head_commit) {
 ```
 
 There may be cases where multiple events have a field you need. For example, we can get the issue number field from the `issue` event and the `issue_comment` event
-```js
+```ts
 const github = require('@actions/github');
 const payload = github.context.payload
 if (payload && "issue" in payload) {
-  core.info(`The issue number is: ${JSON.stringify(payload.issue.number)}`)
+  core.info(`The issue number is: ${payload.issue.number}`)
 }
 ```
 
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -55,3 +55,29 @@ const newIssue = await octokit.issues.create({
   body: 'Hello Universe!'
 });
 ```
+
+### Context Payload Casting
+The [GitHub Webhook Event](https://developer.github.com/webhooks/#events) payload is provided as a part of the context. Based on the type of event that [triggered your workflow](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows), the structure of that object may change. You can cast this field to access these members in a type safe way.
+
+```js
+const github = require('@actions/github');
+
+const payload = github.context.payload
+if ((payload as Webhooks.WebhookPayloadPush).head_commit) {
+  core.info(
+    `The head commit is: ${
+      (payload as Webhooks.WebhookPayloadPush).head_commit
+    }`
+  )
+}
+```
+
+There may be cases where multiple events have a field you need. For example, we can get the issue number field from the `issue` event and the `issue_comment` event
+```js
+const github = require('@actions/github');
+const payload = github.context.payload
+if ("issue" in payload) {
+  core.info(`The issue number is: ${JSON.stringify(payload.issue.number)}`)
+}
+```
+
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -60,24 +60,26 @@ const newIssue = await octokit.issues.create({
 The [GitHub Webhook Event](https://developer.github.com/webhooks/#events) payload is provided as a part of the context. Based on the type of event that [triggered your workflow](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows), the structure of that object may change. You can cast this field to access these members in a type safe way.
 
 ```ts
-const github = require('@actions/github');
+import * as core from '@actions/core'
+import * as github from '@actions/github'
+import * as Webhooks from '@octokit/webhooks'
 
 const payload = github.context.payload
-if ((payload as Webhooks.WebhookPayloadPush).head_commit) {
-  core.info(
-    `The head commit is: ${
-      (payload as Webhooks.WebhookPayloadPush).head_commit
-    }`
-  )
+if (github.context.eventName === 'push') {
+  const pushPayload = github.context.payload as Webhooks.WebhookPayloadPush
+  core.info(`The head commit is: ${pushPayload.head_commit}`)
 }
 ```
 
 There may be cases where multiple events have a field you need. For example, we can get the issue number field from the `issue` event and the `issue_comment` event
 ```ts
-const github = require('@actions/github');
+import * as core from '@actions/core'
+import * as github from '@actions/github'
+import * as Webhooks from '@octokit/webhooks'
+
 const payload = github.context.payload
-if (payload && "issue" in payload) {
-  core.info(`The issue number is: ${payload.issue.number}`)
+if (payload && 'pull_request' in payload) {
+  core.info(`The issue is: ${JSON.stringify(payload.pull_request.number)}`)
 }
 ```
 
--- a/packages/github/__tests__/lib.test.ts
+++ b/packages/github/__tests__/lib.test.ts
@@ -55,7 +55,7 @@ describe('@actions/context', () => {
   it('works with pullRequest payloads', () => {
     delete process.env.GITHUB_REPOSITORY
     context.payload = {
-      pullRequest: {number: 2},
+      pull_request: {number: 2},
       repository: {owner: {login: 'user'}, name: 'test'}
     }
     expect(context.issue).toEqual({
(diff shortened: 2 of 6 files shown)


packages/github/__tests__/lib.test.ts
```diff
@@ -29,18 +41,13 @@ describe('@actions/context', () => {
 
   it('returns attributes from the repository payload', () => {
     delete process.env.GITHUB_REPOSITORY
-
-    context.payload.repository = {
```
thboop: Lots of small no-op cleanup in here

packages/github/__tests__/lib.test.ts
```diff
@@ -55,9 +62,10 @@ describe('@actions/context', () => {
   it('works with pullRequest payloads', () => {
     delete process.env.GITHUB_REPOSITORY
     context.payload = {
-      pullRequest: {number: 2},
-      repository: {owner: {login: 'user'}, name: 'test'}
-    }
+      // eslint-disable-next-line @typescript-eslint/camelcase
```
thboop: we need to match the github event payload, which is not camelcase

packages/github/src/context.ts
```diff
@@ -7,7 +7,7 @@ export class Context {
   /**
    * Webhook payload object that triggered the workflow
    */
-  payload: WebhookPayload
+  payload?: WebhookPayload
```
thboop: Before, we returned an object with every field optional. If we can't actually create the object, we should just return undefined, otherwise we allow users to cast to what they need.

packages/github/src/interfaces.ts
```diff
@@ -1,39 +1,29 @@
-/* eslint-disable @typescript-eslint/no-explicit-any */
+// TODO: https://github.com/actions/toolkit/issues/291, ESLint chokes on the | syntax
+/* eslint-disable no-undef */
+import Webhooks from '@octokit/webhooks'
+type WebhookPayload =
+  | Webhooks.WebhookPayloadPush
```
thboop: The event corresponds to the name pretty cleanly here so I chose not to add comments 
WebhookPayloadPush = "push" event

packages/github/src/interfaces.ts
```diff
@@ -1,39 +1,29 @@
...
-    [key: string]: any
-    type: string
-  }
-  action?: string
-  installation?: {
-    id: number
-    [key: string]: any
-  }
-}
+export {Webhooks, WebhookPayload}
```
thboop: re-export the Webhooks interfaces so consumers can reuse those via

import {Webhooks} from '@actions/github/lib/interfaces'

packages/github/README.md
```diff
@@ -55,3 +55,29 @@ const newIssue = await octokit.issues.create({
   body: 'Hello Universe!'
 });
 ```
+
+### Context Payload Casting in Typescript
+The [GitHub Webhook Event](https://developer.github.com/webhooks/#events) payload is provided as a part of the context. Based on the type of event that [triggered your workflow](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows), the structure of that object may change. You can cast this field to access these members in a type safe way.
+
+```ts
+const github = require('@actions/github');
```
ericsciple: is a require missing for `Webhooks`?

packages/github/README.md
```diff
@@ -55,3 +55,29 @@ const newIssue = await octokit.issues.create({
...
+
+const payload = github.context.payload
```
ericsciple: nit: `const payload = github.context.payload as Webhooks.WebhookPayloadPush` would eliminate the redundant casts further below

packages/github/__tests__/lib.test.ts
```diff
@@ -29,18 +41,13 @@ describe('@actions/context', () => {
...
-      owner: {login: 'user'}
-    }
     expect(context.repo).toEqual({owner: 'user', repo: 'test'})
   })
 
   it("return error for context.repo when repository doesn't exist", () => {
     delete process.env.GITHUB_REPOSITORY
-
-    context.payload.repository = undefined
+    delete process.env.GITHUB_EVENT_PATH
```
ericsciple: should an after all restore the original values? otherwise may affect test suites that execute after

packages/github/README.md
```diff
@@ -55,3 +55,29 @@ const newIssue = await octokit.issues.create({
...
+if ((payload as Webhooks.WebhookPayloadPush).head_commit) {
+  core.info(
+    `The head commit is: ${
+      (payload as Webhooks.WebhookPayloadPush).head_commit
+    }`
+  )
+}
+```
+
+There may be cases where multiple events have a field you need. For example, we can get the issue number field from the `issue` event and the `issue_comment` event
```
ericsciple: would it be safer to check the expected event types?

ericsciple: i.e. isnt there an event name property?

packages/github/package.json
```diff
@@ -38,7 +38,8 @@
   },
   "dependencies": {
     "@octokit/graphql": "^4.3.1",
-    "@octokit/rest": "^16.15.0"
+    "@octokit/rest": "^16.15.0",
+    "@octokit/webhooks": "^7.0.0"
```
ericsciple: what is the size impact to consumers of the `github` module?

packages/github/src/interfaces.ts
```diff
@@ -1,39 +1,29 @@
...
+  | Webhooks.WebhookPayloadDelete
+  | Webhooks.WebhookPayloadProjectCard
+  | Webhooks.WebhookPayloadPageBuild
+  | Webhooks.WebhookPayloadFork
+  | Webhooks.WebhookPayloadGollum
+  | Webhooks.WebhookPayloadMilestone
+  | Webhooks.WebhookPayloadProject
+  | Webhooks.WebhookPayloadLabel
 
-export interface PayloadRepository {
```
ericsciple: if we re-export PayloadRepository then can we avoid a new major version?

packages/github/src/interfaces.ts
ericsciple: do we need to re-export Webhooks? why not let users import webhooks interfaces the normal way (from `@octokit/webhooks`)

packages/github/src/interfaces.ts
ericsciple: is the leading `|` intentional? seems like it would only be between each interface, but before the first interface.

does a leading `|` have a behavioral implication? or is it just an optionally allowed style?

packages/github/__tests__/lib.test.ts
thboop: The beforeeach function sets up these values
https://github.com/actions/toolkit/pull/292/files#diff-75df1656ab4669c70d84cd99613c8a6eL10

packages/github/src/interfaces.ts
thboop: This is [valid typescript] (https://github.com/microsoft/TypeScript/issues/12071) and Prettier formats it this way. I actually coded it the way you suggested and prettier failed the check.

```
type a = b | c
```
is equivalent to 
```
type a = 
| b
| c
```

https://github.com/prettier/prettier/issues/6299

packages/github/README.md
thboop: I don't think we should recommend users do that though. Type assertions rely on the developer to perform relevant checks before casting, and this example shows that.

From their [docs](https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions)

Type assertions are a way to tell the compiler “trust me, I know what I’m doing.” A type assertion is like a type cast in other languages, but performs no special checking or restructuring of data. It has no runtime impact, and is used purely by the compiler. TypeScript assumes that you, ***the programmer, have performed any special checks that you need.***

packages/github/README.md
thboop: Going to update the examples to do this

packages/github/src/interfaces.ts
thboop: I don't believe that there's a great way to keep back-compat and set us on a good foot going forward. 

We took some of the events and combined them into one big type where we marked each field as optional in `WebhookPayload` This doesn't really set us on a great path forward because it makes it hard to tell which fields should exist on the event.

For example, a push event would not have an `issue` or a `pull_request`, but they are marked as optional when consuming. Ideally, this would be distinct from fields that are actually optional fields, but there is no way to tell now.

It conveys the wrong message to a user that those fields are optional, when in fact they depend on the type of event. Our users who want type safety should be casting based on the type of event, we shouldn't combine every event into one big interface with each field is marked as optional, it makes it harder to consume.

It also sends the wrong message because the individual fields depend on the type of event. For example the "pull_request" field on a PullRequestReview is a different object with different fields from the `pull_request` field on a "PullRequest" event.  `payload.pull_request.commits` is only a valid field for PullRequests but not PullRequestReviews. 

By setting these as union types, we enforce that users need to do a type assertion using the `as` keyword, or ignore types by using these as `any` objects

packages/github/package.json
thboop: None of the code gets routed to ncc, so there is no impact for consumers using that.

However, I think we can install this as a dev dependency, so users doing npm install --prod won't even download it. Going to test that out and make that corresponding change.

packages/github/__tests__/lib.test.ts
ericsciple: should an after-all cleanup? so the changes don't spill-over and affect other test suites?

thboop: Oh sorry, I misunderstood your previous comment. Yeah, I'll go ahead and update this

packages/github/package.json
thboop: There's a bit of a mixed reaction in the ts/js community, but it looks like we shouldn't install as dev-dependencies [we export these types](https://stackoverflow.com/a/46011417) [open @types issue](https://github.com/Microsoft/types-publisher/issues/81)

In its current state (saved as a dependency)
Users who upload their production npm packages rather then using ncc are going to have the size of their  action increased by the size of the npm package and its dependencies, which is 291 KB or 400kb on disk for my machine.

thboop:
Closing in favor of #310

Issue #261, opened 2019-12-17
tracking: publish packages


ericsciple:
Bump patch version and release all modules

Related PRs:
- https://github.com/actions/toolkit/pull/257
- https://github.com/actions/toolkit/pull/221
- https://github.com/actions/toolkit/pull/263
- https://github.com/actions/toolkit/pull/264
- https://github.com/actions/toolkit/pull/287

mies386:
#261
> GITHUB_TOKEN#github.org

PR #295, opened 2020-01-09
bump versions

ericsciple:
related #261


packages/tool-cache/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/tool-cache",
-  "version": "1.1.2",
+  "version": "1.2.0",
```
ericsciple: new functionality was added

PR #297, opened 2020-01-10
remove todo comment

ericsciple:


packages/glob/__tests__/internal-globber.test.ts
```diff
@@ -134,7 +134,6 @@ describe('globber', () => {
 
     const itemPaths = await glob(root, {followSymbolicLinks: true})
     expect(itemPaths).toEqual([root, path.join(root, 'file')])
-    // todo: ? expect(itemPaths[2]).toBe(path.join(root, 'symDir'))
```
ericsciple: @TingluoHuang fyi, i left the todo in here to make sure to discuss this behavior explicitly. removing now. its a weird edge case so i think either way is fine. not returning the recursive symlink (when following symlinks) seems good and lines up with how broken symlinks arent returned when following symlinks

PR #298, opened 2020-01-10
audit fix and update http-client

bryanmacfarlane:
gets us proxy support that *just works* as well as es2019 optimal run time code for node 12 runners


bryanmacfarlane:
```bash
[dev] ~/Projects/toolkit2$ npm audit fix
npm WARN eslint-plugin-github@2.0.0 requires a peer of flow-bin@>=0.70.0 but none is installed. You must install peer dependencies yourself.
```

so I had to add flow-bin as a dev dependency

PR #302, opened 2020-01-12
fix command escaping

### LINKED ISSUE #301: Fix command escaping to match runner
The toolkit uses the new command syntax (`::command::`), yet the old escaping rules (for `##[command]`).

The escaping rules are:

```
        private static readonly EscapeMapping[] _escapeDataMappings = new[]
        {
            new EscapeMapping(token: "\r", replacement: "%0D"),
            new EscapeMapping(token: "\n", replacement: "%0A"),
            new EscapeMapping(token: "%", replacement: "%25"),
        };

        private static readonly EscapeMapping[] _escapePropertyMappings = new[]
        {
            new EscapeMapping(token: "\r", replacement: "%0D"),
            new EscapeMapping(token: "\n", replacement: "%0A"),
            new EscapeMapping(token: ":", replacement: "%3A"),
            new EscapeMapping(token: ",", replacement: "%2C"),
            new EscapeMapping(token: "%", replacement: "%25"),
        };
```

ericsciple:
fixes #301

### DIFF
--- a/packages/core/src/command.ts
+++ b/packages/core/src/command.ts
@@ -10,11 +10,11 @@ interface CommandProperties {
  * Commands
  *
  * Command Format:
- *   ##[name key=value;key=value]message
+ *   ::name key=value,key=value::message
  *
  * Examples:
- *   ##[warning]This is the user warning message
- *   ##[set-secret name=mypassword]definitelyNotAPassword!
+ *   ::warning::This is the message
+ *   ::set-env name=MY_VAR::some value
  */
 export function issueCommand(
   command: string,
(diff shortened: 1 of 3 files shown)


ericsciple:
also related to https://github.com/actions/runner/issues/267

packages/core/src/command.ts
```diff
@@ -62,33 +62,29 @@ class Command {
               cmdStr += ','
             }
 
-            // safely append the val - avoid blowing up when attempting to
-            // call .replace() if message is not a string for some reason
-            cmdStr += `${key}=${escape(`${val || ''}`)}`
+            cmdStr += `${key}=${escapeProperty(val)}`
```
tjamet: I was reading it and thinking that eventually, the function being exported, it is diable for users to run `issueCommand('some-command', {'key,1': 'value'}, 'message')` that would break the `::name key1=value1,key2=value2::message` formatting. would it be something worth taking into account?

ericsciple: It is a limitation of the command format - [defined](https://github.com/actions/runner/blob/a9c58d7398d463b0dc5d7d1ca1a8f99d2005a808/src/Runner.Common/ActionCommand.cs) by the runner - that commas are not allowed within a key name.

Today this limitation does not present a problem. All commands (and keys) are first party - not extensible.

We can relax this constraint in the future needed. But not required for any current scenarios.

Issue #118, opened 2019-09-06
Add ability to mock out process.stdout calls


jeffrafter:
Currently `issueCommand` always outputs to `stdout` here: https://github.com/actions/toolkit/blob/master/packages/core/src/command.ts#L25. This is great but when writing tests for actions it quickly fills up the logs:

```
> @ test /Users/njero/Code/Examples/example-github-action-typescript
> jest

 PASS  .github/actions/debug-action/__tests__/debug.test.ts
  debug action debug messages
    ✓ outputs a debug message (5ms)

##[debug]👋 Hello! You are an amazing person! 🙌
##[debug]👋 Hello! You are an amazing person! 🙌
Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        1.423s
Ran all test suites.
```

If the output was wrapped in a call on the `Command` object itself then it could be mocked in the jest setup. Additionally this might pave the way for alternate logging outputs down the road.

jeffrafter:
It is also possible to mock this globally in the `jest.config.js`:

```js
const processStdoutWrite = process.stdout.write.bind(process.stdout)
process.stdout.write = (str, encoding, cb) => {
  // Note: this will soon change to ::
  if (!str.match(/^\#\#/)) {
    return processStdoutWrite(str, encoding, cb)
  }
}
```

bryanmacfarlane:
Thanks for logging this issue!  We will implement this (we also take PRs ;) )

xt0rted:
These log messages from tests are now causing annotations to be created when they shouldn't be. You can see an example of this in my most recent action run here https://github.com/xt0rted/slash-command-action/pull/34/checks?check_run_id=259915650#step:8:9 and here https://github.com/xt0rted/slash-command-action/pull/34/checks?check_run_id=259915650#step:8:63.

bryanmacfarlane:
@xt0rted - not related to mocking but I think proper annotations support will fix your issue.  I can prioritize the first half of that work (ensuring logging warning / error output doesn't create an annotation).  That's a bug and I can fix.  Sound ok?  https://github.com/actions/toolkit/issues/186

xt0rted:
@bryanmacfarlane I meant to post that on #163, both issues a pretty similar in their end goal.

ericsciple:
would mocking the command module make more sense?

ericsciple:
[...]

PR #304, opened 2020-01-13
@actions/artifact package

konradpabjan:
Initial commit for the `@actions/artifact` package. So far only contains the code for uploading an artifact, download will come in a separate PR.

Adds support for uploading artifacts with new APIs that are independent of the runner (non-plugin). The original code for uploading artifacts can be found here (see https://github.com/actions/runner/tree/master/src/Runner.Plugins/Artifact). This PR largely ports over the C# code for uploading artifacts to node and switches over to some recently added APIs. This new package will be used for the `v2` versions of `actions/upload-artifact` and `actions/download-artifact`. 

Been using the structure of `@actions/glob` in this repo as an example. For this PR I deliberately did not do `tsc` and `npm-install` so there is no `lib` folder or `node_modules` folder. Trying to keep it clean and simple for this PR so that it is easy to review.

Been testing with the following two repos for upload artifact: https://github.com/bbq-beets/artifact-package and https://github.com/bbq-beets/upload-artifact-v2

Future work not part of this PR
- ~Tests (all the code in `search.ts` and `util.ts` is a good candidate, will be added to a `_TEST_` directory just like in all the other packages in this repo)~ (Tests have been added)
- All the code related to `downloadArtifact` and `downloadAllArtifacts` (TODOs can be seen in `src/internal-artifact.ts`)
- README with examples (will be added after download artifact is working and the package is almost ready to release)
- `lib` and `node_modules` folder, (just like in `@actions/glob), will be added as we get close to release and everything is tested

~Search uses our new glob package: `@actions/glob`~ (searching will be externally done)
Http calls are made using our new: `@actions/http-client` package in order to support proxies

Much of the work that was done in `@actions/cache` was used as a reference in terms of how concurrent chunked uploads are done, see: https://github.com/actions/cache/pull/128

Formatting was done by doing `prettier --write packages src/*.ts` so the style should be consistent with the other code in this repo.


packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,111 @@
...
+ * @param name the name of the artifact, required
+ * @param path the directory, file, or glob pattern to denote what will be uploaded, required
+ * @returns single UploadInfo object
+ */
+export async function uploadArtifact(
+  name: string,
+  path: string
+): Promise<UploadInfo> {
+  // Check that the required inputs are valid
+  if (!name) {
```
yacaovsnc: Should this check be folded into the `checkArtifactName` function?

packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,111 @@
...
+      `No files were found for the provided path: ${path}. No artifacts will be uploaded.`
+    )
+  } else {
+    /**
+     * Step 1 of 3
+     * Create an entry for the artifact in the file container
+     */
+    const response: CreateArtifactResponse = await createArtifactInFileContainer(
+      name
+    )
```
yacaovsnc: Single line with `;`?

packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (!response.fileContainerResourceUrl) {
+      console.log(response)
```
yacaovsnc: Log this as verbose or debug?

packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,111 @@
...
+      throw new Error('Unable to get fileContainerResourceUrl')
```
yacaovsnc: User doesn't understand this variable.  Try to word it so user can understand what is wrong.

packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,111 @@
...
+     * Upload each of the files that were found concurrently
+     */
+    let artfactSize: Promise<number> = Promise.resolve(
+      uploadArtifactToFileContainer(
+        response.fileContainerResourceUrl!,
+        filesToUpload
+      )
+    )
+    artfactSize.then(async size => {
+      console.log(`Size of what we just uploaded is ${size}`)
```
yacaovsnc: We should add the `name` here so user can quickly search for an artifact.

packages/artifact/src/upload-artifact-http-client.ts
```diff
@@ -0,0 +1,300 @@
...
+  let rawResponse: HttpClientResponse = await client.patch(
+    resourceUrl.toString(),
+    data,
+    requestOptions
+  )
+  let body: string = await rawResponse.readBody()
+
+  if (rawResponse.message.statusCode == 200) {
+    let successResponse: PatchArtifactSizeSuccessResponse = JSON.parse(body)
+    console.log('Artifact size was succesfully updated!')
```
yacaovsnc: Although we are patching for size, but to users this is just finishing up uploading artifacts as a whole.  I would say change this line to indicate artifact uploaded successfully.  You could even move the log around artifact size here, to something like `Artifact $name uploaded successfully, total size $size`.   

A general comment is we should check our logging in those files and make sure to read them again from users' perspective.

packages/artifact/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/actions/toolkit.git",
+    "directory": "packages/artifact"
+  },
+  "scripts": {
+    "audit-moderate": "npm install && npm audit --audit-level=moderate",
+    "test": "echo \"Error: no test specified\" && exit 1",
```
joshmgross: I think tests should be part of this PR, at least covering the Upload APIs

packages/artifact/src/artifact.ts
joshmgross: Should be debug, which is effectively verbose logging in Actions

packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,108 @@
...
+     * Create an entry for the artifact in the file container
+     */
+    const response: CreateArtifactResponse = await createArtifactInFileContainer(
+      name
+    )
+    if (!response.fileContainerResourceUrl) {
+      console.log(response)
+      throw new Error('Unable to get fileContainerResourceUrl')
+    }
+    core.debug(`We will be uploading to: ${response.fileContainerResourceUrl}`)
```
joshmgross: nit: Can be more concise with just `Upload Resource URL: ${response.fileContainerResourceUrl}`

"We" feels odd for debug logging

packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,108 @@
...
+     * Upload each of the files that were found concurrently
+     */
+    const uploadingArtifact: Promise<number> = Promise.resolve(
+      uploadArtifactToFileContainer(
+        response.fileContainerResourceUrl,
+        filesToUpload
+      )
+    )
+    uploadingArtifact.then(async size => {
+      console.log(`Size of what we just uploaded is ${size}`)
```
joshmgross: Don't you need to `await` this promise?
Also no need to mix syntax with `async`/`await`

```ts
const artifactSize = await uploadArtifactToFileContainer(response.fileContainerResourceUrl, filesToUpload)
console.log(`Uploaded Artifact ${name} with size: ${size}`)
```

packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,108 @@
...
+       * Update the size of the artifact to indicate we are done uploading
+       */
+      await patchArtifactSize(size, name)
+      reportedSize = size
+    })
+  }
+
+  return {
+    artifactName: name,
+    artifactItems: filesToUpload.map(item => item.absoluteFilePath),
```
joshmgross: `filesToUpload` is the input files we tried to upload, not necessarily the files that were uploaded.

In the case of errors during upload, will this error out or will this return a partial list of files successfully uploaded?

packages/artifact/src/contracts.ts
```diff
@@ -0,0 +1,28 @@
+export interface CreateArtifactResponse {
```
joshmgross: Why is everything optional in this response contract? I would expect some guaranteed fields

packages/artifact/src/upload-artifact-http-client.ts
```diff
@@ -0,0 +1,299 @@
...
+
+  const body: string = await rawResponse.readBody()
+  const response: CreateArtifactResponse = JSON.parse(body)
+  console.log(response)
+
+  if (rawResponse.message.statusCode === 201 && response) {
+    return response
+  } else {
+    throw new Error(
+      'Non 201 status code when creating file container for new artifact'
```
joshmgross: 1. I don't think we should exclusively check `201`, since any 2xx response is a successful response status
2. The error should contain the failed response status code

packages/artifact/src/upload-artifact-http-client.ts
```diff
@@ -0,0 +1,299 @@
...
+        uploadedFiles += 1
+        fileSizes.push(await uploadFileAsync(currentFileParameters))
+      }
+    })
+  )
+
+  let sum = 0
+  for (const fileSize of fileSizes) {
+    sum += fileSize
+  }
```
joshmgross: minor: Can use `Array.Reduce` to calculate this sum:
```ts
const sum = fileSizes.reduce((acc, val) => acc + val);
```

packages/artifact/src/upload-artifact-http-client.ts
```diff
@@ -0,0 +1,299 @@
...
+          parameters.resourceUrl,
+          chunk,
+          start,
+          end,
+          fileSize
+        )
+      }
+    })
+  )
+  return fs.statSync(parameters.file).size
```
joshmgross: Can you re-use the file size from above here? No need to make the extra file system call:
```suggestion
  return fileSize
```

packages/artifact/src/upload-artifact-http-client.ts
```diff
@@ -0,0 +1,299 @@
...
+    debug(
+      `Chunk for ${start}:${end} was succesfully uploaded to ${resourceUrl}`
+    )
+    return
+  }
+  if (isRetryableStatusCode(response.message.statusCode)) {
+    console.log(
+      `Received ${response.message.statusCode}, will retry chunk at offset ${start} after 10 seconds.`
+    )
+    await new Promise(resolve => setTimeout(resolve, 10000))
```
joshmgross: Retrying 10 seconds after each failed chunk upload seems like a long time, I'm not sure what an appropriate value is.

It looks like the runner retries all failed files after a minute (as opposed to failed chunks): https://github.com/actions/runner/blob/master/src/Runner.Plugins/Artifact/FileContainerServer.cs#L250

packages/artifact/src/upload-artifact-http-client.ts
```diff
@@ -0,0 +1,299 @@
...
+    console.log(`Retrying chunk at offset ${start}`)
+
+    const retryResponse = await uploadChunkRequest()
+    if (!retryResponse.message.statusCode) {
+      console.log(retryResponse)
+      throw new Error('No Status Code returne with response')
+    }
+    if (isSuccessStatusCode(retryResponse.message.statusCode)) {
+      return
+    }
```
joshmgross: What if this is not a successful status code? Shouldn't we throw an error or at least debug the response?

packages/artifact/src/utils.ts
```diff
@@ -0,0 +1,94 @@
...
+  if (!workFlowrunId) {
+    throw new Error('Unable to get workFlowRunId')
+  }
+  return workFlowrunId
+}
+
+/**
+ * Invalid characters that cannot be in the artifact name or an uploaded file. Will be rejected
+ * from the server if attempted to be sent over. These characters are not allowed due to limitations with certain
+ * file systems such as NTFS. To maitain platform-agnostic behavior, all characters that are not supported by an
```
joshmgross:
```suggestion
 * file systems such as NTFS. To maintain platform-agnostic behavior, all characters that are not supported by an
```

packages/artifact/src/artifact.ts
konradpabjan: This was automatically changed by `prettier`. If I change it to a single line it complains: 
image

packages/artifact/src/upload-artifact-http-client.ts
konradpabjan: image


Fancy 😀 Didn't know about this `reduce` function

packages/artifact/src/artifact.ts
konradpabjan: Was able to get it shortened in a single line without `prettier` complaining

packages/artifact/src/contracts.ts
konradpabjan: VSCode must have automatically changed it after I clicked `Apply suggested fixes` or something 🤔 Definitely should not have been optional

konradpabjan:
Working on some major changes, don't review yet

yacaovsnc:
> Working on some major changes (almost ready but not quite), don't review yet

Errr, didn't see this....

packages/artifact/src/artifact.ts
konradpabjan: Changed this to `debug` and the message should be clearer now

packages/artifact/src/artifact.ts
konradpabjan: Changed wording

packages/artifact/src/upload-artifact-http-client.ts
```diff
@@ -0,0 +1,307 @@
...
+): Promise<CreateArtifactResponse> {
+  const parameters: CreateArtifactParameters = {
+    Type: 'actions_storage',
+    Name: artifactName
+  }
+  const data: string = JSON.stringify(parameters, null, 2)
+  const artifactUrl = getArtifactUrl(getRuntimeUrl(), getWorkFlowRunId())
+  const client = createHttpClient(getRuntimeToken())
+  const requestOptions = getRequestOptions(
+    'application/json',
```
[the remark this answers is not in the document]
konradpabjan: Initially one was for the `Accept` header, the other was for the `Content-Type` header. I've removed the `Accept` header parameter and just set it to `application/json` since all of our calls use it. Cleans up the code a little bit

packages/artifact/src/upload-artifact-http-client.ts
```diff
@@ -0,0 +1,307 @@
...
+  isSuccessStatusCode,
+  parseEnvNumber
+} from './utils'
+import {getRuntimeToken, getRuntimeUrl, getWorkFlowRunId} from './env-variables'
+
+const defaultChunkUploadConcurrency = 3
+const defaultFileUploadConcurrency = 2
+
+/**
+ * Step 1 of 3 when uploading an artifact. Creates a file container for the new artifact in the remote blob storage/file service
```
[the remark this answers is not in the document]
konradpabjan: Good tip. Removed any references to `steps`

packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,111 @@
...
+     * Upload each of the files that were found concurrently
+     */
+    const uploadResult = await uploadArtifactToFileContainer(
+      response.fileContainerResourceUrl,
+      filesToUpload,
+      options
+    )
+    // eslint-disable-next-line no-console
+    console.log(
+      `Finished uploading artifact ${name}. Reported size is ${uploadResult.size} bytes. There were ${uploadResult.failedItems.length} items that failed to upload`
```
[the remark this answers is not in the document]
konradpabjan: Good point. Moved it to the very after step 3

packages/artifact/src/upload-artifact-http-client.ts
konradpabjan: Changed all the comments to make them a bit more friendly from a users perspective

packages/artifact/package.json
konradpabjan: After way too many problems & issues I've added unit tests for everything. The structure matches the rest of this repo so tests are meant to be run from the root directory

packages/artifact/src/artifact.ts
konradpabjan: Got rid of `we` and changed up the logging

packages/artifact/src/artifact.ts
konradpabjan: Did a massive overhaul of based of the ADR discussions we had. Now there is the `failedItems` array that gets returned

packages/artifact/src/upload-artifact-http-client.ts
konradpabjan: Ended up doing a massive overhaul around this part of the code so I no longer need to sum up arrays. Nice tip though!

packages/artifact/src/upload-artifact-http-client.ts
konradpabjan: Our API is setup to return a 201. After a lot of testing, I've only seen a a 201 returned if there is ever anything between 200-299. 

That being said, the behavior of our APIs might change and we could return something like a 200 in the future so I've updated it to work with any success code. 

Also, in the event of a non-success code it just logs the entire raw response so there should be plenty of debug info

packages/artifact/src/upload-artifact-http-client.ts
```diff
@@ -0,0 +1,307 @@
...
+  )
+
+  const parameters: UploadFileParameters[] = []
+
+  // Prepare the necessary parameters to upload all the files
+  for (const file of filesToUpload) {
+    const resourceUrl = new URL(uploadUrl)
+    resourceUrl.searchParams.append(
+      'scope',
+      '00000000-0000-0000-0000-000000000000'
```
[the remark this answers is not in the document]
konradpabjan: Updated it to use the controller without the GUID (and it works! 😀). The C# plugin code uses the API that requires the GUID so during cleanup later on we can get rid of it

packages/artifact/src/upload-artifact-http-client.ts
```diff
@@ -0,0 +1,307 @@
...
+    })
+  )
+
+  // Sum up all the files that were uploaded
+  const sum = fileSizes.reduce((acc, val) => acc + val)
+  // eslint-disable-next-line no-console
+  console.log(`Total size of all the files uploaded ${sum}`)
+  return {
+    size: sum,
+    failedItems: []
```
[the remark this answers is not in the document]
konradpabjan: `continueOnError` is done after some exorbitant amount of testing. Should be good now!

packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,111 @@
...
+  checkArtifactName(name)
+
+  if (!path) {
+    throw new Error('Upload path must be provided')
+  }
+
+  // Search for the items that will be uploaded
+  const filesToUpload: SearchResult[] = await findFilesToUpload(name, path)
+
+  if (filesToUpload === undefined) {
```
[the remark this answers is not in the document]
konradpabjan: `filesToUpload` should never be `undefined` now. Did some small updates. Initially I had that check because Typescript would complain that it could be undefined

packages/artifact/src/artifact.ts
konradpabjan: Ended up completely changing the syntax here, no more mixing with `Promise.resolve`

packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,111 @@
...
+    throw new Error(
+      `Unable to succesfully search for files to upload with the provided path: ${path}`
+    )
+  } else if (filesToUpload.length === 0) {
+    core.warning(
+      `No files were found for the provided path: ${path}. No artifacts will be uploaded.`
+    )
+    return {
```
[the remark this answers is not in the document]
konradpabjan: Restructured so there is a single return at the very end

packages/artifact/src/upload-artifact-http-client.ts
konradpabjan: Added a bunch of `console.log` statements if there is a failure. The user might not have debugging on in the event of a failure so I always want some info to show up

konradpabjan:
> > Working on some major changes (almost ready but not quite), don't review yet
> 
> Errr, didn't see this....

Still a lot of great feedback! I should have changed the PR title so it's more obvious that it's not ready for review, I'll do that moving forward. Thanks for taking a look! 😀

packages/artifact/src/search.ts
```diff
@@ -0,0 +1,99 @@
...
+ * @param {string} artifactName The name of the artifact, used as the root directory when uploading artifacts to glob storage
+ * @return A list of files that should be uploaded along with the paths they will be uploaded with
+ */
+export async function findFilesToUpload(
+  artifactName: string,
+  searchPath: string
+): Promise<SearchResult[]> {
+  const searchResults: SearchResult[] = []
+  const itemsToUpload: string[] = []
+  const options: glob.GlobOptions = {
```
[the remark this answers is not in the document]
konradpabjan: Per our discussions, I've moved the `glob` search outside of the package. Now just a list of absolute file paths is required as input along with a root directory

packages/artifact/src/upload-artifact-http-client.ts
konradpabjan: Leaving at 10 seconds for now based off of our discussions. If there are any problems in the future with 10 seconds being too little/too much then we'll re-evaluate

packages/artifact/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "audit-moderate": "npm install && npm audit --audit-level=moderate",
+    "test": "echo \"Error: run tests from root\" && exit 1",
+    "tsc": "tsc"
+  },
+  "bugs": {
+    "url": "https://github.com/actions/toolkit/issues"
+  },
+  "dependencies": {
+    "@actions/core": "^1.2.1",
+    "@actions/http-client": "^1.0.1"
```
yacaovsnc: We should use the version that supports proxy.  Why not use the latest `http-client@1.0.6`?

packages/artifact/src/config-variables.ts
```diff
@@ -0,0 +1,35 @@
+export function getUploadFileConcurrency(): number {
+  return 2
+}
+
+export function getUploadChunkConcurrency(): number {
```
yacaovsnc: Why do we need both file concurrency and chunk concurrency?  So potentially we could open 6 connections for one job?  Seems a little bit aggressive.  Is this what we had in v1?

packages/artifact/src/upload-http-client.ts
```diff
@@ -0,0 +1,337 @@
...
+  const FILE_CONCURRENCY = getUploadFileConcurrency()
+  const CHUNK_CONCURRENCY = getUploadChunkConcurrency()
+  const MAX_CHUNK_SIZE = getUploadChunkSize()
+  debug(
+    `File Concurrency: ${FILE_CONCURRENCY}, Chunk Concurrency: ${CHUNK_CONCURRENCY} and Chunk Size: ${MAX_CHUNK_SIZE}`
+  )
+
+  const parameters: UploadFileParameters[] = []
+  let continueOnError = true
+  if (options) {
```
yacaovsnc: This is not future proof - if we add another option in the future, this check will fail.  Would 
` const continueOnError = !!(options && options.continueOnError)` 
work?

packages/artifact/src/upload-http-client.ts
```diff
@@ -0,0 +1,337 @@
...
+        )
+
+        if (!result) {
+          /**
+           * Chunk failed to upload, report as failed but continue if desired. It is possible that part of a chunk was
+           * successfully uploaded so the server may report a different size for what was uploaded
+           **/
+          isUploadSuccessful = false
+          failedChunkSizes += chunkSize
+          if (!parameters.continueOnError) {
```
yacaovsnc: We should not allow this.  If a chunk failed we should either retry or fail the file - `continueOnError` means if a file failed to upload, we still allow uploading other files; but the atomic unit should be file, a file will either upload successfully or not - we should not allow partially succeeded single file.

packages/artifact/src/upload-specification.ts
```diff
@@ -0,0 +1,87 @@
...
+         ['/home/user/files/plz-upload/file1.txt', 'my-artifact/dir/file3.txt']
+       ]
+  */
+  for (let file of artifactFiles) {
+    if (!fs.existsSync(file)) {
+      throw new Error(`File ${file} does not exist`)
+    }
+
+    if (!fs.lstatSync(file).isDirectory()) {
+      file = normalize(file)
```
yacaovsnc: Without `resolve` we can't guarantee this is absolute path, can we?

packages/artifact/src/upload-http-client.ts
konradpabjan: Good point. I'll go ahead and change this

packages/artifact/src/config-variables.ts
konradpabjan: Currently in the runner it's set to 2 concurrent file uploads and then each file is uploaded in 4MB chunks. As far as I can tell, chunks are not uploaded concurrently so there is a bump from 2 -> 6 connections: https://github.com/actions/runner/blob/6c70d53eead402ba5d53676d6ed649a04e219c9b/src/Runner.Plugins/Artifact/FileContainerServer.cs#L211

https://github.com/actions/runner/blob/6c70d53eead402ba5d53676d6ed649a04e219c9b/src/Runner.Plugins/Artifact/FileContainerServer.cs#L472

Perhaps the chunk upload concurrency can be bumped down to 2 instead of 3?

packages/artifact/src/upload-http-client.ts
konradpabjan: In `upload-options.ts`, `continueOnError` is a required property though, so as long as there are some `uploadOptions` provided this should work even if some extra properties are added to `uploadOptions`

If there ever are more options added, they should be required and the general flow will be something like this
```
// define default options
if (options){
// update default options to whatever is specified in options
}
```

packages/artifact/src/config-variables.ts
yacaovsnc: If we bump this up, we will stress the server more - if there are no requests yet, we should keep the load the same.  This will hit our pipelines AT so we need to be careful.  I would say change the chunk concurrency at 1 for now.

packages/artifact/src/upload-http-client.ts
yacaovsnc: In general, `options` should **not** be required.  Imagine if there are many options available, user should only need to set the ones they care about.

packages/artifact/src/config-variables.ts
konradpabjan: I'll change it to 1. The concurrency is already setup so in the future it will be easy to bump up if we ever want to play around with increasing it

packages/artifact/package.json
konradpabjan: Updated, to `1.0.6`. When I initially installed it, `1.0.1` was the most recent (even this old version does support proxy though). Before I was using the `typed-rest-client` npm package

packages/artifact/src/upload-http-client.ts
konradpabjan: Changed `continueOnError` so it is optional in `upload-options.ts` 

Changed the rest to this:
```
  // by default, file uploads will continue if there is an error unless specified differently in the options
  let continueOnError = true
  if (options) {
    if (options.continueOnError === false) {
      continueOnError = false
    }
  }
```
The double bang operators didn't seem very readable so I tried to avoid it. We do also need to support a default value if no `continueOnError` option is specified. Doing all that in a single line is possible but it gets pretty ugly so I think what I have is easy to follow

packages/artifact/src/upload-specification.ts
konradpabjan: There is a method that checks if a path is absolute, I don't use it anywhere though: https://nodejs.org/api/path.html#path_path_isabsolute_path

We specify that both `files` and `rootDirectory` are supposed to be absolute, but I think think my code will work if everything is relative as well. The only time there could be an issue is if one is absolute and the other is relative so `resolve` should be used to make sure the correct paths are being compared

packages/artifact/src/upload-http-client.ts
```diff
@@ -0,0 +1,327 @@
...
+
+        if (!result) {
+          /**
+           * Chunk failed to upload, report as failed and do not continue uploading any more chunks for the file. It is possible that part of a chunk was
+           * successfully uploaded so the server may report a different size for what was uploaded
+           **/
+          isUploadSuccessful = false
+          failedChunkSizes += chunkSize
+          // eslint-disable-next-line no-console
+          console.log(`Aborting upload for ${parameters.file} due to failure`)
```
yacaovsnc: Probably should be a warning.

packages/artifact/src/upload-specification.ts
konradpabjan: I've updated it so that `normalize` and `resolve` are used for all paths. FYI @joshmgross this strays a little from the ADR where we were requiring absolute file paths, we can use relative file paths now as well

yacaovsnc: If you normalize `file`, you should normalize `rootDirectory` as well.  We need to be consistent, otherwise you will run into issues if you try to upload files outside of your `cwd`.

yacaovsnc: Sorry I missed the line above.  This looks good.

konradpabjan: Yup, both `file` and `rootDirectory` get normalized

konradpabjan:
@ericsciple could you also take a 👀 at this PR. Getting ready to merge this in. A new package to the toolkit is quite a hefty addition and you're probably the best person to look at this. I have another PR pretty much ready to go after this with documentation and all the extra code related to downloading an artifact.

packages/artifact/__tests__/upload-specification.test.ts
```diff
@@ -0,0 +1,340 @@
...
+      } else if (specification.absoluteFilePath === goodItem4Path) {
+        expect(specification.uploadFilePath).toEqual(
+          path.join(artifactName, 'folder-d', 'good-item4.txt')
+        )
+      } else if (specification.absoluteFilePath === badItem3Path) {
+        expect(specification.uploadFilePath).toEqual(
+          path.join(artifactName, 'folder-f', 'bad-item3.txt')
+        )
+      } else {
+        throw new Error('this should never be reached')
```
joshmgross: If we're going to fail the test by throwing an error, we should be more explicit about what went wrong.

In this specific case, it might be better to simplify this into an array contains
```js
expect(badUploadFilePaths).toContain(specification.uploadFilePath)
```
https://jestjs.io/docs/en/expect#tocontainitem

packages/artifact/src/artifact.ts
```diff
@@ -0,0 +1,102 @@
...
+  )
+  const uploadResponse: UploadResponse = {
+    artifactName: name,
+    artifactItems: [],
+    size: 0,
+    failedItems: []
+  }
+
+  if (uploadSpecification.length === 0) {
+    core.warning(`No files found that can be uploaded`)
```
joshmgross: Should this be a warning? It seems like an error case if there are no files to upload.

packages/artifact/src/upload-http-client.ts
joshmgross: I think this would be more readable as Yang suggested:
```js
const continueOnError = !!(options && options.continueOnError)
```

packages/artifact/src/upload-http-client.ts
```diff
@@ -0,0 +1,326 @@
...
+            // Existing uploads will be able to finish however all pending uploads will fail fast
+            abortPendingFileUploads = true
+          }
+        }
+      }
+    })
+  )
+
+  // eslint-disable-next-line no-console
+  console.log(`Total size of all the files uploaded is ${fileSizes} bytes`)
```
joshmgross: Minor: you can use `core.info` here and elsewhere if you want to `console.log`. It's easier to mock `core` than `console` and you don't need the eslint disables

packages/artifact/src/artifact.ts
konradpabjan: The current behavior is that and empty upload will throw a warning but not fail.

It is possible that someone includes only empty directories in the files that they want uploaded. If that is the case, the `uploadspecification` will be empty because directories get ignored by the FCS during upload.

packages/artifact/__tests__/upload-specification.test.ts
konradpabjan: `toContainer` is fairly similar to what I have with `includes`, I can swap that out in a few places that I use it because it seems nicer

I use an array to check for the `absolutePaths` however using just another array for `uploadFilePath` does not 100% check for what we need. It is possible that all the correct strings exist for `uploadFilePath` in an array, however the right combination must match with the `absolutePath` which is what I'm doing with the `if else-if` statements

packages/artifact/src/upload-http-client.ts
konradpabjan: So that one liner would be fine _if_ we didn't care about the default `true` behavior if no options are provided. So it just won't work

Changing to just that one line changes the expected behavior if no options are provided (2 tests start to fail).

packages/artifact/src/upload-http-client.ts
konradpabjan: It seems like `core.info` isn't documented anywhere, even though it exists 🤔 https://github.com/actions/toolkit/tree/master/packages/core

I can use it a few places, though `console.log` is better because it isn't limited to just strings so certain things like formatted http responses look better

packages/artifact/src/upload-http-client.ts
```diff
@@ -0,0 +1,326 @@
...
+    // eslint-disable-next-line no-console
+    console.log(
+      `Received http ${response.message.statusCode} during chunk upload, will retry at offset ${start} after 10 seconds.`
+    )
+    await new Promise(resolve => setTimeout(resolve, 10000))
+    const retryResponse = await uploadChunkRequest()
+    if (isSuccessStatusCode(retryResponse.message.statusCode)) {
+      return true
+    } else {
+      // eslint-disable-next-line no-console
```
ericsciple: i think stdout is preferred so unit tests can be less noisy

e.g. here is a snippet from somewhere else in toolkit
```js
        process.stdout.write(`GITHUB_EVENT_PATH ${path} does not exist${EOL}`)
```

ericsciple: not sure, the linter may have another reason for avoiding console.log

konradpabjan: Perhaps we could disable console.log in the jest configuration? https://stackoverflow.com/questions/44467657/jest-better-way-to-disable-console-inside-unit-tests

Some of the response logging that I have is really noisy, for the tests I'll just mock `console.log` so that there is no annoying output

xt0rted: This was a suggestion for handling commands in tests https://github.com/actions/toolkit/issues/118#issuecomment-528888332 and it was later added to https://github.com/actions/cache/blob/master/jest.config.js#L15-L23. I add this to all of my jest configs now. It might be good to add to the TS template.

konradpabjan: I ended up doing the following for each of the tests
 ```
    jest.spyOn(console, 'log').mockImplementation(() => {})
    jest.spyOn(core, 'debug').mockImplementation(() => {})
    jest.spyOn(core, 'info').mockImplementation(() => {})
    jest.spyOn(core, 'warning').mockImplementation(() => {})
```
Shouldn't be any noise in the tests now

PR #307, opened 2020-01-14
generated files

ericsciple:


packages/glob/package-lock.json
```diff
@@ -4,11 +4,6 @@
 	"lockfileVersion": 1,
 	"requires": true,
 	"dependencies": {
-		"@actions/core": {
```
ericsciple: @bryanmacfarlane it looks like your recent commit `audit fix and update http-client` added this. When I build locally it removes it.

Curious what version of npm you have?

I have:
npm --version
6.12.1
node --version
v12.13.1

PR #310, opened 2020-01-15
Update readme with type assertion information

thboop:
Related to #147

### DIFF
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -55,3 +55,19 @@ const newIssue = await octokit.issues.create({
   body: 'Hello Universe!'
 });
 ```
+
+## Webhook Payload Typescript Definitions
+`@octokit/webhooks` provides complete typescript definitions for the payloads associated with each event. You can cast the provided payload to these types for more type information as needed.
+
+To do this, first install the npm module `npm install --save @octokit/webhooks`
+
+Then, assert the type based on the eventName
+```ts
+import * as core from '@actions/core'
+import * as github from '@actions/github'
+import * as Webhooks from '@octokit/webhooks'
+if (github.context.eventName === 'push') {
+  const pushPayload = github.context.payload as Webhooks.WebhookPayloadPush
+  core.info(`The head commit is: ${pushPayload}`)
+}
+```
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -56,10 +56,11 @@ const newIssue = await octokit.issues.create({
 });
 ```
 
-## Webhook Payload Typescript Definitions
-`@octokit/webhooks` provides complete typescript definitions for the payloads associated with each event. You can cast the provided payload to these types for more type information as needed.
+## Webhook payload typescript definitions
 
-To do this, first install the npm module `npm install --save @octokit/webhooks`
+The npm module `@octokit/webhooks` provides type definitions for the response payloads. You can cast the payload to these types for better type information.
+
+First, install the npm module `npm install @octokit/webhooks`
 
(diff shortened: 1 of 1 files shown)


packages/github/README.md
```diff
@@ -55,3 +55,19 @@ const newIssue = await octokit.issues.create({
   body: 'Hello Universe!'
 });
 ```
+
+## Webhook Payload Typescript Definitions
+`@octokit/webhooks` provides complete typescript definitions for the payloads associated with each event. You can cast the provided payload to these types for more type information as needed.
+
+To do this, first install the npm module `npm install --save @octokit/webhooks`
```
ericsciple: is `--save` an option? the [docs](https://docs.npmjs.com/cli/install) dont show that option. `--save-prod` is the default so i dont think we need any save flag at all?

packages/github/README.md
```diff
@@ -55,3 +55,19 @@ const newIssue = await octokit.issues.create({
   body: 'Hello Universe!'
 });
 ```
+
+## Webhook Payload Typescript Definitions
```
ericsciple: Consider sentence-case rather than title-case. Sentence-case seems to be the common trend, e.g. [here](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions)

ericsciple: nit: consider blank line separating `##` header and paragraph (typical markdown style)

packages/github/README.md
ericsciple: which version of octokit?  it should match package.json

packages/github/README.md
```diff
@@ -55,3 +55,19 @@ const newIssue = await octokit.issues.create({
...
+`@octokit/webhooks` provides complete typescript definitions for the payloads associated with each event. You can cast the provided payload to these types for more type information as needed.
+
+To do this, first install the npm module `npm install --save @octokit/webhooks`
+
+Then, assert the type based on the eventName
```
ericsciple: > assert the type

consider "cast to the type". Assert implies validation - connotation from common unit test terminology and also assert module built-in to nodejs

packages/github/README.md
```diff
@@ -55,3 +55,19 @@ const newIssue = await octokit.issues.create({
...
+```ts
+import * as core from '@actions/core'
+import * as github from '@actions/github'
+import * as Webhooks from '@octokit/webhooks'
+if (github.context.eventName === 'push') {
+  const pushPayload = github.context.payload as Webhooks.WebhookPayloadPush
+  core.info(`The head commit is: ${pushPayload}`)
```
ericsciple: did you mean `pushPayload.head`?

packages/github/README.md
```diff
@@ -55,3 +55,19 @@ const newIssue = await octokit.issues.create({
   body: 'Hello Universe!'
 });
 ```
+
+## Webhook Payload Typescript Definitions
+`@octokit/webhooks` provides complete typescript definitions for the payloads associated with each event. You can cast the provided payload to these types for more type information as needed.
```
ericsciple: nit: consider removing unnecessary words. E.g: *The npm module* `@octokit/webhooks` provides ~complete typescript~ *type* definitions for the *response* payloads ~associated with each event~. You can cast the ~provided~ payload to these types for ~more~ *better* type information ~as needed~.

packages/github/README.md
thboop: It isn't a cast though:
https://www.typescriptlang.org/docs/handbook/basic-types.html#type-assertions

PR #311, opened 2020-01-15
Correctly pull issue number for pull request review events

### LINKED ISSUE #289: Issue Number does not populate correctly in `@actions/github` Context.issue

thboop:
Resolves #289

### DIFF
--- a/packages/github/src/context.ts
+++ b/packages/github/src/context.ts
@@ -44,7 +44,7 @@ export class Context {
 
     return {
       ...this.repo,
-      number: (payload.issue || payload.pullRequest || payload).number
+      number: (payload.issue || payload.pull_request || payload).number
     }
   }
 
--- a/packages/github/__tests__/lib.test.ts
+++ b/packages/github/__tests__/lib.test.ts
@@ -52,10 +52,11 @@ describe('@actions/context', () => {
     })
   })
 
-  it('works with pullRequest payloads', () => {
+  it('works with pull_request payloads', () => {
     delete process.env.GITHUB_REPOSITORY
     context.payload = {
-      pullRequest: {number: 2},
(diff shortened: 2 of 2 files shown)


bryanmacfarlane:
Let's get this in since we're about to release an update github pkg for other work (proxy)

packages/github/src/context.ts
```diff
@@ -44,7 +44,7 @@ export class Context {
 
     return {
       ...this.repo,
-      number: (payload.issue || payload.pullRequest || payload).number
+      number: (payload.issue || payload.pull_request || payload).number
```
ericsciple: @thboop a pull request is not an issue?

ericsciple: The issue linked to this PR indicates the \"linked issue\" is desired.

It looks like the change above returns the pull request number, not the inked issue number.

ericsciple: discussed offline. before only PR events works, this change adds PR comment and PR review comments

PR #314, opened 2020-01-17
octokit client should follow proxy settings

ericsciple:


packages/github/src/github.ts
```diff
@@ -17,11 +23,90 @@ export const context = new Context.Context()
...
-  constructor(token: string, opts: Omit<Octokit.Options, 'auth'> = {}) {
-    super({...opts, auth: `token ${token}`})
+  /* eslint-disable no-dupe-class-members */
+  // Disable no-dupe-class-members due to false positive for method overload
+  // https://github.com/typescript-eslint/typescript-eslint/issues/291
 
-    this.graphql = graphql.defaults({
-      headers: {authorization: `token ${token}`}
-    })
+  constructor(token: string, opts?: Octokit.Options)
```
ericsciple: I made a change to allow users to specify their own auth settings (in opts) instead of providing a token.

[Customers who are using other authentication methods](https://github.com/actions/toolkit/issues/106) should still be able to use this class for the proxy benefits.

However I'm not sure what to do yet about graphql. Since the opts parameter is specifically `Octokit.options`.

I need to look closer tomorrow and see if setting graphql `defaults.headers.authorization` equal to octokit `opts.auth` will work.

packages/github/package.json
```diff
@@ -37,8 +37,10 @@
     "url": "https://github.com/actions/toolkit/issues"
   },
   "dependencies": {
+    "@actions/http-client": "^1.0.2",
     "@octokit/graphql": "^4.3.1",
-    "@octokit/rest": "^16.15.0"
+    "@octokit/rest": "^16.15.0",
+    "tunnel": "0.0.6"
```
ericsciple: remove

packages/github/src/github.ts
```diff
@@ -17,11 +22,114 @@ export const context = new Context.Context()
...
+    }
+
+    return typeof options.auth === 'string' ? options.auth : `token ${token}`
+  }
+
+  private static getProxyAgent(
+    options: Octokit.Options
+  ): http.Agent | undefined {
+    if (!options.request?.agent) {
+      const proxyUrl = process.env['https_proxy']
```
ericsciple: also HTTPS_PROXY

packages/github/src/github.ts
```diff
@@ -17,11 +22,114 @@ export const context = new Context.Context()
...
-    })
+  /**
+   * Sets up the REST client and GraphQL client with auth and proxy support.
+   * The parameter `token` or `opts.auth` must be supplied. The GraphQL client
+   * authorization is not setup when `opts.auth` is a function or object.
+   *
+   * @param token  Auth token
+   * @param opts   Octokit options
+   */
+  constructor(token: string, opts?: Octokit.Options)
```
ericsciple: keep `Omit<...>` on the first overload

packages/github/__tests__/github.test.ts
```diff
@@ -0,0 +1,170 @@
+import * as http from 'http'
+import {GitHub} from '../src/github'
+
+describe('@actions/github', () => {
```
ericsciple: New tests to exercise GitHub REST client and GraphQL client with and without proxy

Only runs when env var GITHUB_TOKEN is defined, otherwise warns for discoverability

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
   "compilerOptions": {
+    "esModuleInterop": true,
```
ericsciple: I moved this setting from `packages/github/tsconfig.json` to here.

It's needed for the GitHub unit tests to work. Otherwise the base class properties don't work.

For example, without this setting the unit test fails attempting to access `repos` property fails:

```js
    const octokit = new GitHub(token)
    const branch = await octokit.repos.getBranch({
      owner: 'actions',
      repo: 'toolkit',
      branch: 'master'
    })
```

Typescript [\"highly recommend applying it\"](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html)

packages/tool-cache/src/tool-cache.ts
```diff
@@ -5,7 +5,7 @@ import * as os from 'os'
 import * as path from 'path'
 import * as httpm from '@actions/http-client'
 import * as semver from 'semver'
-import * as uuidV4 from 'uuid/v4'
+import uuidV4 from 'uuid/v4'
```
ericsciple: reaction due to moving `"esModuleInterop": true` to the root `tsconfig.json` - which previously was only defined in `packages/github/tsconfig.json`

tsconfig.json
ericsciple: @hross thoughts regarding consumer impact. I'm assuming shouldnt be any

packages/tool-cache/__tests__/tool-cache.test.ts
```diff
@@ -1,8 +1,8 @@
 import * as fs from 'fs'
-import * as nock from 'nock'
 import * as path from 'path'
 import * as io from '@actions/io'
 import * as exec from '@actions/exec'
+import nock from 'nock'
```
ericsciple: reaction due to moving `"esModuleInterop": true` to the root `tsconfig.json` - which previously was only defined in `packages/github/tsconfig.json`

packages/glob/src/internal-pattern.ts
```diff
@@ -48,8 +48,13 @@ export class Pattern {
   // https://github.com/typescript-eslint/typescript-eslint/issues/291
 
   constructor(pattern: string)
+  constructor(pattern: string, segments: undefined, homedir: string)
```
ericsciple: all the changes to this file are reaction due to moving `"esModuleInterop": true` to the root `tsconfig.json` - which previously was only defined in `packages/github/tsconfig.json`

mocking `os` in the module registry from the unit tests, stopped working when esModuleInterop=true so i switched to a constructor overload pattern

packages/glob/src/internal-path.ts
```diff
@@ -1,6 +1,6 @@
-import * as assert from 'assert'
 import * as path from 'path'
 import * as pathHelper from './internal-path-helper'
+import assert from 'assert'
```
ericsciple: reaction due to moving `"esModuleInterop": true` to the root `tsconfig.json` - which previously was only defined in `packages/github/tsconfig.json`

packages/glob/src/internal-path-helper.ts
```diff
@@ -1,5 +1,5 @@
-import * as assert from 'assert'
 import * as path from 'path'
+import assert from 'assert'
```
ericsciple: reaction due to moving `"esModuleInterop": true` to the root `tsconfig.json` - which previously was only defined in `packages/github/tsconfig.json`

packages/glob/__tests__/internal-pattern.test.ts
```diff
@@ -1,17 +1,9 @@
 import * as io from '../../io/src/io'
+import * as os from 'os'
```
ericsciple: all the changes to this file are reaction due to moving `"esModuleInterop": true` to the root `tsconfig.json` - which previously was only defined in `packages/github/tsconfig.json`

i left a comment on internal-pattern.ts (further below) which explains more

packages/github/tsconfig.json
```diff
@@ -2,7 +2,6 @@
   "extends": "../../tsconfig.json",
   "compilerOptions": {
     "baseUrl": "./",
-    "esModuleInterop": true,
```
ericsciple: moved up to the root tsconfig.json

i left a comment on the root tsconfig.json which explains this more

packages/github/src/github.ts
```diff
@@ -17,11 +22,113 @@ export const context = new Context.Context()
...
-  constructor(token: string, opts: Omit<Octokit.Options, 'auth'> = {}) {
-    super({...opts, auth: `token ${token}`})
+  /* eslint-disable no-dupe-class-members */
+  // Disable no-dupe-class-members due to false positive for method overload
+  // https://github.com/typescript-eslint/typescript-eslint/issues/291
 
-    this.graphql = graphql.defaults({
-      headers: {authorization: `token ${token}`}
-    })
+  /**
```
ericsciple: In addition to setting proxy, I changed the interface so other auth mechanisms can be used. For example see [this issue](https://github.com/actions/toolkit/issues/106).

When customers are using other auth settings, they should still be able to leverage the toolkit to setup proxy.

packages/github/__tests__/proxy.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module 'proxy' {
```
ericsciple: typescript made me do this :(

bryanmacfarlane:
👀

tsconfig.json
bryanmacfarlane: I think it's fine.

PR #317, opened 2020-01-21
release notes

ericsciple:


packages/github/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/github",
-  "version": "2.0.2",
+  "version": "2.0.3",
```
ericsciple: revert

packages/tool-cache/RELEASES.md
```diff
@@ -1,5 +1,9 @@
 # @actions/tool-cache Releases
 
+### 1.3.1
+
+- Increase http-client min version
```
thboop: We should specify the PR, even if it includes other changes: https://github.com/actions/toolkit/pull/314

PR #322, opened 2020-01-26
exec: Expose ExecOptions interface

rhysd:
`@actions/exec` package exposes `exec` function but does not expose `ExecOptions` for now.

Since `exec` function's option is [big](https://github.com/actions/toolkit/blob/d9347d4ab99fd507c0b9104b2cf79fb44fcc827d/packages/exec/src/interfaces.ts#L5), I want to make the option object in advance with some complicated logic.

For example, let's say to have a function to run a command and whether stdout value is captured or not is controlled by flag:

```typescript
async function runCommand(cmd: string, capture?: boolean) {
    let stdout = '';

    const options = { listeners: {} };
    if (capture) {
        options.listeners.stdout = (data) => {
            stdout += data.toString();
        };
    }
    await exec(cmd, [], options);
    return stdout;
}
```

This code cannot be compiled because we cannot give proper type to `options` variable. If the `ExecOptions` interface is available, we can write a safe code as follows:

```typescript
import { exec, ExecOptions } from '@actions/exec';

async function runCommand(cmd: string, capture?: boolean) {
    let stdout = '';

    const options: ExecOptions = { listeners: {} };
    if (capture) {
        options.listeners.stdout = (data) => {
            stdout += data.toString();
        };
    }
    await exec(cmd, [], options);
    return stdout;
}
```

### DIFF
--- a/packages/exec/src/exec.ts
+++ b/packages/exec/src/exec.ts
@@ -1,5 +1,7 @@
-import * as im from './interfaces'
+import {ExecOptions} from './interfaces'
 import * as tr from './toolrunner'
+
+export {ExecOptions}
 
 /**
  * Exec a command.
@@ -14,7 +16,7 @@ import * as tr from './toolrunner'
 export async function exec(
   commandLine: string,
   args?: string[],
-  options?: im.ExecOptions
+  options?: ExecOptions
 ): Promise<number> {
   const commandArgs = tr.argStringToArray(commandLine)
   if (commandArgs.length === 0) {


rhysd:
CI failure seems not related to change in this branch

https://github.com/actions/toolkit/pull/322/checks?check_run_id=409532530#step:5:18

packages/exec/src/exec.ts
```diff
@@ -1,6 +1,8 @@
 import * as im from './interfaces'
 import * as tr from './toolrunner'
 
+export {ExecOptions} from './interfaces'
```
ericsciple: should this import the interface, and then export? for example, see [here](https://github.com/actions/toolkit/blob/master/packages/glob/src/glob.ts)

that way the `exec` function can be declared as:

`export declare function exec(commandLine: string, args?: string[], options?: ExecOptions): Promise<number>;`

(better .d.ts)

rhysd: It looks better for me too. I'll update this branch.

rhysd:
@ericsciple Thank you for your review. I updated this branch in favor of the comment. Would you take a look?

packages/exec/src/exec.ts
```diff
@@ -14,7 +16,7 @@ import * as tr from './toolrunner'
 export async function exec(
   commandLine: string,
   args?: string[],
-  options?: im.ExecOptions
+  options?: ExecOptions
```
bryanmacfarlane: @ericsciple - I don't think this introduces a compat issue since it's a build time thing but I want to make sure.

ericsciple: i think either should work since they have all the same members. i'll test shortly...

ericsciple: Yes, either works. I added another source file sibling to exec:

```js
import * as im from './interfaces'
import * as exec from './exec'

let options: im.ExecOptions = {}
exec.exec("foo", [], options)

let options2: exec.ExecOptions = {}
exec.exec("foo", [], options2)
```

Both ways compile. I think it's because they have all the same members.

Also this is the pattern i used in the glob module. Callers reference interface re-exported from the main file. Internal code references the internal interface.

rhysd:
What is the status of this PR? Is there any action should I do?

rhysd:
Ping @ericsciple @bryanmacfarlane

ericsciple:
Discussed w/ @thboop. I think this is good to merge. I'll defer to Thomas.

jhpratt:
Any chance we can get a release with this?

bryanmacfarlane:
Done.  1.0.4

Issue #184, opened 2019-10-10
Support list input type


bryanmacfarlane:
Thank you 🙇‍♀ for wanting to create an issue in this repository. Before you do, please ensure you are filing the issue in the right place. Issues should only be opened on if the issue **relates to code in this repository**.  

* If you have found a security issue [please submit it here](https://hackerone.com/github)
* If you have questions about writing workflows or action files, then please [visit the GitHub Community Forum's Actions Board](https://github.community/t5/GitHub-Actions/bd-p/actions)
* If you are having an issue or question about GitHub Actions then please [contact customer support](https://help.github.com/en/articles/about-github-actions#contacting-support)

If your issue is relevant to this repository, please include the information below:

**Describe the enhancement**
An action should be able to have an input type which is a list of strings.  The toolkit should provide a getInput overload or new method to retrieve it.

**Code Snippet**
```javascript
let vals: string[] = core.getListInput('inputName');
```

**Additional information**
Add any other context about the feature here.

sarpik:
I was going through https://github.com/actions/cache trying to implement multiple path support, and when going through source code, I only noticed the `core.getInput` function, but not one for `core.getInputs` to get multiple of them.

Then, later, I [found this](https://github.com/actions/cache/blob/master/src/restore.ts#L30-L33):

```ts
const restoreKeys = core
	.getInput(Inputs.RestoreKeys)
	.split("\n")
	.filter(x => x !== "");
```

so, is

```ts
let vals: string[] = core.getInput('inputName').split("\n").filter(x => x !== "");
```

the official way of doing it?

If so, I've created a PR to implement this: **https://github.com/actions/toolkit/pull/336**

AB-xdev:
Was this fixed with #829?

asbjornu:
@AB-xdev, as far as I can tell, #829 implements support for values spanning multiple lines, not lists. With the implemented `core.getMultilineInput()` the following should be possible:

```yaml
- uses: myaction
  with:
    multiline-param: |
      line1
      line2
      line3
```

However, the following is still not possible:

```yaml
- uses: myaction
  with:
    list-param: ['item1', 'item2', 'item3']
```

Neither is this:


```yaml
- uses: myaction
  with:
    list-param:
      - item1
      - item2
      - item3
```

rickstaa:
@asbjornu Good extension! I created #1183 to also implement the second case. Not sure how to implement the last case since these parameters are passed as an environment parameter.

[...]

PR #336, opened 2020-02-06
core: Create `getInputList` utility

kiprasmel:
Inspired by https://github.com/actions/cache/blob/master/src/restore.ts#L30-L33

Would fix https://github.com/actions/toolkit/issues/184,

but there's another issue - https://github.com/actions/cache/issues/44:
an input list should use arrays instead of strings,
separated by newlines (`\n`).


bryanmacfarlane:
yeah, I think the real list type would be cooperative with the yaml, the service and come down in the job message to the runner as a line delimited or array of strings.

sarpik:
@bryanmacfarlane I agree - let's not rush this.

abhilash1in:
@bryanmacfarlane why do you think this may not be the right path?

KaelWD:
This would be useful for me, I'm trying to upload ignored files as artefacts but `actions/upload-artifact` doesn't support a newline-delimited string:

```yml
- run: echo ::set-env name=GITHUB_ARTIFACTS_LOCATIONS::$(git status --ignored --porcelain | grep -oP '(?!.*node_modules\/.*)\!\! \K.*')
- uses: actions/upload-artifact@v2
  with:
    path: ${{ env.GITHUB_ARTIFACTS_LOCATIONS }}
```

```
##[warning]No files were found for the provided path: packages/vuetify/es5/ packages/vuetify/lib-temp/ packages/vuetify/lib/. No artifacts will be uploaded.
```

mmagician:
Any updates on this feature?

thboop:
Hey @kiprasmel , thanks for the contribution, I've made some minor updates and created: https://github.com/actions/toolkit/pull/829/files

Going to merge that pr and release a new version this week. Going to close this out in favor of that.

kiprasmel:
Hey @thboop, that's great - thanks!

PR #340, opened 2020-02-11
@actions/artifact download artifacts

konradpabjan:
Adds the ability to 
- Download a Single Artifact
- Download All artifacts

Also includes
- Documentation for upload and download functions 🎨 [README.md](https://github.com/actions/toolkit/blob/konradpabjan/artifact-download/packages/artifact/README.md) 🎨
- Additional documentation 🖌 [additional-information.md](https://github.com/actions/toolkit/blob/konradpabjan/artifact-download/packages/artifact/additional-information.md) 🖌

Will be used in the v2 versions of `download-artifact` and `upload-artifact`

After these changes the package will be ready to publish 😎

Things missing from this PR that will be added once the package gets published
- Addition to `README.md` in the `src` for toolkit where all the packages get documented

### DIFF
--- a/packages/artifact/README.md
+++ b/packages/artifact/README.md
@@ -34,9 +34,9 @@ Method Name: `uploadArtifact`
 #### Available Options
 
  - `continueOnError`
-    - Indicates if the artifact upload should continue in the event a file or chunk fails to upload. If there is a error during upload, a partial artifact will always be created and available for download at the end. The `size` reported will be the amount of storage that the user or org will be charged for the partial artifact.
-    - If set to `false`, and an error is encountered, all other uploads will stop and any files or chunks that were queued will not be attempted to be uploaded. The partial artifact available will only include files and chunks up until the failure
-    - If set to `true` and an error is encountered, the failed file will be skipped and ignored and all other queued files will be attempted to be uploaded. There will be a an artifact available for download at the end with everything excluding the file/chunks that failed to upload
+    - Indicates if the artifact upload should continue in the event a file fails to upload. If there is a error during upload, a partial artifact will always be created and available for download at the end. The `size` reported will be the amount of storage that the user or org will be charged for the partial artifact.
+    - If set to `false`, and an error is encountered, all other uploads will stop and any files that were queued will not be attempted to be uploaded. The partial artifact available will only include files up until the failure.
+    - If set to `true` and an error is encountered, the failed file will be skipped and ignored and all other queued files will be attempted to be uploaded. There will be an artifact available for download at the end with everything excluding the file that failed to upload
     - Optional, defaults to `true` if not specified
 
 #### Example using Absolute File Paths
--- a/packages/artifact/README.md
+++ b/packages/artifact/README.md
@@ -1,0 +1,228 @@
+# `@actions/artifact`
+
+## Usage
+
+You can use this package to interact with the actions artifact service to upload and download artifacts.
+- [Upload an Artifact](##Upload-an-Artifact)
+- [Download a Single Artifact](##Download-a-Single-Artifact)
+- [Download All Artifacts](##Download-all-Artifacts)
+
+Relative paths and absolute paths are both allowed. Relative paths are rooted against the current working directory.
+
+## Upload an Artifact
+
+#### Inputs
+ - `Name`
+    - The name of the artifact that is being uploaded
+    - Required
+ - `files`
+    - A list of file paths that describe what should be uploaded as part of the artifact
+    - If a path is provided that does not exist, an error will be thrown
+    - Can be absolute or relative. Internally everything is normalized and resolved
+    - Required
+ - `rootDirectory`
+    - A file path that denotes the root directory of the files being uploaded. This path is used to strip the paths provided in `files` to control how they are uploaded and structured
+    - If a file specified in `files` is not in the `rootDirectory`, an error will be thrown
+    - Required
+ - `options`
+    - Extra options that allow for the customization of the upload behavior
+    - Optional
+
+#### Available Options
+
+ - `continueOnError`
+    - Indicates if the artifact upload should continue in the event a file or chunk fails to upload. If there is a error during upload, a partial artifact will always be created and available for download at the end. The `size` reported will be the amount of storage that the user or org will be charged for the partial artifact.
+    - If set to `false`, and an error is encountered, all other uploads will stop and any files or chunks that were queued will not be attempted to be uploaded. The partial artifact available will only include files and chunks up until the failure
+    - If set to `true` and an error is encountered, the failed file will be skipped and ignored and all other queued files will be attempted to be uploaded. There will be a an artifact available for download at the end with everything excluding the file/chunks that failed to upload
+    - Optional, defaults to `true` if not specified
+
+#### Example using Absolute File Paths
+
+```js
+const artifact = require('@actions/artifact');
+const artifactClient = artifact.create()
+const artifactName = 'my-artifact';
+const files = [
+    '/home/user/files/plz-upload/file1.txt',
+    '/home/user/files/plz-upload/file2.txt',
+    '/home/user/files/plz-upload/dir/file3.txt'
+]
+const rootDirectory = '/home/user/files/plz-upload'
+const options = {
+    continueOnError: true
+}
+
+const uploadResult = await artifactClient.uploadArtifact(artifactName, files, rootDirectory, options)
+```
+
+#### Example using Relative File Paths
+```js
+// Assuming the current working directory is /home/user/files/plz-upload
+const artifact = require('@actions/artifact');
+const artifactClient = artifact.create()
+const artifactName = 'my-artifact';
+const files = [
+    'file1.txt',
+    'file2.txt',
+    'dir/file3.txt'
+]
+
+const rootDirectory = '.' // Also possible to specify __dirname if using node
+const options = {
+    continueOnError: false
+}
+
+const uploadResponse = await artifactClient.uploadArtifact(artifactName, files, rootDirectory, options)
+```
+
+#### Upload Result
+
+The returned `UploadResponse` will contain the following information
+
+- `artifactName`
+    - The name of the artifact that was uploaded
+- `artifactItems`
+    - A list of all files that describe what is uploaded if there are no errors encountered. Usually this will be equal to the inputted `files` with the exception of empty directories (will not be uploaded)
+- `size`
+    - Total size of the artifact that was uploaded in bytes
+- `failedItems`
+    - A list of items that were not uploaded successfully (this will include queued items that were not uploaded if `continueOnError` is set to false). This is a subset of `artifactItems`
+
+#### Non-Supported Characters
+
+The inputted `name` and files specified in `files` cannot contain any of the following characters. They will be rejected by the server if attempted to be sent over and the upload will fail. These characters are not allowed due to limitations and restrictions with certain file systems such as NTFS. To maintain platform-agnostic behavior, all characters that are not supported by an individual filesystem/platform will not be supported on all filesystems/platforms.
+
+- "
+- :
+- <
+- \>
+- |
+- \*
+- ?
+- empty space
+
+In addition to the aforementioned characters, the inputted `name` also cannot include the following
+- \
+- /
+
+
+#### Permission loss
+
+Artifacts are uploaded and stored in blob storage rather than a full fledged file system. Because of this, file permissions are not always maintained between uploaded and downloaded artifacts. If file permissions are something that need to be maintained (such as an executable), consider archiving all of the files using something like `tar` and then uploading the single archive. After downloading the artifact, you can `un-tar` the individual file and permissions will be preserved.
+
+```js
+const artifact = require('@actions/artifact');
+const artifactClient = artifact.create()
+const artifactName = 'my-artifact';
+const files = [
+    '/home/user/files/plz-upload/my-archive.tgz',
+]
+const rootDirectory = '/home/user/files/plz-upload'
+const uploadResult = await artifactClient.uploadArtifact(artifactName, files, rootDirectory)
+```
+
+#### Considerations related to the number of files being uploaded
+
+During upload, each file is uploaded concurrently in 4MB chunks using a separate HTTPS connection per file. Chunked uploads are used so that in the event of a failure (which is entirely possible because the internet is not perfect), the upload can be retried. If there is an error, a retry will be attempted after waiting for 10 seconds.
+
+Uploading will be generally be faster if there are fewer files that are larger in size vs if there are lots of smaller files. Depending on the types and quantities of files being uploaded, it might be beneficial to separately compress and archive everything into a single archive (using something like `tar` or `zip`) before starting and artifact upload to speed things up.
+
+## Download a Single Artifact
+
+#### Inputs
+ - `Name`
+    - The name of the artifact that is being downloaded
+    - Required
+ - `path`
+    - Path that denotes where the artifact will be downloaded to
+    - Optional. Defaults to the GitHub workspace directory(`$GITHUB_WORKSPACE`) if not specified
+ - `options`
+    - Extra options that allow for the customization of the download behavior
+    - Optional
+
+
+#### Available Options
+
+ - `createArtifactFolder`
+    - Specifies if a folder (the artifact name) is created for the artifact that is downloaded (contents downloaded into this folder),
+    - Optional. Defaults to false if not specified
+
+#### Example
+
+```js
+const artifact = require('@actions/artifact');
+const artifactClient = artifact.create()
+const artifactName = 'my-artifact';
+const path = 'some/directory'
+const options = {
+    createArtifactFolder: false
+}
+
+const downloadResponse = await artifactClient.downloadArtifact(artifactName, path, options)
+// Post download, the directory structure will look like this
+/some
+    /directory
+        /file1.txt
+        /file2.txt
+        /dir
+            /file3.txt
+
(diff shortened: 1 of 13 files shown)


packages/artifact/src/internal-artifact-client.ts
```diff
@@ -100,25 +133,114 @@ export class DefaultArtifactClient implements ArtifactClient {
...
+      )
+    }
+
+    const artifactToDownload = artifacts.value.find(artifact => {
+      return artifact.name === name
+    })
+
+    if (!artifactToDownload) {
+      throw new Error(`Unable to find an artifact with the name: ${name}`)
+    } else {
```
yacaovsnc: Don't need `else` if you throw in the above `if` clause.   The code is easier to read with fewer braces.

packages/artifact/src/internal-artifact-client.ts
```diff
@@ -100,25 +133,114 @@ export class DefaultArtifactClient implements ArtifactClient {
...
 
-  export async function downloadArtifact(
-      name: string,
-      path?: string,
-      options?: DownloadOptions
-    ): Promise<DownloadResponse> {
+      // Check only for downloadable items that are not directories. During upload, empty directories are rejected by the remote server so there should be no
+      // artifacts that consist of only empty directories
+      if (items.value.length === 0) {
+        throw new Error(
```
yacaovsnc: I don't think we should throw here.  If the artifact is empty it is empty.  We can log an info here saying no items were found.  

The above will happen if a file is uploaded partially and never finished.

packages/artifact/src/internal-artifact-client.ts
```diff
@@ -100,25 +133,114 @@ export class DefaultArtifactClient implements ArtifactClient {
...
   }
 
-  Downloads all artifacts associated with a run. Because there are multiple artifacts being downloaded, a folder will be created for each one in the specified or default directory
+  async downloadAllArtifacts(
+    path?: string | undefined
+  ): Promise<DownloadResponse[]> {
+    const response: DownloadResponse[] = []
+    const artifacts = await listArtifacts()
+    if (artifacts.count === 0) {
+      throw new Error(
```
yacaovsnc: Should not throw here, just log an info message.  We throw if `listArtifacts` throw, otherwise this is perfectly fine.  The action itself can decide if they want to throw or not.

packages/artifact/src/internal-artifact-client.ts
```diff
@@ -100,25 +133,114 @@ export class DefaultArtifactClient implements ArtifactClient {
...
+            downloadedArtifacts += 1
+
+            // Promise.All is not correctly inferring that 'path' is no longer possibly undefined: https://github.com/microsoft/TypeScript/issues/34925
+            const downloadSpecification = getDownloadSpecification(
+              currentArtifactToUpload.name,
+              items.value,
+              path!, // eslint-disable-line @typescript-eslint/no-non-null-assertion
+              true
+            )
+            await createDirectoriesForArtifact(
```
yacaovsnc: This will not work cross platformly because the backing end is case insensitive.   We need to create folders as we download each individual files.

packages/artifact/src/internal-download-specification.ts
```diff
@@ -0,0 +1,73 @@
...
+      // Ignore other artifacts in the container that don't begin with the same name
+    } else {
+      // entry.path always starts with the artifact name, if includeRootDirectory is false, remove the name from the beginning of the path
+      const filePath = path.join(
+        downloadPath,
+        includeRootDirectory ? entry.path : entry.path.replace(artifactName, '')
+      )
+
+      if (entry.itemType === 'folder') {
+        directories.push(filePath)
```
yacaovsnc: This is where we will miss case insensitive folders.

packages/artifact/src/internal-artifact-client.ts
konradpabjan: The list of directories now comes from each individual file, I switched over to using `path.dirname` if the `itemType` is a file. Had to modify some tests and after some local testing it still works 😀 

This change has the side effect of not making every sub-directory be included in the `directoryStructure` result, but that's okay because everything gets created recursively

Example, before for the files `root/file1.txt` and `root/dir1/dir2/file2.txt` we would get 3 directories `root/ root/dir1/, root/dir1/dir2/`, now we only get 2: `root` and `root/dir1/dir2`

The directories still get created before the individual download starts (this allows for easier testing)

packages/artifact/src/internal-download-specification.ts
konradpabjan: Modified, should work now. See comment above 😀

packages/artifact/src/internal-artifact-client.ts
konradpabjan: Updated. Just `core.info` is used now and an empty list is returned

packages/artifact/src/internal-artifact-client.ts
konradpabjan: Changed it to just `core.info`. The returned `DownloadResponse` will be the same so it will be up to the user to figure out that the artifact is actually empty if they don't look at the logging somehow

packages/artifact/src/internal-artifact-client.ts
```diff
@@ -100,25 +133,113 @@ export class DefaultArtifactClient implements ArtifactClient {
...
+    if (!path) {
+      path = getWorkSpaceDirectory()
+    }
+    path = normalize(path)
+    path = resolve(path)
 
-      TODO
+    // Get container items for all artifacts. The resource url is the same for all artifacts so one call is sufficient
+    const items = await getContainerItemsForAllArtifacts(
+      artifacts.value[0].fileContainerResourceUrl
```
yacaovsnc: The container should have a `containerId` field, we should make one call per `containerId`.

yacaovsnc: The principle is we should not assume any server-side implementation details if possible.

packages/artifact/src/internal-artifact-client.ts
```diff
@@ -100,25 +133,113 @@ export class DefaultArtifactClient implements ArtifactClient {
...
+    if (items.count === 0) {
+      core.warning(`No downloadable files were found for any artifacts`)
+    } else {
+      const ARTIFACT_CONCURRENCY = getDownloadArtifactConcurrency()
+      const parallelDownloads = [...new Array(ARTIFACT_CONCURRENCY).keys()]
+      let downloadedArtifacts = 0
+      await Promise.all(
+        parallelDownloads.map(async () => {
+          while (downloadedArtifacts < artifacts.count) {
+            const currentArtifactToUpload = artifacts.value[downloadedArtifacts]
```
yacaovsnc: Copy / pasted var name. :D

packages/artifact/src/internal-download-specification.ts
```diff
@@ -0,0 +1,77 @@
...
+  const specifications: DownloadSpecification = {
+    rootDownloadLocation: includeRootDirectory
+      ? path.join(downloadPath, artifactName)
+      : downloadPath,
+    directoryStructure: [],
+    filesToDownload: []
+  }
+
+  for (const entry of artifactEntries) {
+    if (!entry.path.startsWith(artifactName)) {
```
yacaovsnc: Is this right?  What if the entry is `abc/def` and the artifact name is `a`?

packages/artifact/src/internal-download-specification.ts
```diff
@@ -0,0 +1,77 @@
...
+          sourceLocation: entry.contentLocation,
+          targetPath: filePath
+        })
+      } else if (entry.itemType === 'folder') {
+        // case insensitive folder structure maintained in the backend, not every folder is created so the
+        // path for the file must be used to determine the directory structure
+      } else {
+        // eslint-disable-next-line no-console
+        console.log(entry)
+        throw new Error(`Unsupported item type: ${entry.itemType}`)
```
yacaovsnc: Why do we throw?  Do we really care in this case?

packages/artifact/src/internal-artifact-client.ts
konradpabjan: So `fileContainerResourceUrl` is effectively just `${runtimeUrl}_apis/resources/Containers/${containerId}`.  Both the `containerId` and the `fileContainerResourceUrl` are the same for all the artifacts since the same container is shared.

Regardless if we use `containerId` + `runtimeUrl` or just the `fileContainerResourceUrl` URL, in both cases we need to know that artifacts for the same run share the same container

packages/artifact/README.md
```diff
@@ -0,0 +1,228 @@
...
+- [Upload an Artifact](##Upload-an-Artifact)
+- [Download a Single Artifact](##Download-a-Single-Artifact)
+- [Download All Artifacts](##Download-all-Artifacts)
+
+Relative paths and absolute paths are both allowed. Relative paths are rooted against the current working directory.
+
+## Upload an Artifact
+
+#### Inputs
+ - `Name`
```
joshmgross:
```suggestion
 - `name`
```
We should be consistent with the casing of our inputs

packages/artifact/README.md
```diff
@@ -0,0 +1,228 @@
...
+const artifact = require('@actions/artifact');
+const artifactClient = artifact.create()
+const artifactName = 'my-artifact';
+const files = [
+    'file1.txt',
+    'file2.txt',
+    'dir/file3.txt'
+]
+
+const rootDirectory = '.' // Also possible to specify __dirname if using node
```
joshmgross: "if using node"

Aren't all uses Node?

packages/artifact/README.md
```diff
@@ -0,0 +1,228 @@
...
+- empty space
+
+In addition to the aforementioned characters, the inputted `name` also cannot include the following
+- \
+- /
+
+
+#### Permission loss
+
+Artifacts are uploaded and stored in blob storage rather than a full fledged file system. Because of this, file permissions are not always maintained between uploaded and downloaded artifacts. If file permissions are something that need to be maintained (such as an executable), consider archiving all of the files using something like `tar` and then uploading the single archive. After downloading the artifact, you can `un-tar` the individual file and permissions will be preserved.
```
joshmgross: "Artifacts are uploaded and stored in blob storage rather than a full fledged file system"

I think this statement is misleading. Each file is uploaded to blog storage separately, but internally we mimic a file system. Regardless, I don't think the implementation details are necessary to say "file permissions are not preserved, use tar if you want them"

packages/artifact/README.md
```diff
@@ -0,0 +1,228 @@
...
+const files = [
+    '/home/user/files/plz-upload/my-archive.tgz',
+]
+const rootDirectory = '/home/user/files/plz-upload'
+const uploadResult = await artifactClient.uploadArtifact(artifactName, files, rootDirectory)
+```
+
+#### Considerations related to the number of files being uploaded
+
+During upload, each file is uploaded concurrently in 4MB chunks using a separate HTTPS connection per file. Chunked uploads are used so that in the event of a failure (which is entirely possible because the internet is not perfect), the upload can be retried. If there is an error, a retry will be attempted after waiting for 10 seconds.
```
joshmgross: I think this may be too much detail for a package consumer who's goal is to learn the relevant APIs and start using artifacts in their action.

Implementation details are useful when there's a bug or a complaint about performance. It might be good to break this out into a separate advanced section if you think it's worth documenting (since the code is open-source that may not be necessary to go this much detail.)

packages/artifact/src/internal-download-specification.ts
konradpabjan: Good catch, I changed it to check `${artifactName}/` or `${artifactName}\` and updated the tests to very the behavior

The `entry.path` is always returned with a forward slash but I didn't want to hardcode that assumption in so I added checks for both

packages/artifact/src/internal-download-specification.ts
konradpabjan: Got rid of this error. Did some digging on the server-side and we throw an error if a non `folder` or `file` type is hit beforehand so this was redundant too

packages/artifact/src/internal-artifact-client.ts
konradpabjan: Did some changes to make one call per artifact with each `fileContainerResourceUrl`. Removes the assumptions related to how the artifacts get stored

packages/artifact/README.md
konradpabjan: I got rid of the server implementation details and moved this section over entirely to anther documentation page

packages/artifact/README.md
konradpabjan: Moved this over to another page so if anyone is curious about this information they can dig it up (or we can link to it if need be). I always lean to the side of more documentation is better.

packages/artifact/README.md
```diff
@@ -0,0 +1,199 @@
...
+    - Required
+ - `options`
+    - Extra options that allow for the customization of the upload behavior
+    - Optional
+
+#### Available Options
+
+ - `continueOnError`
+    - Indicates if the artifact upload should continue in the event a file or chunk fails to upload. If there is a error during upload, a partial artifact will always be created and available for download at the end. The `size` reported will be the amount of storage that the user or org will be charged for the partial artifact.
+    - If set to `false`, and an error is encountered, all other uploads will stop and any files or chunks that were queued will not be attempted to be uploaded. The partial artifact available will only include files and chunks up until the failure
```
yacaovsnc: Don't mention failed chunks - it won't be available for download as the file will be marked `pending` internally.

packages/artifact/README.md
```diff
@@ -0,0 +1,199 @@
...
+    - If set to `true` and an error is encountered, the failed file will be skipped and ignored and all other queued files will be attempted to be uploaded. There will be a an artifact available for download at the end with everything excluding the file/chunks that failed to upload
```
yacaovsnc: > There will be a an artifact

Typo

packages/artifact/src/internal-config-variables.ts
```diff
@@ -10,6 +12,16 @@ export function getUploadChunkSize(): number {
   return 4 * 1024 * 1024 // 4 MB Chunks
 }
 
+export function getDownloadFileConcurrency(): number {
+  // limit the number of concurrent file downloads per artifact to the number of CPU cores available
+  return cpus().length
```
yacaovsnc: We should still do 2 here.  The concern isn't about the client - it's about how many connections we open on the server.

packages/artifact/src/internal-download-http-client.ts
```diff
@@ -0,0 +1,132 @@
...
+  const resourceUrl = new URL(containerUrl)
+  resourceUrl.searchParams.append('itemPath', artifactName)
+
+  const client = createHttpClient()
+  const rawResponse = await client.get(resourceUrl.toString())
+  const body: string = await rawResponse.readBody()
+  if (isSuccessStatusCode(rawResponse.message.statusCode) && body) {
+    return JSON.parse(body)
+  }
+  throw new Error(`Unable to get ContainersItems from ${resourceUrl}`)
```
yacaovsnc: You have different styles within the same file.  :)  Why not log the raw response here too like you did on line 34?

packages/artifact/src/internal-config-variables.ts
konradpabjan: For v1 though, the processor count is used to control how many concurrent downloads are going on: https://github.com/actions/runner/blob/6c70d53eead402ba5d53676d6ed649a04e219c9b/src/Runner.Plugins/Artifact/FileContainerServer.cs#L171

packages/artifact/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Upload an Artifact
+
+Method Name: `uploadArtifact`
+
+#### Inputs
+ - `name`
+    - The name of the artifact that is being uploaded
+    - Required
+ - `files`
```
chrispat: In the new version of the artifact action we discussed supporting wild cards.  Is the intent here that an action that uses this library can support wildcards but it must pass fully qualified paths to the module?

packages/artifact/src/internal-config-variables.ts
yacaovsnc: Maybe that's why we see those high cpu/db load... :D

packages/artifact/src/internal-download-specification.ts
yacaovsnc: Can you point to me where you added the `/` to artifact names?

packages/artifact/README.md
konradpabjan: Originally I had the wildcard search directly backed into this package using `@actions/glob`, however, we decided to move it out (that will go into `@actions/upload-artifact` which will then call this package with the search result). This gives users more flexibility in terms of how they use this package, they can even use something else for search if `@actions/glob` doesn't fit their needs. This package just solely focuses on interactions with artifacts at a low level. 

Both relative and absolute paths work (everything gets normalized and resolved behind the scenes), so fully qualified paths are not required

packages/artifact/src/internal-config-variables.ts
konradpabjan: Perhaps 🙄 I'll bump it down to 2, we can always adjust the number in the future

packages/artifact/src/internal-download-http-client.ts
konradpabjan: Updated!

packages/artifact/src/internal-download-specification.ts
konradpabjan: Line 49 & 50 `internal-download-specification.ts` The forward and backslash only get added during this check

PR #344, opened 2020-02-14
Update unit-tests.yml

sglclm:


thboop:
@sglclm what is the motivation for this change?

Issue #164, opened 2019-09-29
Add stdin support to exec


domenkozar:

bryanmacfarlane:
recasting as a feature request.  If you can add any details or scenarios to the description, that would be appreciated.

clareliguori:
+1 As an example, it would be nice to pass a password over stdin for docker, i.e. `docker login -u username --password-stdin`

Issue #346, opened 2020-02-18
Support piped commands 


lstocchi:
**Describe the enhancement**
Support pipes `|` to redirect stdout of first toolrunner to the next one.

**Code Snippet**
An idea could be to have a method that connects a list of ToolRunners and start executing them `ToolRunner.execWithPiping(trs: ToolRunner[])`

**Additional information**
If i have a sequence of commands like `tr1 | tr2 | tr3` i would like to redirect stdout through all tollrunners. (e.g. `kubectl create deployment my-dep -o yaml --dry-run | kubectl label --local -f - environment=qa -o yaml | kubectl create -f -` )

I am also available to contribute on this if there is no workaround to achieve this and you want to add this feature.

sittie82:
An idea could be to have a method that connects a list and start executing them ToolRunner.execWithPiping(trs: ToolRunne)

rickstaa:
Related to #346, #461, #359. This would be a very nice feature to have I currently use the following workaround that was given by @hrueger.

*LINUX/MAC*:

`````python
exec.exec(`/bin/bash -c "my command"`)
```

*WINDOWS*:

`````python
exec.exec(`cmd /c "my command"`)
```

PR #360, opened 2020-02-20
Allow specifying stdin

sethvargo:
Fixes GH-164


packages/exec/src/interfaces.ts
```diff
@@ -30,6 +30,9 @@ export interface ExecOptions {
   /** optional. How long in ms to wait for STDIO streams to close after the exit event of the process before terminating. defaults to 10000 */
   delay?: number
 
+  /** optional. input to write to the process on STDIN. */
+  stdin?: Buffer
```
sethvargo: I kept going back-and-forth about whether to accept a Buffer or Stream here. I think Buffer is a better API for most users, but it's not quite as flexible as giving users the ability to specify a stream.

sethvargo:
Test failure seems unrelated?

joshmgross:
👋 Thanks for this! I've added @bryanmacfarlane and @ericsciple to help review this change.

sittie82:
@actions/tool-cache

sittie82:
sethvargo:sethvargo/stdin

ericsciple:
adding to @thboop's radar

ericsciple:
Would be great to solve in a way that enables https://github.com/actions/toolkit/issues/346 (chain output of one tool runner to input of second)

Might be good to start with generic bare minimum to enable the scenario, before adding specific helpers like \"execWithPiping\".

Workaround today, piping can be achieved today by running command line, but then lose arg escaping goodness built into tool runner.

Extremely leary of adding piping into toolrunner. Was in a lib before and weird race never solved. I think the second process shouldnt be started until output is actually ready from the first process? Worried about adding solution into toolrunner but not finishing.

sethvargo:
Hey @ericsciple thanks for the feedback. I think this does solve it in a way that supports #346. This is a lower-level API that enables Action authors to accept an input stream in their action. The framework itself could be extended to leverage this API to chain action output, but I don't think that belongs in this PR.

The proposed solution in #346 wouldn't solve the use case this PR is solving. I have actions that invoke a command. The command accepts either a filepath on disk or input on stdin. The current action performs additional work to write the contents to disk before executing the command, but the extra steps could be removed if the toolrunner accepted a stream instead.

- It's not feasible to change the command's behavior
- Accepting stdin on a pipe is quite common among many commands
- Altering the command to run `echo 'thing' | command` isn't cross-platform compatible and relies on shell-specifics (like pipes), in addition to your mention of arg escaping; spawning that additional subcommand is an unnecessary fork if the process just accepts stdin
- Writing files to disk is sometimes seen as a security hole, and the Actions framework makes it trivally easy to create a world-readable file

ericsciple:
@sethvargo sorry i intended much of the feedback for @thboop to consider when reviewing.

@thboop two scenarios i think we should cover: stdin, and piping.

craigdbarber:
@imjohnbo could you PTAL.

packages/exec/__tests__/exec.test.ts
```diff
@@ -286,6 +286,27 @@ describe('@actions/exec', () => {
...
+      __dirname,
+      'scripts',
+      'wait-for-input.sh'
+    )
+
+    const _testExecOptions = getExecOptions()
+
+    _testExecOptions.listeners = {
+      stdout: (data: Buffer) => {
+        expect(data).toEqual(Buffer.from('this is my input\n'))
```
thboop, line 300: This changes based on OS. On windows this should be `'this is my input\r\n'`

packages/exec/__tests__/exec.test.ts
```diff
@@ -286,6 +286,27 @@ describe('@actions/exec', () => {
     expect(stderrCalled).toBeTruthy()
   })
 
+  it('Handles stdin', async () => {
```
thboop, line 289: This test isn't currently working on windows. You may want to create a 
wait-for-input.cmd
```
@echo off
set /p input=
echo %input%
``` 
and invoke that if `IS_WINDOWS` is true. We do a similar thing here: https://github.com/actions/toolkit/blob/master/packages/exec/__tests__/exec.test.ts#L453

For example
```
    let command: string
    if (IS_WINDOWS) {
      command = './wait-for-input.cmd'
    } else {
      command = './wait-for-input.sh'
    }
    const waitForInput: string = path.join(
      __dirname,
      'scripts',
      command
    )
```

packages/exec/src/interfaces.ts
thboop: I agree that buffer is the better api for most users. Its simple and should work for most use cases here. If we end up implementing more complicated scenarios with streams in the future, we can always support passing in a stream or a buffer mutually exclusive. We some use cases for [that scenario](https://github.com/actions/toolkit/issues/346)

@bryanmacfarlane , FYI.

packages/exec/src/toolrunner.ts
```diff
@@ -524,6 +524,16 @@ export class ToolRunner extends events.EventEmitter {
...
       })
+
+      if (this.options.stdin) {
+        if (!cp.stdin) {
+          throw new Error('child process missing stdin')
+        }
+
+        cp.stdin.setDefaultEncoding('utf-8')
+        cp.stdin.write(this.options.stdin)
+        cp.stdin.end()
```
thboop, line 535: Nit: end allows us to write write a chunk, we don't seem to need the `stdin.write` call

packages/exec/__tests__/exec.test.ts
sethvargo: Used `os.EOL`

packages/exec/__tests__/exec.test.ts
sethvargo: Done

packages/exec/src/toolrunner.ts
sethvargo: I seem to recall having issues with that, but I don't remember why. Updated.

sethvargo:
@thboop updated. seems happy now.

packages/exec/src/toolrunner.ts
```diff
@@ -524,6 +524,15 @@ export class ToolRunner extends events.EventEmitter {
           resolve(exitCode)
         }
       })
+
+      if (this.options.stdin) {
+        if (!cp.stdin) {
+          throw new Error('child process missing stdin')
+        }
+
+        cp.stdin.setDefaultEncoding('utf-8')
```
ericsciple, line 533: Does `setDefaultEncoding` need to be called? I would think the buffer would just be written verbatim. Or does calling `setDefaultEncoding` cause a BOM to be written first?

sethvargo: Honest answer: every single example calls `setDefaultEncoding`, and the documentation for the behavior without seems undefined. My best guess is that it falls back to the OS encoding, which can different between OSes (Windows is UTF-16 IIRC).

As far as I can tell, it doesn't send a BOM. I'm open to changing it, because I admittedly don't understand it.

ericsciple: Thanks! Makes sense a BOM would not be sent. I think BOMs are typically used for file streams only, not stdin/stdout.

Based on the docs for [`stream.Writable.write()`](https://nodejs.org/api/stream.html#stream_writable_write_chunk_encoding_callback) my interpretation is `setDefaultEncoding` is only useful when `write` is called with a string, and encoding is not specified.

In our case, the caller is always providing a Buffer (raw bytes). I think it's safe to omit setting the default encoding. Omitting the line shouldnt change behavior, but might avoid confusion.

sethvargo: Done

packages/exec/__tests__/exec.test.ts
```diff
@@ -286,6 +286,30 @@ describe('@actions/exec', () => {
...
+    let command: string
+    if (IS_WINDOWS) {
+      command = 'wait-for-input.cmd'
+    } else {
+      command = 'wait-for-input.sh'
```
bryanmacfarlane, line 294: I would like to see a test that shows what an action author would do.  e.g. all js instead of the sh and cmd scripts.

packages/exec/src/interfaces.ts
bryanmacfarlane, line 34: https://nodejs.org/api/child_process.html#child_process_child_process_execfilesync_file_args_options

I think we should follow the pattern here and consider calling it options.input

bryanmacfarlane: Are we sure this is a buffer?  subprocess.stdin advertises it as a writable stream.  Another reason I wanted to see a pure js test instead of the script.  https://nodejs.org/api/child_process.html#child_process_subprocess_stdin

packages/exec/src/toolrunner.ts
```diff
@@ -524,6 +524,14 @@ export class ToolRunner extends events.EventEmitter {
           resolve(exitCode)
         }
       })
+
+      if (this.options.stdin) {
+        if (!cp.stdin) {
```
bryanmacfarlane, line 529: is this a buffer or a stream?  https://nodejs.org/api/child_process.html#child_process_subprocess_stdin

packages/exec/src/interfaces.ts
thboop: The input we are allowing a user to pass in here is a buffer.

We write the contents of that buffer to the process's StdIn Stream.

Very similar to how node does this with the `input` option you outlined below, which also takes a string and various other inputs.
https://nodejs.org/api/child_process.html#child_process_child_process_execfilesync_file_args_options

packages/exec/src/toolrunner.ts
thboop: Responded to this above: https://github.com/actions/toolkit/pull/360#discussion_r406372534

packages/exec/src/interfaces.ts
thboop: Agreed, great suggestion

sethvargo: Done

packages/exec/__tests__/exec.test.ts
sethvargo: I added a js-specific test, although I admit I'm not sure exactly what you're asking. All of these tests spawn a childprocess using exec - there's no "all js". The test spawns a node subprocess with a node file that prints what it gets on stdin.

packages/exec/src/interfaces.ts
ericsciple: Sorry I just realized node also has `input` option. Should we just pass the value through and let node handle writing to the stream?

ericsciple: ...that said, seems like would be an implementation detail and wouldn't change behavior exposed by this PR. So i think this PR is OK the way it is (incremental)

thboop: `input` doesn't appear to be available for any of the async process creation function. In particular we use spawn. https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options

ericsciple: oh weird, didnt realize it was only on the sync functions. makes sense now why we dont pass thru :)

bryanmacfarlane:
added to todo list for one last look

sethvargo:
Yay! Looks like we're all green 🟢

Issue #361, opened 2020-02-24
core.getInput treats YAML booleans as strings


andrewdhazlett:
**Describe the bug**
When I try to use a boolean value in my `action.yaml` file or workflow definition and retrieve it using `core.getInput` the result is a string.

**To Reproduce**
1. Go to 'https://github.com/andrewdhazlett/gh-action-boolean-input-repro'
2. Click on 'Actions > `gh-action-boolean-input-repro` > Bug repro'
3. Scroll down to `Run /./`

**Expected behavior**
No error is thrown, the input is interpreted as a boolean

ericsciple:
`getInput` always returns a string by design. The inputs are passed as env var across process.

You can use a pattern like `getInput('my-input').toUpper() === 'true'` to convert to boolean.

Workflow uses yaml 1.2 so all the yaml 1.1 boolean values are intentionally not supported.

andrewdhazlett:
@ericsciple please also see my response on the PR I submitted.

I don't see why the implementation detail of passing env vars should be of concern to the consuming code. The values are defined in a format that supports non-string scalar types, so I expect those types to be available in the consuming code. Is there anywhere this design decision is documented?

likern:
> @ericsciple please also see my response on the PR I submitted.
> 
> I don't see why the implementation detail of passing env vars should be of concern to the consuming code. The values are defined in a format that supports non-string scalar types, so I expect those types to be available in the consuming code. Is there anywhere this design decision is documented?

The whole API is **very-very badly written and designed**. I would highly recommend to find out Typescript / Javascript expert in your team (I think @JasonEtco is that guy)  to do that from scratch. Otherwise you have to spend years to fix up GitHub Actions on top of that errorneous foundation.

I started writing my own action and struggle with that decision every minute. The whole idea of GitHub Actions is awesome, but developing experience not so much. 

I wanted to write action to setup React Native on top of **@actions/setup-java** action. And what I found out.

### Actions are not composable
**I can't call other action from my own action**. React Native **requires Java to be already installed**.
In normal scenario user would expect to do something like this:
```yaml
jobs:
  build:
    name: Build library
    runs-on: ubuntu-20.04
    steps:
      - name: Checkout repository
        uses: actions/checkout@v2

      - name: Setup Java environment
        uses: actions/setup-java@v1
        with:
          java-version: ${{ env.JAVA_VERSION }}

      - name: Setup React Native environment
        uses: actions/setup-react-native@v1
        with:
[...]

PR #362, opened 2020-02-24
fix(core): handle boolean input values from yaml

andrewdhazlett:
Fixes https://github.com/actions/toolkit/issues/361

### DIFF
--- a/packages/artifact/package-lock.json
+++ b/packages/artifact/package-lock.json
@@ -4,11 +4,6 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "@actions/core": {
-      "version": "1.2.1",
-      "resolved": "https://registry.npmjs.org/@actions/core/-/core-1.2.1.tgz",
-      "integrity": "sha512-xD+CQd9p4lU7ZfRqmUcbJpqR+Ss51rJRVeXMyOLrZQImN9/8Sy/BEUBnHO/UKD3z03R686PVTLfEPmkropGuLw=="
-    },
     "@actions/http-client": {
       "version": "1.0.6",
       "resolved": "https://registry.npmjs.org/@actions/http-client/-/http-client-1.0.6.tgz",
--- a/packages/core/__tests__/core.test.ts
+++ b/packages/core/__tests__/core.test.ts
@@ -18,6 +18,8 @@ const testEnvVars = {
   INPUT_MISSING: '',
   'INPUT_SPECIAL_CHARS_\'\t"\\': '\'\t"\\ response ',
   INPUT_MULTIPLE_SPACES_VARIABLE: 'I have multiple spaces',
+  INPUT_YAML_BOOLEAN_FALSE: 'OFF',
+  INPUT_YAML_BOOLEAN_TRUE: 'y',
 
   // Save inputs
   STATE_TEST_1: 'state_val'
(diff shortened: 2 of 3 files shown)


ericsciple:
Workflow uses yaml 1.2, so all the yaml 1.1 boolean values should be strings not booleans

getInput always returns a string by design.

andrewdhazlett:
@ericsciple I'm not really sure what you mean about yaml 1.1 vs 1.2 values. From reading the spec it seems like 1.2 supports the same set of scalar and collection values as 1.2, which makes sense as they all map cleanly onto JSON types.

Regardless, can we at least map booleans to their canonical representations so I can just handle `"true"` and `"false"` in the consuming code?

Is there anywhere that this is documented? I've had issues in the past determining what subset of yaml GitHub supports, and even a link to the relevant source would be super helpful.

Thanks!

bryanmacfarlane:
@andrewdhazlett - see actions/checkout and actions/go for examples of handling booleans.

I think we should add a `core.getBooleanInput` method since the action author knows what's expected.  

Inputs are set over envvars so that's why the marshalling is a string.

andrewdhazlett:
@bryanmacfarlane thanks for the reference. After some more digging, it seems that only the [failsafe schema](https://yaml.org/spec/1.2/spec.html#schema/failsafe/) is supported. It would be nice to reference this in the documentation. Happy to submit a PR for that as well.

Looking at #336 it looks like the design around this is still a bit up in the air, can you give me any insight into where these API improvements lie on the roadmap?

ericsciple:
@andrewdhazlett sorry i forgot about the different schemas... The yaml 1.2 core schema is implemented, let me know if a specific form isnt working. Good point about docs, i can skim and see if it's mentioned anywhere.

jclem:
👋 I'm not a Toolkit reviewer at this time, I'll leave it to those who are! cc @ericsciple

ericsciple:
@rachmari @chrispat is there a good place in the docs we can clarify we use [YAML 1.2 \"core schema\"](https://yaml.org/spec/1.2/spec.html#id2804923) for tag resolution.

YAML 1.1 supported boolean values like \"off\" but YAML 1.2 dropped support.

YAML 1.2 has three different tag resolution schemas. Whoever implements a YAML 1.2 parser may choose not to implement the full \"core schema\" tag resolution. \"Core schema\" \> \"JSON schema\" \> \"failsafe schema\"

For example, in \"core schema\":

```yaml
my-key-1: true    # value interpreted as a boolean
my-key-2: TRUE    # value interpreted as a boolean
```

But for parsers that only implement \"JSON schema\" or \"failsafe schema\":

```yaml
my-key-1: true    # value interpreted as a boolean
my-key-2: TRUE    # value interpreted as a string
```

It doesnt come up often, so I wouldnt necessarily lead with it. But feels like we should have a note about it somewhere in the docs.

ericsciple:
also @lucascosti regarding the ^ comment

lucascosti:
> Is there a good place in the docs we can clarify we use YAML 1.2 "core schema" for tag resolution.

👋 @ericsciple If you'd like, we could insert more information about the YAML version/schema in the workflow syntax article: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#about-yaml-syntax-for-workflows

Open an issue in the docs repo with the details and we'll take care of it 👍 🙂

ericsciple:
@lucascosti thanks! will do

WilliamMichelMSFT:
@lucascosti @ericsciple Any updates on this?

thboop:
Closing in favor of a dedicated getBooleanInput function https://github.com/actions/toolkit/issues/723

PR #367, opened 2020-03-03
Updates to @actions/artifact package

konradpabjan:
## Overview

Various fixes and improvements to the `@actions/artifact` package 

- Fixes issues related to too many TCP connections being made when uploading large amounts of files. This would cause weird timeouts and errors when large amounts of files would be uploaded (usually every 1000-1500 there would be some timeout or error). Previously a single httpclient was used, now there is a separate http manager hat handles all http clients. If an error is encountered, the client is disposed of and a new connection is reset.
- GZip compression is now used to compress files before upload, just like in V1. If a file is over 64k in size, a temporary file is used for compression that is then deleted. If a file is less than 64k in size, an in memory buffer is used for compression. Without this compression, uploads on self-hosted runners would sometimes be 10x slower.
- Enforce legal characters in file paths of the artifacts uploaded. An extra check is added to check when getting the upload specification
- Refactoring to make `upload-http-client` and `download-http-client` classes instead of just a collection of functions. This makes it easier to maintain state and it allows multiple artifacts to be uploaded/downloaded at the same time (there is a lot of whitespace/spacing changes which is why everything is showing up as modified)
- Various improvements to logging so that it is easier to see the status of an upload. Total upload status gets displayed every 10 seconds while large files over 100MB get their status updated every 1 second if there is a change in the upload status
image

## Testing and Performance

Used alongside a fork of `upload-artifact` and `npm link` for local development and testing

#### Self Hosted

1GB artifact on `self-hosted`: https://github.com/konradpabjan/artifact-test/runs/482016447?check_suite_focus=true
**v1:** 4min 43 seconds
**v2:** 6min 12 seconds

1GB artifact on `self-hosted`: https://github.com/konradpabjan/artifact-test/runs/482076877?check_suite_focus=true
**v1:** 5min 17 seconds
**v2:** 6min 27 seconds

1GB file + 5000 small files: https://github.com/konradpabjan/artifact-test/runs/485214944?check_suite_focus=true
**v1:** 12min 5sec
**v2:** 17min 46sec

1GB file + 5000 small files again: https://github.com/konradpabjan/artifact-test/runs/485295847?check_suite_focus=true
**v1:** 15min 57sec
**v2:** 20min 39sec

#### Ubuntu Latest

1GB artifact: https://github.com/konradpabjan/artifact-test/runs/481974558?check_suite_focus=true
**v1:** 5min 30 seconds
**v2:** 5min 43 seconds

1GB artifact:  https://github.com/konradpabjan/artifact-test/runs/482117480?check_suite_focus=true
**v1:** 5min 53 seconds
**v2:** 6min 6 seconds

5000 small files that are only 34k total in size: https://github.com/konradpabjan/artifact-test/runs/485122402?check_suite_focus=true
**v1:** 4min 21 seconds
**v2:** 5min 1 second

5000 small files again: https://github.com/konradpabjan/artifact-test/runs/485146951?check_suite_focus=true
**v1:** 3min 41 seconds
**v2:** 4min 24 seconds

5000 small files one more time: https://github.com/konradpabjan/artifact-test/runs/485180362?check_suite_focus=true
**v1:** 3min 44 seconds
**v2:** 4min 11 seconds

The performance can be tweaked in the future to hopefully make it even faster. Theoretically, it should be identical to `v1`. There is a large difference between multi-threading in C# vs Node though and at least some of the slower performance can be attributed to porting over between languages.


packages/artifact/src/internal-utils.ts
```diff
@@ -59,17 +60,40 @@ export function getContentRange(
   return `bytes ${start}-${end}/${total}`
 }
 
+/**
+ * Sets all the necessary headers when making HTTP calls
+ * @param {string} contentType the type of content being uploaded
+ * @param {boolean} isKeepAlive is the same connection being used to make multiple calls
+ * @param {boolean} isGzip is the connection being used to upload GZip compressed content
+ * @param {number} uncompressedLength the original size of the content if something is being uploaded that has been compressed
+ * @param {number} contentLength the lenght of the content that is being uploaded
```
joshmgross:
```suggestion
 * @param {number} contentLength the length of the content that is being uploaded
```

packages/artifact/src/internal-utils.ts
```diff
@@ -97,19 +121,42 @@ export function getArtifactUrl(): string {
  * file systems such as NTFS. To maintain platform-agnostic behavior, all characters that are not supported by an
  * individual filesystem/platform will not be supported on all fileSystems/platforms
  */
-const invalidCharacters = ['\\', '/', '"', ':', '<', '>', '|', '*', '?', ' ']
+const invalidArtifactFileCharacters = ['"', ':', '<', '>', '|', '*', '?', ' ']
+const invalidArtifactNameCharacters = ['\\', '/']
```
joshmgross: Is this correct? I would've thought any invalid file character would also be an invalid name character (since the name is the root directory of the artifact)

packages/artifact/src/internal-http-manager.ts
```diff
@@ -0,0 +1,36 @@
+import {HttpClient} from '@actions/http-client/index'
+import {createHttpClient} from './internal-utils'
+
+/**
+ * Used for managing all http Connections during either upload and download in order to limit the number of tcp connections created
+ * Separate clients for download and upload are used so that there are no conflicts if both are used at the same time
+ */
+export abstract class HttpManager {
```
joshmgross: Neat! Is this pattern used in some of our other actions?

konradpabjan: I haven't seen this pattern used in any of our other actions, came up with this myself.

Using design patterns for the sake of using design patterns is never a good thing, but I thought this was as good use case for the singleton to ensure there is only ever class managing http connections

packages/artifact/src/internal-utils.ts
konradpabjan: I'll reword some comments and variable names for clarity, the name and path are technically restricted to the set of of characters. The original restrictions comes from `v1` [here]( https://github.com/actions/runner/blob/0cba42590f47d73d811a8bb25823e78a42bd766f/src/Runner.Plugins/Artifact/PublishArtifact.cs#L43)

Previously we would only check the artifact name. The path however that is used for upload was never checked and it could cause problems if illegal characters are included, this would result in the following issue: https://github.com/github/c2c-actions-service/issues/469

Because the upload path can include `/` or `\`, (while the name shouldn't), I split up the character sets

packages/artifact/src/internal-utils.ts
```diff
@@ -96,20 +119,61 @@ export function getArtifactUrl(): string {
...
+  ':',
+  '<',
+  '>',
+  '|',
+  '*',
+  '?',
+  ' ',
+  '\\',
+  '/'
+]
```
joshmgross: I think this would be more clear if `invalidArtifactFilePathCharacters` was explicitly a subset of `invalidArtifactNameCharacters`. Something like:
```suggestion
const invalidArtifactNameCharacters = [...invalidArtifactFilePathCharacters, '\\', '/']
```

konradpabjan:
The performance discrepancy between `v1` and `v2` for self-hosted is driving me crazy 😓 but I still can't get a good answer, it's not as bad as it was without GZip, but theoretically is should be fairly close to `v1` performance.

Here are some interesting queries that show the amount of http calls, bytes sent over, response time  etc between the two [Kusto](https://dataexplorer.azure.com/clusters/vso/databases/VSO?query=H4sIAAAAAAAAA8WUTW/aQBCG7/yKUS4YqbDY2BBaoYr2kjY0RRSSphdrWQa8xd6l+1GLqD++s4GUlt5yaA+2xvvOzuy8z8qMwdg4uebCgbdSbeA2Bq2Ag8Vy3S60dbgC45VC02AMCud29iVjG+kKv+wIXbGtVoavdnz5lSvGj8XaDq1jtM+y9DIb9BJ6XosCxTa3XjrM11p4O3LGYygbX0IlFWQZtRVarWyjRAfWUbm5rBBGwDc66q343rZePWqoVkdF6TqixXfSTvTGNn5AXaBBEDav0BV6BaMRXEwX84vfJW9kTqNVQWyy2dXnjzN3v5wW8+UXOy73af3+zXV591D37yY319V4Ivc3368+LDCb3nf5oj9jOd9Jywxa7Y1Ay95q5bgklyxL4jTu9/rN837fPJo9iEOiheaT8bZ9m5ySpwaFtBiG++R4tYMluhpRQfTLjk7nOH2LdllfVdzIBwxRRK2We/K+9YIaeeUiCsK6o+zc8S2q1qPf3WF8OUzSwRCyJOvBsDvopUlQDhHRKEt5ZEEex1knDovKU+1GyDu/Ncnp1sAft4b6c+JEH2HbXDteQjgNnSEgT5J/jvxE4P+Dj58LnrHno0+JfDZMAvohDLrpoBsH5RD9hT7uDJ7IQ1QXUhQgLVSajuwKruh1jhJcoEzdNYlhNMqnYRyaYILwxqASe6jpH0K/m9ZPq2B59IQEAAA=). The individual time for most `v1` and `v2` calls is 70-120 milliseconds, however there are occasional blimps for `v2` that take 5-6 seconds and this adds up over time and largely contributes to time difference recorded. For all small file uploads, `v1` doesn't have anything below a 1 second response time while `v2` does and the times are all over the place

I also added more telemetry for a `v2` self-hosted runner and the bottleneck keep coming back to making the `PUT` call. I ended up doing this and the call time is always over 200+ milliseconds
```
        const httpStartTime = performance.now()
        const response = await uploadChunkRequest()
        debug(`HTTP time ${(performance.now() - httpStartTime).toFixed(3)} for ${resourceUrl}`)
```
You can see the output [here](https://github.com/konradpabjan/artifact-test/runs/487344030?check_suite_focus=true) so the GZip and the buffering is definitely not the bottleneck

packages/artifact/src/__mocks__/internal-config-variables.ts
```diff
@@ -12,6 +12,14 @@ export function getUploadChunkConcurrency(): number {
 export function getUploadChunkSize(): number {
   return 4 * 1024 * 1024 // 4 MB Chunks
 }
+
+export function getUploadRetryCount(): number {
+  return 1
+}
+
+export function getRetryWaitTime(): number {
+  return 1
```
yacaovsnc: Would be clearer if we put the unit (milliseconds, second, or minute) in the function, `getRetryWaitTimeInSeconds()` or something like that.

packages/artifact/src/internal-artifact-client.ts
```diff
@@ -110,23 +106,27 @@ export class DefaultArtifactClient implements ArtifactClient {
...
         uploadSpecification,
         options
       )
 
-      //Update the size of the artifact to indicate we are done uploading
-      await patchArtifactSize(uploadResult.size, name)
+      // Update the size of the artifact to indicate we are done uploading
+      // The uncompressed size is used for display when downloading a zip of the artifact from the UI
+      await uploadHttpClient.patchArtifactSize(
+        uploadResult.uncompressedSize,
```
yacaovsnc: This has been a hotly debated topic - if we upload compressed and charge for _uncompressed_, it's odd.  To reduce questions, I would call this `totalSize` instead of calling out the compressiveness.

packages/artifact/src/internal-http-manager.ts
```diff
@@ -0,0 +1,36 @@
+import {HttpClient} from '@actions/http-client/index'
+import {createHttpClient} from './internal-utils'
+
+/**
+ * Used for managing all http Connections during either upload or download in order to limit the number of tcp connections created
```
joshmgross: Out of curiosity, why do we need multiple http connections?

Does Node not handle multiple requests on a single client well?

konradpabjan: Multiple connections are setup so that it's easier to reset/dispose of one if something goes wrong.

We use the `keep-alive` header to not close the connection and to increase responsiveness. If something goes wrong, the connection should be disposed of. See https://github.com/microsoft/typed-rest-client/blob/5ce8cd7ae336384bd2d1851f5e845d89885c3d21/lib/HttpClient.ts#L313 (same thing applies to the rest client we use)

Disposing of one client should not interfere with any other concurrent uploads going on with another client so that's why multiple connections are used.

Using a single connection for every concurrent upload was also a recommendation that I got from @ericsciple when I was fighting with too many connections being open and weird timeouts

packages/artifact/src/internal-download-http-client.ts
```diff
@@ -12,122 +12,174 @@ import {
   QueryArtifactResponse
 } from './internal-contracts'
 import {IHttpClientResponse} from '@actions/http-client/interfaces'
-import {HttpClient} from '@actions/http-client'
+import {HttpManager} from './internal-http-manager'
```
yacaovsnc: It seems everything is `internal` 😄 , and naming every file with prefix `internal` adds clutter.  We could either create a folder `internal` and move all files there, or name the only `artifact-client` file `external-artifact-client`.   Then we can drop the `internal` prefix from everywhere.

packages/artifact/src/internal-download-http-client.ts
```diff
@@ -12,122 +12,174 @@ import {
...
-): Promise<QueryArtifactResponse> {
-  // The itemPath search parameter controls which containers will be returned
-  const resourceUrl = new URL(containerUrl)
-  resourceUrl.searchParams.append('itemPath', artifactName)
+  /**
+   * Gets a list of all artifacts that are in a specific container
+   */
+  async listArtifacts(): Promise<ListArtifactsResponse> {
+    const artifactUrl = getArtifactUrl()
+    this.downloadHttpManager.createClients(1)
```
yacaovsnc: This is unnecessary - what you are doing here is just create a single http client.  Involving the `downloadHttpManager` here complicates the logic and hide your intention.  Also this `downloadHttpManager` is a singleton in this class, so if the consumer calls `listArtifacts()` twice on async threads, your http clients may be corrupted - thread b is creating the http client, and thread a is getting the client in line 36.

packages/artifact/src/internal-download-http-client.ts
```diff
@@ -12,122 +12,174 @@ import {
...
+   * @param containerUrl the artifact container URL for the run
+   */
+  async getContainerItems(
+    artifactName: string,
+    containerUrl: string
+  ): Promise<QueryArtifactResponse> {
+    // the itemPath search parameter controls which containers will be returned
+    const resourceUrl = new URL(containerUrl)
+    resourceUrl.searchParams.append('itemPath', artifactName)
+    this.downloadHttpManager.createClients(1)
```
yacaovsnc: Same here.  If the intention is just to have a single http client, just initiate a single http client.

packages/artifact/src/internal-download-http-client.ts
```diff
@@ -12,122 +12,174 @@ import {
...
-      `Received http ${response.message.statusCode} during file download, will retry ${artifactLocation} after 10 seconds`
+  /**
+   * Concurrently downloads all the files that are part of an artifact
+   * @param downloadItems information about what items to download and where to save them
+   */
+  async downloadSingleArtifact(downloadItems: DownloadItem[]): Promise<void> {
+    const DOWNLOAD_CONCURRENCY = getDownloadFileConcurrency()
+    // limit the number of files downloaded at a single time
+    const parallelDownloads = [...new Array(DOWNLOAD_CONCURRENCY).keys()]
+    this.downloadHttpManager.createClients(DOWNLOAD_CONCURRENCY)
```
yacaovsnc: You may run into concurrency issues since the `downloadHttpManager` is a single instance and isn't thread-safe.

packages/artifact/src/internal-download-http-client.ts
```diff
@@ -12,122 +12,174 @@ import {
...
+   * @param artifactLocation origin location where a file will be downloaded from
+   * @param downloadPath destination location for the file being downloaded
+   */
+  private async downloadIndividualFile(
+    httpClientIndex: number,
+    artifactLocation: string,
+    downloadPath: string
+  ): Promise<void> {
+    const stream = fs.createWriteStream(downloadPath)
+    const client = this.downloadHttpManager.getClient(httpClientIndex)
```
yacaovsnc: This client may be lost if the consumer called `listArtifacts()` or `getContainerItems()` again.

packages/artifact/src/internal-download-http-client.ts
yacaovsnc: If you really want to create a pool of connections, we should:
1. Let the `HttpManager` handle all client creation and disposal.  So as soon as a `httpManager` is initiated, it creates let's say 8 http clients and keep a reference to them.  No other part of the system should be able to reinitiate this maanger.  If they want more http client, they have to create a new `httpManager`.
2. This manager should not expose the client out otherwise other parts of the system could accidentally close and dispose client without notifying the manager. 
3. This manager should handle allocating http client to make requests - i.e. all the `post`, `put`, `get` method should be on the manager, not on the actual http client.    

As you can see, this is a pretty major change.  So I would just suggest you create an array of http clients yourself in this single method, and pass the http client out to `downloadIndividualFile` method.

packages/artifact/src/internal-download-http-client.ts
konradpabjan: The user won't be able to call `listArtifacts` twice because it's an internal method. What we publicly expose is `uploadArtifact()`, `downloadAritfact()` and `downloadAllArtifacts()`. These are the methods that are exposed as part of the public interface for an artifactClient after `create()` gets called. They could poke around and try to call these methods themselves, but that's a large reason as to why `internal` is used in the naming everywhere so it's obvious what is public and what isn't.

Whenever `downloadArtifact` is called, a new downloadhttpClient class is created, right [here](https://github.com/actions/toolkit/blob/640fc522a7d780531cd610c1d473829a4d237e8a/packages/artifact/src/internal-artifact-client.ts#L140) so there should not be any problems with corruption. When a new class gets created, a new manager gets created as part of the constructor [here](https://github.com/actions/toolkit/blob/640fc522a7d780531cd610c1d473829a4d237e8a/packages/artifact/src/internal-download-http-client.ts#L27).

I did some refactoring to make this a class so it's easier to maintain state and not run into any problems if `downloadArtifact()` is called multiple times in quick succession. Each one of these classes has their own `downloadHttpManager` (or `uploadHttpManager` if uploading). I experimented with a singleton earlier on but decided against it precisely because there could be problems if `uploadArtifact` or `download` artifact get called multiple times in quick succession

packages/artifact/src/internal-download-http-client.ts
konradpabjan: See my above comment, `listArtifacts()`, much like `getContainerItems()` is an internal method and users won't be able to call it unless they do 
```
const downloadhttpclient = new DownloadHttpClient() // creates a new download http manager with its own group of clients
downloadhttpclient .listArtifacts()
```

packages/artifact/src/internal-download-http-client.ts
konradpabjan: The original motivation behind this is the `glob` package: https://github.com/actions/toolkit/tree/master/packages/glob/src

I've been following the structure pretty closely and `internal` is used to clearly differentiate between what is public and what isn't. There are a lot of internal methods like `listArtifact` or `patchSize` that I don't want people poking around with and this helps set a clear boundary.

Given the large amount of files though, I'll probably move them into a separate directory.

packages/artifact/src/internal-http-manager.ts
joshmgross: >Using a single connection for every concurrent upload was also a recommendation

Would it be worth going down that route? It would simplify a lot of this code

packages/artifact/src/internal-download-http-client.ts
yacaovsnc: Okay, as long as we don't call dispose on those clients then it's good.

packages/artifact/src/internal-http-manager.ts
konradpabjan: That's what I have in my implementation. There is effectively an array of http clients and each concurrent upload gets it's own. 

Initialized here: https://github.com/actions/toolkit/blob/2bc6c35d66e24f5fce0af8201e13284ce6a8e206/packages/artifact/src/internal/upload-http-client.ts#L119

joshmgross: Why not share a single http client among all concurrent uploads?

packages/artifact/src/internal/upload-gzip.ts
```diff
@@ -0,0 +1,51 @@
...
+  tempFilePath: string
+): Promise<number> {
+  return new Promise((resolve, reject) => {
+    const inputStream = fs.createReadStream(originalFilePath)
+    const gzip = zlib.createGzip()
+    const outputStream = fs.createWriteStream(tempFilePath)
+    inputStream.pipe(gzip).pipe(outputStream)
+    outputStream.on('finish', () => {
+      // wait for stream to finish before calculating the size which is needed as part of the Content-Length header when starting an upload
+      const size = fs.statSync(tempFilePath).size
```
yacaovsnc: I am guessing those `sync` functions are slowing down your throughput.  Can you `await` on the async version?

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -0,0 +1,189 @@
...
+      )
+      // if an error is encountered, dispose of the http connection, wait, and create a new one
+      this.downloadHttpManager.disposeClient(httpClientIndex)
+      await new Promise(resolve =>
+        setTimeout(resolve, getRetryWaitTimeInMilliseconds())
+      )
+      this.downloadHttpManager.replaceClient(httpClientIndex)
+      const retryResponse = await client.get(artifactLocation)
+      if (isSuccessStatusCode(retryResponse.message.statusCode)) {
+        if (
```
yacaovsnc: This logic is duplicated with line 119 above.  You could have created a private method to reduce duplication. Try apply DRY principle as much as possible.  :) 
```
var isGzip = (response) => {
...
}
```

and call `pipeResponseToStream(.., .., isGzip(response))`

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -0,0 +1,189 @@
...
+  constructor() {
+    this.downloadHttpManager = new HttpManager()
+  }
+
+  /**
+   * Gets a list of all artifacts that are in a specific container
+   */
+  async listArtifacts(): Promise<ListArtifactsResponse> {
+    const artifactUrl = getArtifactUrl()
+    this.downloadHttpManager.createClients(1)
```
yacaovsnc: I still don't think this is necessary - if we know how many clients we need (from the `DOWNLOAD_CONCURRENCY`), then we should just create those clients in the constructor.  Just initiate the httpmanager in constructor with `createClients(DOWNLOAD_CONCURRENCY)`.

Then we will not recreate http clients in each method - for `listArtifacts()` just take one from the manager and user it.

packages/artifact/src/internal/upload-gzip.ts
konradpabjan: Changed it up using `promisify` and the async `fs.stat`

packages/artifact/src/internal/download-http-client.ts
konradpabjan: I overhauled the `HttpManager` after our discussions yesterday. All instantiation of clients now happens inside the manager and the number of clients is equal to the upload or download concurrency.

If a call needs to be made without `keep-alive`, the first client in the manager is used

packages/artifact/src/internal/download-http-client.ts
konradpabjan: Nice private method 😀

konradpabjan:
I'll add contribution guidelines and some end-to-end tests that run for each PR after this goes in. Want to get `v2` out in the `upload-artifact` and `download-artifact` repo and this PR is already massive.

packages/artifact/src/internal-download-http-client.ts
konradpabjan: I've updated the directory structure, previously `internal` files now are in a separate directory

packages/artifact/docs/implementation-details.md
```diff
@@ -0,0 +1,43 @@
+# Implementation Details
```
yacaovsnc: Don't have to be in this PR, but I think we should call out explicitly that this is not a contract we are going to maintain - it's great to be transparent about the internal choices we make, but at the same time we are free to make any changes here.  Don't code any solution based on information in this doc.  For example, if the customer scenario warrants compressing the artifacts, please do it themselves, we may not compress for them in the future.

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -0,0 +1,195 @@
...
+  async getContainerItems(
+    artifactName: string,
+    containerUrl: string
+  ): Promise<QueryArtifactResponse> {
+    // the itemPath search parameter controls which containers will be returned
+    const resourceUrl = new URL(containerUrl)
+    resourceUrl.searchParams.append('itemPath', artifactName)
+
+    // no concurrent calls so a single httpClient without the http-manager is sufficient
+    const client = createHttpClient()
```
yacaovsnc: Why not do `const client = this.downloadHttpManager.getClient(0)`?  We don't need to recreate the http client.

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    // check the response headers to determine if the file was compressed using gzip
+    const isGzip = (headers: IncomingHttpHeaders): boolean => {
+      if (
+        headers['content-encoding'] &&
+        headers['content-encoding'] === 'gzip'
+      ) {
+        return true
+      }
+      return false
```
yacaovsnc: This is `return headers['content-encoding'] &&  headers['content-encoding'] === 'gzip'`

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -0,0 +1,195 @@
...
+  private async downloadIndividualFile(
+    httpClientIndex: number,
+    artifactLocation: string,
+    downloadPath: string
+  ): Promise<void> {
+    const stream = fs.createWriteStream(downloadPath)
+    const client = this.downloadHttpManager.getClient(httpClientIndex)
+    const requestOptions = getRequestOptions('application/octet-stream', true)
+    const response = await client.get(artifactLocation, requestOptions)
+    await response.readBody()
```
yacaovsnc: Does this load the entire file into memory?  Do we need this?

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -0,0 +1,195 @@
...
+      warning(
+        `Received http ${response.message.statusCode} during file download, will retry ${artifactLocation} after 10 seconds`
+      )
+      // if an error is encountered, dispose of the http connection, and create a new one
+      this.downloadHttpManager.disposeAndReplaceClient(httpClientIndex)
+      await new Promise(resolve =>
+        setTimeout(resolve, getRetryWaitTimeInMilliseconds())
+      )
+      const retryResponse = await client.get(artifactLocation)
+      await retryResponse.readBody()
```
yacaovsnc: Same here.  Do we need this?

packages/artifact/src/internal/upload-gzip.ts
yacaovsnc: This is still sync, did you upload the latest code yet?

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -0,0 +1,195 @@
...
+            index,
+            currentFileToDownload.sourceLocation,
+            currentFileToDownload.targetPath
+          )
+        }
+      })
+    )
+
+    // done downloading, safety dispose all connections
+    this.downloadHttpManager.disposeAndReplaceAllClients()
```
yacaovsnc: You probably should separate those two functions - here there is no need to replace all clients.

packages/artifact/src/internal/download-http-client.ts
konradpabjan: Oops, missed this one. Got `listArtifacts` but accidentally skipped over this method. Nice catch, should be the same thing

packages/artifact/src/internal/download-http-client.ts
konradpabjan: Needs slight modification, but this works: `return 'content-encoding' in headers && headers['content-encoding'] === 'gzip'`

packages/artifact/src/internal/download-http-client.ts
konradpabjan: I added the `response.readBody` as a precaution to make sure the connection gets closed so that there is no resource leak. Similar to what was causing the weird 1 minute wait with upload after everything was one.

I tested with and without `response.readBody()` and it both cases it's working fine for download so I'll remove it. In the initial version of the package I don't have this snippet and it works fine so it looks like only upload needs to read the body.

packages/artifact/docs/implementation-details.md
konradpabjan: I'll add information about this in my next PR when I'm adding contribution guidelines and tests

packages/artifact/src/internal/download-http-client.ts
konradpabjan: Disposal and replacement is necessary when calling `downloadAllArtifacts` because the same `downloadHttpClient` gets used sequentially and we need to guarantee that the clients are properly replaced and disposed.

When uploading an artifact or downloading an individual one, the extra replacement is a little wasteful because some of the clients will never be used again, but it's safer to properly dispose and replace each client so that it's guaranteed that all clients can be used if need be

PR #368, opened 2020-03-04
save-state doc clarification

bryanmacfarlane:
Make it clear that an environment variable will be available for use in future steps.
EDIT: post of current action


bryanmacfarlane:
crap - it's only available in post of action

thboop:
Resolved by https://github.com/actions/toolkit/pull/467

PR #369, opened 2020-03-04
handle response body error

ericsciple:
Update tool-cache downloadTool() to handle errors from response body stream

related to https://github.com/actions/setup-node/issues/116

### DIFF
--- a/packages/tool-cache/package.json
+++ b/packages/tool-cache/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@actions/tool-cache",
-  "version": "1.3.1",
+  "version": "1.3.2",
   "description": "Actions tool-cache lib",
   "keywords": [
     "github",
--- a/packages/tool-cache/src/retry-helper.ts
+++ b/packages/tool-cache/src/retry-helper.ts
@@ -1,0 +1,55 @@
+import * as core from '@actions/core'
+
+/**
+ * Internal class for retries
+ */
+export class RetryHelper {
+  private maxAttempts: number
+  private minSeconds: number
+  private maxSeconds: number
+
+  constructor(maxAttempts: number, minSeconds: number, maxSeconds: number) {
+    if (maxAttempts < 1) {
+      throw new Error('max attempts should be greater than or equal to 1')
+    }
+
+    this.maxAttempts = maxAttempts
+    this.minSeconds = Math.floor(minSeconds)
+    this.maxSeconds = Math.floor(maxSeconds)
+    if (this.minSeconds > this.maxSeconds) {
+      throw new Error('min seconds should be less than or equal to max seconds')
+    }
+  }
+
+  async execute<T>(action: () => Promise<T>): Promise<T> {
+    let attempt = 1
+    while (attempt < this.maxAttempts) {
+      // Try
+      try {
+        return await action()
+      } catch (err) {
+        core.info(err.message)
+      }
+
+      // Sleep
+      const seconds = this.getSleepAmount()
+      core.info(`Waiting ${seconds} seconds before trying again`)
+      await this.sleep(seconds)
+      attempt++
+    }
+
+    // Last attempt
+    return await action()
+  }
+
+  private getSleepAmount(): number {
+    return (
+      Math.floor(Math.random() * (this.maxSeconds - this.minSeconds + 1)) +
+      this.minSeconds
+    )
+  }
+
+  private async sleep(seconds: number): Promise<void> {
+    return new Promise(resolve => setTimeout(resolve, seconds * 1000))
+  }
+}
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -5,10 +5,13 @@ import * as os from 'os'
 import * as path from 'path'
 import * as httpm from '@actions/http-client'
 import * as semver from 'semver'
+import * as stream from 'stream'
+import * as util from 'util'
 import uuidV4 from 'uuid/v4'
 import {exec} from '@actions/exec/lib/exec'
 import {ExecOptions} from '@actions/exec/lib/interfaces'
 import {ok} from 'assert'
+import {RetryHelper} from './retry-helper'
 
 export class HTTPError extends Error {
   constructor(readonly httpStatusCode: number | undefined) {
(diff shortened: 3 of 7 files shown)


packages/tool-cache/src/tool-cache.ts
```diff
@@ -83,23 +83,19 @@ export async function downloadTool(
 
       const file: NodeJS.WritableStream = fs.createWriteStream(dest)
       file.on('open', async () => {
-        try {
```
ericsciple: this try catch looked unnecessary

packages/tool-cache/src/tool-cache.ts
```diff
@@ -83,23 +83,19 @@ export async function downloadTool(
...
-          const stream = response.message.pipe(file)
-          stream.on('close', () => {
```
ericsciple: `stream` is the writable stream. To improve readability - i had trouble figuring out which stream this is until after reading nodejs docs a bit - i moved the closed event handler further below (`file.on('close', ...)`)

ericsciple: Which also makes it consistent with the error handler for the writable stream - registered on local variable `file` not `stream`

packages/tool-cache/src/tool-cache.ts
```diff
@@ -83,23 +83,19 @@ export async function downloadTool(
...
-            core.debug('download complete')
-            resolve(dest)
-          })
-        } catch (err) {
-          core.debug(
-            `Failed to download from "${url}". Code(${response.message.statusCode}) Message(${response.message.statusMessage})`
-          )
+        response.message.pipe(file)
+        response.message.on('error', (err: Error) => {
```
ericsciple: This error handler for the response body stream is new.

I was able to confirm it was missing by forcing an error: i hooked up to the on data, and then called `response.message.emit('error', ...)`. And an unhandled error bubbled.

Looking into a way to unit test.

packages/tool-cache/src/tool-cache.ts
```diff
@@ -83,23 +83,21 @@ export async function downloadTool(
...
-            core.debug('download complete')
-            resolve(dest)
-          })
-        } catch (err) {
-          core.debug(
-            `Failed to download from "${url}". Code(${response.message.statusCode}) Message(${response.message.statusMessage})`
-          )
+        const readStream = response.message
+        readStream.pipe(file)
+        readStream.on('error', (err: Error) => {
```
ericsciple: This error handler for the response body stream is new.

I was able to confirm it was missing by forcing an error: i hooked up to the on data, and then called response.message.emit('error', ...). And an unhandled error bubbled.

packages/tool-cache/src/tool-cache.ts
```diff
@@ -83,23 +84,26 @@ export async function downloadTool(
...
-          )
+        /* eslint-disable @typescript-eslint/no-explicit-any */
+        const mockResponseMessage = (global as any)[
+          'UNIT_TEST_TOOL_CACHE_DOWNLOAD_TOOL_RESPONSE_MESSAGE'
+        ] as stream.Readable | undefined
+        /* eslint-enable @typescript-eslint/no-explicit-any */
+        const readStream = mockResponseMessage || response.message
+        readStream.pipe(file)
+        readStream.on('error', (err: Error) => {
+          file.end()
```
ericsciple: should delete file in this case

packages/tool-cache/src/tool-cache.ts
```diff
@@ -83,23 +84,26 @@ export async function downloadTool(
 
       const file: NodeJS.WritableStream = fs.createWriteStream(dest)
```
ericsciple: todo: add retry

packages/tool-cache/src/tool-cache.ts
```diff
@@ -81,24 +97,43 @@ export async function downloadTool(
...
-          reject(err)
-        }
+        const responseMessageFactory = getGlobal<() => stream.Readable>(
+          'TEST_DOWNLOAD_TOOL_RESPONSE_MESSAGE_FACTORY',
+          () => response.message
+        )
+        const readStream = responseMessageFactory()
+        readStream.pipe(file)
+        readStream.on('error', (err: Error) => {
+          file.end()
```
ericsciple: per docs, when piping error on the read stream does not close the write stream. needs to be called manually

packages/tool-cache/src/tool-cache.ts
```diff
@@ -81,24 +97,43 @@ export async function downloadTool(
...
+      })
+      file.on('close', () => {
+        // Success
+        if (!error) {
+          core.debug('download complete')
+          resolve(dest)
+          return
+        }
+
+        // Error, delete dest before retry
```
ericsciple: since retry the response stream now, need to delete dest before retrying.

packages/tool-cache/src/retry-helper.ts
```diff
@@ -0,0 +1,55 @@
+import * as core from '@actions/core'
+
+/**
+ * Internal class for retries
```
ericsciple: Private for now. If expose a class for consumers we probably want to put more thought into the interface and behaviors. And probably do exponent backoff in that case.

packages/tool-cache/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/tool-cache",
-  "version": "1.3.1",
+  "version": "1.3.2",
```
ericsciple: todo: update release notes

packages/tool-cache/src/tool-cache.ts
```diff
@@ -81,24 +97,43 @@ export async function downloadTool(
...
-            `Failed to download from "${url}". Code(${response.message.statusCode}) Message(${response.message.statusMessage})`
-          )
-          reject(err)
-        }
+        const responseMessageFactory = getGlobal<() => stream.Readable>(
+          'TEST_DOWNLOAD_TOOL_RESPONSE_MESSAGE_FACTORY',
+          () => response.message
+        )
+        const readStream = responseMessageFactory()
+        readStream.pipe(file)
```
SimenB: Thoughts on using [`pipeline`](https://nodejs.org/api/stream.html#stream_stream_pipeline_source_transforms_destination_callback)? It can be promisified, and deals with errors from all streams

ericsciple: will look into switching over to that

ericsciple: that is waaaaay better

thanks!

SimenB: great! 👍

packages/tool-cache/src/retry-helper.ts
```diff
@@ -0,0 +1,55 @@
...
+    this.minSeconds = Math.floor(minSeconds)
+    this.maxSeconds = Math.floor(maxSeconds)
+    if (this.minSeconds > this.maxSeconds) {
+      throw new Error('min seconds should be less than or equal to max seconds')
+    }
+  }
+
+  async execute<T>(action: () => Promise<T>): Promise<T> {
+    let attempt = 1
+    while (attempt < this.maxAttempts) {
```
TingluoHuang: what if `maxAttempts` is 1?

packages/tool-cache/src/tool-cache.ts
```diff
@@ -55,55 +58,68 @@ export async function downloadTool(
...
+    await pipeline(readStream, fs.createWriteStream(dest))
+    core.debug('download complete')
+    succeeded = true
+    return dest
+  } finally {
+    // Error, delete dest before retry
+    if (!succeeded) {
+      core.debug('download failed')
+      try {
+        await io.rmRF(dest)
```
TingluoHuang: if we can't delete the `dest` on error, then the next retry will just error with `Destination file path ${dest} already exists`.

ericsciple: it shouldnt happen. If so then a bug we need to figure out.

packages/tool-cache/src/tool-cache.ts
```diff
@@ -55,55 +58,68 @@ export async function downloadTool(
...
+  // Download the response body
+  const pipeline = util.promisify(stream.pipeline)
+  const responseMessageFactory = getGlobal<() => stream.Readable>(
+    'TEST_DOWNLOAD_TOOL_RESPONSE_MESSAGE_FACTORY',
+    () => response.message
+  )
+  const readStream = responseMessageFactory()
+  let succeeded = false
+  try {
+    await pipeline(readStream, fs.createWriteStream(dest))
```
ericsciple: > A module method to pipe between streams and generators forwarding errors and properly cleaning up and provide a callback when the pipeline is complete.

PR #373, opened 2020-03-09
only retry downloadtool on 500s and 408 and 429

ericsciple:
downloadtool should only retry on 500s and 408 and 429

### DIFF
--- a/packages/tool-cache/__tests__/retry-helper.test.ts
+++ b/packages/tool-cache/__tests__/retry-helper.test.ts
@@ -66,7 +66,7 @@ describe('retry-helper tests', () => {
     expect(info[3]).toMatch(/Waiting .+ seconds before trying again/)
   })
 
-  it('all attempts fail succeeds', async () => {
+  it('all attempts fail', async () => {
     let attempts = 0
     let error: Error = (null as unknown) as Error
     try {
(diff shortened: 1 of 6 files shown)


packages/tool-cache/src/retry-helper.ts
```diff
@@ -21,13 +21,20 @@ export class RetryHelper {
...
+    isRetryable?: (e: Error) => boolean
+  ): Promise<T> {
     let attempt = 1
     while (attempt < this.maxAttempts) {
       // Try
       try {
         return await action()
       } catch (err) {
+        if (isRetryable && !isRetryable(err)) {
+          throw err
```
ericsciple: nodejs retains the original stack trace

packages/tool-cache/src/tool-cache.ts
```diff
@@ -23,30 +24,6 @@ export class HTTPError extends Error {
 const IS_WINDOWS = process.platform === 'win32'
 const userAgent = 'actions/tool-cache'
 
-// On load grab temp directory and cache directory and remove them from env (currently don't want to expose this)
```
ericsciple: 1. i moved this to a function otherwise can't override the location from unit tests (because loads on module import)
2. we shouldn't fallback to /Users/actions/temp - i got permissions issue when attempting to run setup-node tests locally. I added an assert that RUNNER_TEMP and RUNNER_TOOL_CACHE are defined.

ericsciple: I copied the build output to setup-node to verify the unit tests are runnable with these changes

packages/tool-cache/src/tool-cache.ts
```diff
@@ -533,10 +528,28 @@ function _evaluateVersions(versions: string[], versionSpec: string): string {
   return version
 }
 
+/**
+ * Gets RUNNER_TOOL_CACHE
+ */
+function _getCacheDirectory(): string {
+  const cacheDirectory = process.env['RUNNER_TOOL_CACHE'] || ''
```
thboop: Thoughts on defaulting to `.`?
`Path.join` ignores empty strings, it should be the same result, but using `.` means these functions do return valid paths.
Same for below.

packages/tool-cache/__tests__/retry-helper.test.ts
```diff
@@ -84,4 +84,42 @@ describe('retry-helper tests', () => {
...
+      )
+    } catch (err) {
+      error = err
+    }
+    expect(error.message).toBe('some error 1')
+    expect(attempts).toBe(1)
+    expect(info).toHaveLength(0)
+  })
+
+  it('checks retryable after second attempt', async () => {
```
thboop: 👍

packages/tool-cache/src/tool-cache.ts
ericsciple: i added `|| ''` to satisfy typescript. The `ok` check verifies not empty.

packages/tool-cache/src/tool-cache.ts
```diff
@@ -58,23 +34,37 @@ export async function downloadTool(
...
   return await retryHelper.execute(
-    async () => await downloadToolAttempt(url, dest || '')
+    async () => {
+      return await downloadToolAttempt(url, dest || '')
+    },
+    (err: Error) => {
+      if (err instanceof HTTPError && err.httpStatusCode) {
+        // Don't retry anything less than 400
+        // Don't retry 404 Not Found
+        if (err.httpStatusCode < 400 || err.httpStatusCode === 404) {
```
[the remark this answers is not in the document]
ericsciple: fixed

packages/tool-cache/__tests__/retry-helper.test.ts
immense055: i want to learn

PR #374, opened 2020-03-09
Update jest to 25.1

thboop:
Audit tools step is failing, lets up date the packages to address this

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "eslint-plugin-github": "^2.0.0",
     "eslint-plugin-jest": "^22.5.1",
     "flow-bin": "^0.115.0",
-    "jest": "^24.9.0",
+    "jest": "^25.1.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "eslint-plugin-github": "^2.0.0",
     "eslint-plugin-jest": "^22.5.1",
     "flow-bin": "^0.115.0",
-    "jest": "^24.9.0",
+    "jest": "^25.1.0",
```
ericsciple: is this PR causing the \"audit tools\" step to fail?

ericsciple: or was it failing before?

ericsciple: looks like it's failing before. Make sure not adding new problems :)

thboop: It was failing before, the intention was to fix it but i missed a few packages, pushing up more updates :)

PR #382, opened 2020-03-16
Tool cache install from a manifest file

bryanmacfarlane:
If a setup-* tool is cached to insulate against target disriptions and the repo contains a versions-manifest.json advertising versions available by platform, version and arch, this api will match a semver spec (ranges, etc.) against the latest and discover where to pull it from.  The abstraction allows for putting on a CDN and changing locations later which are opaque and non breaking to the consumers (the setup-* tasks).

This also allows the hosted images to cache n versions directly on the image for instant resolution and as older versions roll off the image, it still works but may incur a performance of typically 10s of seconds to pull from the CDN cache.  On miss, the setup task can ultimately still pull from the origin (the source tools distribution) JIT.


bryanmacfarlane:
Adding @ericsciple to have a glance.  This is early exploration on an idea in the referenced ADR.  Basically if a repo has a json file in the root with all the tool binary distributions as releases in that repo, then a setup-*** need only pull that file, at a couple lines and it just works.  It also protects against external service distruptions.  See the ADR referenced.

packages/tool-cache/src/manifest.ts
```diff
@@ -0,0 +1,96 @@
...
+  files: IToolReleaseFile[]
+}
+
+export async function _findMatch(
+  versionSpec: string,
+  stable: boolean,
+  candidates: IToolRelease[]
+): Promise<IToolRelease | undefined> {
+  const archFilter = os.arch()
+  const platFilter = os.platform()
```
maxim-lobanov, line 59: I think we should pass these values from action itself instead of getting from `os` so different actions will be able to specify these values differently. For example, `platFilter` will be different:
- Python (`macos-1014`, `ubuntu-1604`, `ubuntu-1804`, `windows-2016`)
- Node (`linux`, `darwin`, `windows`) 

For example, currently, we install package `python-3.0.1-macos-1014-x64.tar.gz` to both `macOS 10.14` and `macOS 10.15` images. But if we find any issue with 10.15 tomorrow, we will push additional asset to releases `python-3.0.1-macos-1015-x64.tar.gz` and will modify task to consume it for macOS 10.15.

`platform` value in `versions-manifest.json` doesn't point to target platform where this package will be used. It points to source platform where this package was built. (it was done because the single package can be used on a few platforms).
For example, Node.JS doesn't depend on version of linux, we have the single package for all linux systems. But Python depends. So this mapping `os.platform` -> `versions-manifest.platform` can be different for different tools

If we pass these values from action to toolkit, our code will be more flexible from this side.

packages/tool-cache/src/tool-cache.ts
```diff
@@ -454,6 +455,42 @@ export function findAllVersions(toolName: string, arch?: string): string[] {
...
+
+export async function getManifestFromUrl(url: string): Promise<IToolRelease[]> {
+  const http: httpm.HttpClient = new httpm.HttpClient('tool-cache')
+
+  //let versions: IToolRelease[] =
+  return (await http.getJson<IToolRelease[]>(url)).result || []
+  //return
+}
+
+export async function findFromManifest(
```
maxim-lobanov, line 479: As I understand, currently, we should use toolkit the following way:
```
const manifest = toolkit.getManifestFromUrl("<url>");
const release = toolkit.findFromManifest(versionSpec, stable, manifest);
```

I think it will be better to pass manifestUrl to `findFromManifest` and invoke `getManifestFromUrl` there:
```
export async function findFromManifest(
  versionSpec: string,
  stable: boolean,
  manifestUrl: string
): Promise<IToolRelease | undefined> {
  const manifest = getManifestFromUrl(manifestUrl);
  const match: mm.IToolRelease | undefined = await mm._findMatch(
    versionSpec,
    stable,
    manifest
  )

  return match
}

```
So it will be easier to use

packages/tool-cache/src/manifest.ts
```diff
@@ -0,0 +1,96 @@
...
+
+  let result: IToolRelease | undefined
+  let match: IToolRelease | undefined
+
+  let files: IToolReleaseFile | undefined
```
maxim-lobanov, line 64: `asset` or `file` will be better names because it is the single file

packages/tool-cache/src/tool-cache.ts
```diff
@@ -454,6 +455,42 @@ export function findAllVersions(toolName: string, arch?: string): string[] {
...
+  // wrap the internal impl
+  const match: mm.IToolRelease | undefined = await mm._findMatch(
+    versionSpec,
+    stable,
+    manifest
+  )
+
+  return match
+}
+
```
maxim-lobanov, line 493: I think that function `installAsset` is a good candidate for toolkit:
```
export async function installReleaseFile(release: IToolRelease): Promise<void> {
  const releaseFileUrl = release.files[0];
  // 1. download file
  // 2. unpack archive
  // 3. run setup.ps1 script from archive if it exists
}
```
This logic will work for all tools so it can be reused between actions.
`setup.ps1` script copies archive content to hosted tool-cache directory and and perform any other actions if needs.
Please see [nix example](https://github.com/actions/python/blob/master/installers/nix-setup-template.ps1) and [win example](https://github.com/actions/python/blob/master/installers/win-setup-template.ps1).

packages/tool-cache/src/manifest.ts
bryanmacfarlane: Yes, I had a TODO to handle this next since I saw that in the python json.

We haven't and shouldn't push the problem to the action author.  All of that information can be read from the environment (and mocked in tests).  Remember the toolkit already reads this from the environment in other places.  For example, when it writes what's downloaded to the cache.

However supporting platforms like `ubuntu-1804` should be straight forward.  We can match on os.platform(), then ImageOS variable (set by hosted env or provisioner) and optionally os.platform() + versionMMmm 

I'll make that change next so it's clearer.

packages/tool-cache/src/tool-cache.ts
bryanmacfarlane: Good suggestion.  I will make this change.

packages/tool-cache/src/manifest.ts
bryanmacfarlane: OK.

packages/tool-cache/src/tool-cache.ts
bryanmacfarlane: Actions are pure javascript and the runner carries the runtime ( a version of node 12 to run them).  The runner currently runs actions inside of containers (from the host) and self-hosted machines that do not have powershell.  We even run actions inside alpine containers (8 MB) and the runner carries an alpine version of node to do that.

Hold off on this and it will be clearer when we integrate these changes into something like setup-node with progressive checks on cache, dist and then source.

packages/tool-cache/src/manifest.ts
bryanmacfarlane: Discussed with Maxim.  We are going to split platform and platform_version in the json.  It will match and platform and if optional version filter is specified it will match on that version spec.

packages/tool-cache/src/tool-cache.ts
bryanmacfarlane: Actually, I remembered why I split.  While the common case may be to get the manifest from an url, I didn't want to dictate that or allow it to come from other places.   Separating the retrieval of the manifest from the querying, allows that flexibility and two lines is still not bad :)

packages/tool-cache/src/manifest.ts
bryanmacfarlane: done

packages/tool-cache/src/tool-cache.ts
bryanmacfarlane: Also note that there's already existing methods tool-cache to install into the cache. See it used here: https://github.com/actions/setup-node/blob/master/src/installer.ts#L173  As part of this change, I don't want to revisit the download, extract and cache.  We're just covering another way to discover an url by version spec.

packages/tool-cache/src/tool-cache.ts
```diff
@@ -454,6 +455,42 @@ export function findAllVersions(toolName: string, arch?: string): string[] {
...
+//
+// 2. if no match, download
+//   a. query versions manifest to match
+//   b. if no match, fall back to source if exists (tool distribution)
+//   c. with download url, download, install and preprent path
+
+export type IToolRelease = mm.IToolRelease
+export type IToolReleaseFile = mm.IToolReleaseFile
+
+export async function getManifestFromUrl(url: string): Promise<IToolRelease[]> {
```
ericsciple, line 471: How confident do you feel about the behavior?

To buy leeway, I'm wondering if whether it would be good to add a function description, indicating experimental and behavior may change.

bryanmacfarlane: Good suggestion.  I may go one step further and release a actions/tool-cache@preview out of this branch.  Same on the setup-node side.  Start a v2-beta

packages/tool-cache/src/manifest.ts
```diff
@@ -0,0 +1,158 @@
...
+      (!stable || candidate.stable === stable)
+    ) {
+      file = candidate.files.find(item => {
+        debug(
+          `${item.arch}===${archFilter} && ${item.platform}===${platFilter}`
+        )
+
+        let chk = item.arch === archFilter && item.platform === platFilter
+        if (chk && item.platform_version) {
+          const osVersion = module.exports._getOsVersion()
```
ericsciple, line 89: super duper nit: have you considered calling using the syntax `_getOsVersion()` ? i think that would work too

ericsciple, line 89: Since `_getOsVersion()` is heavy (launches a process (mac) or reads a file from disk (linux)), consider caching the result.

For example:

```ts
          if (osVersion === undefined) {
            osVersion = _getOsVersion()
          }
```

packages/tool-cache/src/manifest.ts
```diff
@@ -0,0 +1,158 @@
...
+      if (file) {
+        debug(`matched ${candidate.version}`)
+        match = candidate
+        break
+      }
+    }
+  }
+
+  if (match && file) {
+    // clone since we're mutating the file list to be only the file that matches
```
ericsciple, line 110: +1 for cloning the object :)

packages/tool-cache/src/manifest.ts
ericsciple: nevermind i just realized this must be for unit testing

.eslintrc.json
```diff
@@ -18,7 +18,7 @@
     "@typescript-eslint/await-thenable": "error",
     "@typescript-eslint/ban-ts-ignore": "error",
     "camelcase": "off",
-    "@typescript-eslint/camelcase": "error",
+    "@typescript-eslint/camelcase": "off",
```
bryanmacfarlane, line 21: I got tired of adding all the line by line exceptions.  Fact is as we deserialize these requests into structures dictated externally, many have non camel cased properties.

packages/tool-cache/src/tool-cache.ts
```diff
@@ -32,7 +34,8 @@ const userAgent = 'actions/tool-cache'
  */
 export async function downloadTool(
```
joshmgross, line 35: Can we update the function's description to include this new input?

PR #384, opened 2020-03-16
Update Status badge to point to master

thboop:


thboop:
Looks like it does this by default now, master is failing because a package is out of date

PR #385, opened 2020-03-18
Check if tool path exists before executing

reiniertimmer:
For: tool-cache

This clears up the error messages when the runner does not have a `unzip` or `powershell` binary installed. This can be an issue on self-hosted runners, and makes it easier to track the root cause of the error.


reiniertimmer:
Is there no interest in this? Otherwise I could just close this PR

thboop:
Hey @reiniertimmer , sorry for the delay thank you for the contribution!

PR #396, opened 2020-03-30
Updates to @actions/artifact

konradpabjan:
Updates to the `@actions/artifact` package

- Fix issues related to artifacts not downloading correctly
- Support for handling 429 http response codes which can be sent if the user is making too many requests
-  Support retries when downloading an artifact, previously if something exploded, it wouldn't even retry. Supports the same set of retryable codes as upload-artifact
- Exponential backoff when downloading or uploading an artifact and a retry code is encountered. Previously a fixed interval was always used
- Refactor `getRequestOptions` to `getUploadRequestOptions` and `getDownloadRequestOptions` and add tests so it is easier to keep track of headers that are used during upload and download
- Refactor `UploadStatusReporter` to `StatusReporter` and change it so it can also be used when downloading an artifact. Now we have detailed status information when downloading an artifact, displays the percentage every 1 second (example output below)

image

Retry logic for downloading a file. The same chart is getting added to the documentation around implementation details. You can see it  🎨 [here](https://github.com/actions/toolkit/blob/konradpabjan/artifact-fixes/packages/artifact/docs/implementation-details.md) 🖌

image

## Testing

Succesfull upload/download
https://github.com/konradpabjan/artifact-test/actions/runs/69029543

Upload using v1 -> download using v2 -> upload again using v2 -> download again using v1

For testing I have been uploading about 50MB of node modules from the `typescript` npm package.

Afterwards, I downloaded the artifacts, and compared files using `fc.exe /b` in powershell. I was able to test `.js`, '`json`, `.txt`, `.md` and `.d` The only thing that is different is that in certain files, line endings are different (`crlf` vs `lf`), but after changing them in `notepad ++`, `FC` reports no differences encountered.


packages/artifact/src/internal/download-http-client.ts
```diff
@@ -124,66 +194,117 @@ export class DownloadHttpClient {
...
+        await this.pipeResponseToFile(
           response,
-          stream,
+          destinationStream,
           isGzip(response.message.headers)
         )
+        return
+      } else if (isThrottledStatusCode(response.message.statusCode)) {
+        info(
+          'A 429 response code has been recieved when attempting to download an artifact'
```
pjquirk, line 240: Nit, typo: should be `received`

konradpabjan: Quite a few spelling bloopers 😑 My spelling extension has been off for some reason

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -112,10 +138,54 @@ export class DownloadHttpClient {
...
+
+    // Back off exponentially based off of the retry count
+    const backoffExponentially = async (): Promise<void> => {
+      this.downloadHttpManager.disposeAndReplaceClient(httpClientIndex)
+      const backoffTime = getExponentialRetryTimeInMilliseconds(retryCount)
+      info(
+        `Exponential backoff for retry #${retryCount}. Waiting for ${backoffTime} milliseconds before continuing the download`
+      )
+      await new Promise(resolve =>
+        setTimeout(resolve, getExponentialRetryTimeInMilliseconds(retryCount))
```
yacaovsnc, line 182: Isn't this the `backoffTime` from above?

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -112,10 +138,54 @@ export class DownloadHttpClient {
...
+          console.log(response)
+        }
+        throw new Error(
+          `Unable to download ${artifactLocation}. Retry limit has been reached`
+        )
+      }
+    }
+
+    // Back off exponentially based off of the retry count
+    const backoffExponentially = async (): Promise<void> => {
```
yacaovsnc, line 175: I would suggest DRY here again.  This function is largely the same as the other `backOffUsingRetryValue`, both can be simplified to a single function `backoff(timespan)`.

yacaovsnc: Also, if `incrementAndCheckRetryLimit` is **always** called before backoff exponentially, you can even fold that method into `backoff`.  If a value is passed it, use that timeout value; otherwise calculate the retry automatically and backoff.

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -124,66 +194,117 @@ export class DownloadHttpClient {
...
         const gunzip = zlib.createGunzip()
         response.message
           .pipe(gunzip)
-          .pipe(stream)
+          .pipe(destinationStream)
           .on('close', () => {
             resolve()
           })
+          .on('error', error => {
+            info(
```
yacaovsnc, line 289: If you want to add an info before we throw the error, consider adding some more information to it so it's not a static sentence.  If it shows which file failed it's more useful, otherwise the error below would be enough?

packages/artifact/src/internal/upload-http-client.ts
```diff
@@ -359,7 +363,54 @@ export class UploadHttpClient {
...
+
+    const backOffUsingRetryValue = async (
+      retryAfterValue: number
+    ): Promise<void> => {
+      info(`disposing client with index ${httpClientIndex}`)
+      this.uploadHttpManager.disposeAndReplaceClient(httpClientIndex)
+      info(
+        `Backoff due to too many requests, retry #${retryCount}. Waiting for ${retryAfterValue} milliseconds before continuing the upload`
+      )
+      info(
```
yacaovsnc, line 409: Where is the `await`?

packages/artifact/src/internal/download-http-client.ts
konradpabjan: I added some extra information here. The raw error might be something ambiguous related to gzip or the file system so to the end user they might have no idea what is going.

An extra `info(..)` can give users at least some extra clarity

packages/artifact/src/internal/upload-http-client.ts
konradpabjan: OOF, nice catch

packages/artifact/src/internal/download-http-client.ts
konradpabjan: Copy-Pasta Error by me again, nice catch 😖

packages/artifact/src/internal/download-http-client.ts
konradpabjan: I did some pretty big refactoring in both` upload-http-client` and `download-http-client`

In combined both of the back off functions into a single `backOff()` for both http clients, while in download I also combined `incrementAndCheckRetryLimit()` into the same function. Simplifies it a little bit more.

packages/artifact/src/internal/download-http-client.ts
yacaovsnc: If the error doesn't give enough information to the user, then consider throw a new user facing error, and log the existing error for us to debug.  

At the high level, errors should be indicative of the issue, and best give the user some way to move forward themselves.

packages/artifact/src/internal/config-variables.ts
```diff
@@ -6,12 +6,16 @@ export function getUploadChunkSize(): number {
...
 
-export function getUploadRetryCount(): number {
-  return 3
+export function getRetryLimit(): number {
+  return 5
 }
 
-export function getRetryWaitTimeInMilliseconds(): number {
-  return 10000
+export function getRetryMultiplier(): number {
```
joshmgross, line 13: Could you add comments explaining what each config option does?

It's not clear to me what exactly `RetryMultiplier` controls

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -81,22 +85,44 @@ export class DownloadHttpClient {
...
+          currentFile += 1
+
+          const startTime = performance.now()
           await this.downloadIndividualFile(
             index,
             currentFileToDownload.sourceLocation,
             currentFileToDownload.targetPath
           )
+
+          debug(
```
joshmgross, line 113: You could wrap these debug calls in an if statement with `core.isDebug` so that we can skip the performance math when it's not needed. See https://github.com/actions/toolkit/tree/master/packages/core#logging

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -124,66 +160,124 @@ export class DownloadHttpClient {
...
+            )
+          : await backOff()
       } else {
+        // Some unexpected response code, fail immediately and stop the download
         // eslint-disable-next-line no-console
-        console.log(retryResponse)
-        throw new Error(`Unable to download ${artifactLocation}`)
+        console.log(response)
+        throw new Error(
+          `###ERROR### Unable to download ${artifactLocation} ###`
```
joshmgross, line 235: What's up with the `###ERROR###` here? If we want to log an error you can just use `core.Error` and then throw the normal message.
We should also be logging `response` as an error too.

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -124,66 +160,124 @@ export class DownloadHttpClient {
...
-        response.message.pipe(stream).on('close', () => {
-          resolve()
-        })
+        response.message
+          .pipe(destinationStream)
+          .on('close', () => {
+            resolve()
+          })
+          .on('error', error => {
+            info(
```
joshmgross, line 274: These errors should not be output to `info`, use `core.error` so that it's easy to find in the logs.

packages/artifact/src/internal/status-reporter.ts
```diff
@@ -0,0 +1,94 @@
...
+    this.largeFileStatus = undefined
+    this.displayFrequencyInMilliseconds = displayFrequencyInMilliseconds
+  }
+
+  setTotalNumberOfFilesToProcess(fileTotal: number): void {
+    this.totalNumberOfFilesToProcess = fileTotal
+  }
+
+  start(): void {
+    const _this = this
```
joshmgross, line 30: Is there a benefit to doing this? Why not use `this` everywhere?

packages/artifact/src/internal/upload-http-client.ts
```diff
@@ -359,58 +363,91 @@ export class UploadHttpClient {
...
+    // Increments the current retry count and then checks if the retry limit has been reached
+    // If there have been too many retries, fail so the download stops
+    const incrementAndCheckRetryLimit = (
+      message?: IHttpClientResponse
+    ): boolean => {
+      retryCount++
+      if (retryCount > retryLimit) {
+        if (message) {
+          // eslint-disable-next-line no-console
+          console.log(message)
```
joshmgross, line 377: If there's an eslint rule against this, we can just use `core.info`

packages/artifact/src/internal/utils.ts
```diff
@@ -42,11 +66,48 @@ export function isRetryableStatusCode(statusCode?: number): boolean {
...
   return retryableStatusCodes.includes(statusCode)
 }
 
+export function isThrottledStatusCode(statusCode?: number): boolean {
+  if (!statusCode) {
+    return false
+  }
+
+  // TODO, change when a new version of @actions/http-client gets released
+  return statusCode === 429
```
joshmgross, line 81: Why is this new version not released yet? Would be good to consume those changes since it's already merged to master

packages/artifact/src/internal/upload-http-client.ts
konradpabjan: Not quite, `core.info` only takes a string, while `console.log()` can take any type of object, that's why in a lot of places I'm having to use `console.log()`. 

Doing `core.info(whatever.toString())` also doesn't cut it because certain types of objects like json strings become near unreadable

packages/artifact/src/internal/utils.ts
konradpabjan: New version is not yet released yet. I'm following up on that: https://www.npmjs.com/package/@actions/http-client

packages/artifact/src/internal/status-reporter.ts
konradpabjan: There is a interesting typescript rule that comes up which is why I had to set it up the way it is:

image

packages/artifact/src/internal/download-http-client.ts
konradpabjan: It's not so much about logging an error, but here we want to throw one so that execution halts. 

`core.error(response.toString())` is useless because it just outputs `[object Object]` which is why you see `console.log` being used in a few places.

I've also cleaned this up a little. Outputting the entire response if pretty useless and just pollutes the logs, the only thing we really care about is the status code and url that we're using for the request.

packages/artifact/src/internal/status-reporter.ts
joshmgross: Are you referring to the implicit any error or the outer value of this?

Either way, this seems way an odd way to work around it. Is this the recommended solution to that outer value this?

packages/artifact/src/internal/upload-http-client.ts
joshmgross: Could you post an example of what you're trying to log?

It seems like an explicit error format or some even just `JSON.stringify` would be better than just throwing the error into `console.log`

packages/artifact/src/internal/status-reporter.ts
pjquirk: It looks like [arrow functions](https://www.typescriptlang.org/docs/handbook/functions.html#this-and-arrow-functions) would solve this issue, it makes `this` behave like you'd expect (coming from a C# world).

packages/artifact/src/internal/upload-http-client.ts
konradpabjan: In this particular case, `message` is of type `http.IncomingMessage`. https://github.com/actions/http-client/blob/60c8242c8cbb55e4ac31434511874b68bad4afa8/index.ts#L68
It's not really an error format, but rather an object dump of what is part of the http call.

If there are more than 5 retryable status codes during a single file upload then who knows what could be going on which is why I output the entire message. It has headers, raw headers and a host of other information that doesn't get logged into Kusto which i've learned the hard way, here is an example (very long output, but it basically everything we could ever want to know) . https://github.com/actions/upload-artifact/issues/62#issuecomment-606562787

packages/artifact/src/internal/status-reporter.ts
konradpabjan: When I added this code in my previous PR, there was some stack overflow answer that recommended using `_this` which is why I went down that route (can't find the exact thread now).

Basically though, there are 3 ways that you can deal with `this` while inside another function. All of them work, each one is acceptable, just important to be consistent.

https://medium.com/@denny.headrick/this-bind-this-a-javascript-story-313ff84a12c8
https://github.com/airbnb/javascript/issues/81#issuecomment-23173941

I've changed everything to arrow functions, seems nicer.

packages/artifact/src/internal/upload-http-client.ts
joshmgross: That seems like a lot more info than is necessary. It would be cleaner to log the status code and the response from the server as an error. If you need the headers, then I'd recommend logging those separately in a debug statement. Logging the entire object seems like the wrong approach here.

packages/artifact/src/internal/utils.ts
konradpabjan: Version 1.0.7 is out

packages/artifact/src/internal/upload-http-client.ts
konradpabjan: I did some refactoring and I added a new `displayHttpDiagnostics` method inside `utils`. You can see it [here](https://github.com/actions/toolkit/blob/a4a3055ab024d935749026095611024390d9c205/packages/artifact/src/internal/utils.ts#L222). I swapped out existing `console.log(response)` with this new method.

In action this is what it looks like:
image

Test run with just to see what it looks like is here: https://github.com/konradpabjan/artifact-test/runs/565950616?check_suite_focus=true#step:5:20

There is some other information that could be useful that I can't display without doing a full `console.log`, but I _think_ this should be sufficient (debugging low level http errors really sucks).

I am using `core.info` instead of `core.debug` because this should only be called when an error is hit. Users should always see this regardless if they have have diagnostics on/off.

packages/artifact/src/internal/artifact-client.ts
```diff
@@ -173,6 +173,7 @@ export class DefaultArtifactClient implements ArtifactClient {
       await createDirectoriesForArtifact(
         downloadSpecification.directoryStructure
       )
+      core.info('Directory structure has been setup for the artifact')
```
pjquirk, line 176: Is it worth including the name of the directory that was created?

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -112,10 +146,20 @@ export class DownloadHttpClient {
     artifactLocation: string,
     downloadPath: string
   ): Promise<void> {
```
pjquirk, line 148: Nit: this method is big enough that I'd prefer it to be its own class, but that may be the C# developer in me.  Not needed for this PR.

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -124,66 +168,128 @@ export class DownloadHttpClient {
...
+      if (retryCount > retryLimit) {
+        return Promise.reject(
+          new Error(
+            `Retry limit has been reached. Unable to download ${artifactLocation}`
+          )
+        )
+      } else {
+        this.downloadHttpManager.disposeAndReplaceClient(httpClientIndex)
+        if (retryAfterValue) {
+          // Back off exponentially based off of the retry count
```
pjquirk, line 185: Nit, comment is wrong, this branch is using the retry-after header

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -124,66 +168,128 @@ export class DownloadHttpClient {
...
+
+    // keep trying to download a file until a retry limit has been reached
+    while (retryCount <= retryLimit) {
+      let response: IHttpClientResponse
+      try {
+        response = await makeDownloadRequest()
+      } catch (error) {
+        // if an error is caught, it is usually indicative of a timeout so retry the download
+        core.info(
+          'An error has been caught, while attempting to download a file'
```
pjquirk, line 212: Nit: `caught` doesn't feel right in an error message to the user.  Maybe `An error occurred while attempting to download a file`?

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -124,66 +168,128 @@ export class DownloadHttpClient {
...
+        )
+        // eslint-disable-next-line no-console
+        console.log(error)
```
pjquirk, line 215: I may have missed some conversation, but why use `console.log` instead of `core.info` or `core.warn`?

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -124,66 +168,128 @@ export class DownloadHttpClient {
...
         response.message
           .pipe(gunzip)
-          .pipe(stream)
+          .pipe(destinationStream)
           .on('close', () => {
             resolve()
           })
+          .on('error', error => {
+            core.error(
+              `An error has been encountered while gunzipping and writing a downloaded file to ${destinationStream.path}`
```
pjquirk, line 275: Nit: replace `gunzipping` with `decompressing`

packages/artifact/src/internal/status-reporter.ts
```diff
@@ -0,0 +1,92 @@
...
+      )
+    }, this.displayFrequencyInMilliseconds)
+
+    // displays extra information about any large files that take a significant amount of time to upload or download every 1 second
+    this.largeFileStatus = setInterval(() => {
+      for (const value of Array.from(this.largeFiles.values())) {
+        info(value)
+      }
+      // delete all entires in the map after displaying the information so it will not be displayed again unless explicitly added
+      this.largeFiles = new Map<string, string>()
```
pjquirk, line 53: Nit: any reason to do this versus `this.largeFiles.clear()`?

packages/artifact/src/internal/utils.ts
```diff
@@ -42,11 +66,47 @@ export function isRetryableStatusCode(statusCode?: number): boolean {
...
+      info(`retry-after headers is present with a value of ${retryTime}`)
+      return retryTime * 1000
+    }
+    info(
+      `returned retry-after header value: ${retryTime} is non-numeric and cannot be used`
+    )
+    return undefined
+  }
+  info(
+    `no retry-after header was found. Dumping all headers for diagnostic purposes`
```
pjquirk, line 103: Nit: capitalize `no`

packages/artifact/src/internal/utils.ts
```diff
@@ -42,11 +66,47 @@ export function isRetryableStatusCode(statusCode?: number): boolean {
...
+  headers: IncomingHttpHeaders
+): number | undefined {
+  if (headers['retry-after']) {
+    const retryTime = Number(headers['retry-after'])
+    if (!isNaN(retryTime)) {
+      info(`retry-after headers is present with a value of ${retryTime}`)
+      return retryTime * 1000
+    }
+    info(
+      `returned retry-after header value: ${retryTime} is non-numeric and cannot be used`
```
pjquirk, line 98: Nit: capitalize `returned`

packages/artifact/src/internal/utils.ts
```diff
@@ -42,11 +66,47 @@ export function isRetryableStatusCode(statusCode?: number): boolean {
...
+ * is originally denoted in seconds, so if present, it is converted to milliseconds
+ * @param headers all the headers received when making an http call
+ */
+export function tryGetRetryAfterValueTimeInMilliseconds(
+  headers: IncomingHttpHeaders
+): number | undefined {
+  if (headers['retry-after']) {
+    const retryTime = Number(headers['retry-after'])
+    if (!isNaN(retryTime)) {
+      info(`retry-after headers is present with a value of ${retryTime}`)
```
pjquirk, line 94: Should be `Retry-After header is present...`

packages/artifact/src/internal/download-http-client.ts
konradpabjan: `console.log` can take `any` as an input while `core.info` and `core.warn` only take a string. It is unknown what format the error will be in so it could just show up as `[object Object]`. `console.log` handles formatting really well.

Doing something like `core.info(JSON.stringify(error, undefined, 2))` should work but I see no real benefit when `console.log` handles everything out of the box.

The only time I'm using `console.log` now is if there is a unknown error type thrown like here. I was using `console.log` earlier in a few places but I swapped that out with a custom `displayHttpDiagnostics` method that displays only certain information after doing some selective formatting and filtering.

packages/artifact/src/internal/artifact-client.ts
konradpabjan: Not worth it. There is actually more than 1 directory that gets created and this all happens before any downloading actually starts.

Issue #386, opened 2020-03-18
core.setFailed should accept Errors


jasonkarns:
**Describe the enhancement**
It's extremely common (nay, the _purpose_) of `core.setFailed` to report Errors that are caught during an action. And yet, its method signature does not accept `Error` instances.

It would be beneficial to accept Error instances instead of requiring users to convert to strings.

**Code Snippet**
This would be _especially_ beneficial for promise chains:

```js
somePromise()
  .then(doThing)
  .then(doOther)
  .then(whatever)
  .catch(core.setFailed)
```

Present behavior:

`core.setFailed(Error("foo"))` results in: `TypeError: (s || "").replace is not a function`

thboop:
I agree.

bnb:
+1. Putting the burden on users to pass the proper type rather than handling coercion in the module is not good DX.

thboop:
Resolved in #405

PR #405, opened 2020-04-06
Be more lenient in accepting command inputs

### LINKED ISSUE #398: TypeError: (s || "").replace is not a function
**Describe the bug**
When trying to run an action, I am getting the following error from `@actions/core`:

```
(node:2588) UnhandledPromiseRejectionWarning: TypeError: (s || "").replace is not a function
    at escapeData (/home/runner/work/_actions/cutenode/action-meeting/dev/node_modules/@actions/core/lib/command.js:66:10)
    at Command.toString (/home/runner/work/_actions/cutenode/action-meeting/dev/node_modules/@actions/core/lib/command.js:60:35)
    at issueCommand (/home/runner/work/_actions/cutenode/action-meeting/dev/node_modules/@actions/core/lib/command.js:23:30)
    at Object.issue (/home/runner/work/_actions/cutenode/action-meeting/dev/node_modules/@actions/core/lib/command.js:27:5)
    at error (/home/runner/work/_actions/cutenode/action-meeting/dev/node_modules/@actions/core/lib/core.js:127:15)
    at Object.setFailed (/home/runner/work/_actions/cutenode/action-meeting/dev/node_modules/@actions/core/lib/core.js:101:5)
    at action (/home/runner/work/_actions/cutenode/action-meeting/dev/index.js:62:10)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(node:2588) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:2588) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

**To Reproduce**
Last commit where it works:
- https://github.com/cutenode/action-meeting/commit/f88068b6ec7c79c60712f1ec7cedf71d13aaada4

It stops working in this commit:
- https://github.com/cutenode/action-meeting/commit/71fa544290caaf49e3d17626208b5b9c2a3875dc

**EDIT:** Updated to only reflect the commit it stopped working in. This was triggered by passing an incorrect type to Octokit.js for the `labels` property - I was passing a string and it required an array. Since the project was erroring because of that and the error being passed was not a string, I was encountering this issue.

**Expected behavior**
For `@actions/core` to handle errors more effectively.

thboop:
Users have been encountering friction when trying to use command inputs.

[This commit](https://github.com/actions/toolkit/commit/8b0300129f08728419263b016de8630f1d426d5f#diff-30c957fdb0e7818041c155a0e3ac6e0b) broke some our handling when users don't pass in strings. We re-added that handling and implemented a test for it. 

We also have had a number of users ask for additional input types. 
https://github.com/actions/toolkit/issues/386 -> Set Failed and Error should accept errors
https://github.com/actions/toolkit/issues/370 -> Set Output should format data into strings

This commit resolves those use cases.

Resolves #398, #386, #370

### DIFF
--- a/packages/core/src/command.ts
+++ b/packages/core/src/command.ts
@@ -64,15 +64,13 @@ class Command {
             } else {
               cmdStr += ','
             }
-            // safely escape the property - avoid blowing up when attempting to
-            // call .replace() if property is not a string for some reason
+
             cmdStr += `${key}=${escapeProperty(val)}`
           }
         }
       }
     }
-    // safely append the message - avoid blowing up when attempting to
-    // call .replace() if message is not a string for some reason
+
     cmdStr += `${CMD_STRING}${escapeData(this.message)}`
     return cmdStr
   }
--- a/packages/core/src/command.ts
+++ b/packages/core/src/command.ts
@@ -61,13 +61,15 @@ class Command {
             } else {
               cmdStr += ','
             }
-
+            // safely escape the property - avoid blowing up when attempting to
+            // call .replace() if property is not a string for some reason
             cmdStr += `${key}=${escapeProperty(`${val || ''}`)}`
           }
         }
       }
     }
-
+    // safely append the message - avoid blowing up when attempting to
+    // call .replace() if message is not a string for some reason
     cmdStr += `${CMD_STRING}${escapeData(`${this.message || ''}`)}`
     return cmdStr
   }
--- a/packages/core/src/command.ts
+++ b/packages/core/src/command.ts
@@ -62,13 +62,13 @@ class Command {
               cmdStr += ','
             }
 
-            cmdStr += `${key}=${escapeProperty(val)}`
+            cmdStr += `${key}=${escapeProperty(`${val || ''}`)}`
           }
         }
       }
     }
 
-    cmdStr += `${CMD_STRING}${escapeData(this.message)}`
+    cmdStr += `${CMD_STRING}${escapeData(`${this.message || ''}`)}`
     return cmdStr
   }
 }
--- a/packages/core/src/command.ts
+++ b/packages/core/src/command.ts
@@ -2,8 +2,11 @@ import * as os from 'os'
 
 // For internal use, subject to change.
 
+// We use any as a valid input type
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
 interface CommandProperties {
-  [key: string]: string
+  [key: string]: any
 }
 
 /**
@@ -19,7 +22,7 @@ interface CommandProperties {
 export function issueCommand(
   command: string,
   properties: CommandProperties,
-  message: string
+  message: any
 ): void {
   const cmd = new Command(command, properties, message)
   process.stdout.write(cmd.toString() + os.EOL)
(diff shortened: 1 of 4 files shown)


bryanmacfarlane:
interested in @ericsciple feedback

packages/core/src/command.ts
```diff
@@ -61,14 +61,16 @@ class Command {
             } else {
               cmdStr += ','
             }
-
-            cmdStr += `${key}=${escapeProperty(val)}`
+            // safely escape the property - avoid blowing up when attempting to
```
ericsciple, line 64: Should we fix the typings instead?

ericsciple, line 64: It looks like these lines should be reverted. The functions `escapeData` and `escapeProperty` already OR the value with empty string

packages/core/src/core.ts
```diff
@@ -204,3 +214,20 @@ export function saveState(name: string, value: string): void {
 export function getState(name: string): string {
   return process.env[`STATE_${name}`] || ''
 }
+
+/**
+ * Sanatizes an input into a string so it can be passed into issueCommand safely
+ * @param input input to sanitize into a string
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function sanitizeInput(input: any): string {
```
ericsciple, line 223: Would `toCommandValue` (or something similar) be a better name? "Sanitize" seems like a misnomer.

packages/core/src/core.ts
```diff
@@ -204,3 +214,20 @@ export function saveState(name: string, value: string): void {
...
+  if (input == null) {
+    return ''
+  } else if (typeof input === 'string' || input instanceof String) {
+    return input as string
+  } else if (input instanceof Error) {
+    return input.toString()
+  } else {
```
ericsciple, line 230: should special case `null` and `undefined` (return empty string)

packages/core/src/core.ts
```diff
@@ -204,3 +214,20 @@ export function saveState(name: string, value: string): void {
...
   return process.env[`STATE_${name}`] || ''
 }
+
+/**
+ * Sanatizes an input into a string so it can be passed into issueCommand safely
+ * @param input input to sanitize into a string
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function sanitizeInput(input: any): string {
+  if (input == null) {
```
ericsciple, line 224: suprised linter didnt complain. Can we check `=== null` and `=== undefined`?

packages/core/src/core.ts
```diff
@@ -191,8 +198,11 @@ export async function group<T>(name: string, fn: () => Promise<T>): Promise<T> {
  * @param     name     name of the state to store
  * @param     value    value to store
  */
-export function saveState(name: string, value: string): void {
-  issueCommand('save-state', {name}, value)
+export function saveState(
+  name: string,
+  value: string | number | boolean
```
ericsciple, line 203: if we accept any, shouldnt we fix the type to be any? and comment should describe behavior (e.g. tojson for objects)

packages/core/src/core.ts
ericsciple, line 223: have you considered driving this change into command.ts instead? Lowest level would be `escapeData` and `escapeProperty` function to call.

packages/core/src/core.ts
```diff
@@ -121,8 +128,8 @@ export function debug(message: string): void {
  * Adds an error issue
  * @param message error issue message
  */
-export function error(message: string): void {
-  issue('error', message)
+export function error(message: string | Error): void {
```
ericsciple, line 131: should `warning` allow Error also?

packages/core/src/core.ts
```diff
@@ -204,3 +214,20 @@ export function saveState(name: string, value: string): void {
...
+ * Sanatizes an input into a string so it can be passed into issueCommand safely
+ * @param input input to sanitize into a string
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function sanitizeInput(input: any): string {
+  if (input == null) {
+    return ''
+  } else if (typeof input === 'string' || input instanceof String) {
+    return input as string
+  } else if (input instanceof Error) {
```
ericsciple, line 228: is converting error to string only something the error/setFailed functions should do? I'm thinking about the description for the setOutput function... seems like weird behavior.

packages/core/src/core.ts
thboop: Thats intentional but we can make it more clear.

```
null == undefined
true
null == null
true
null === undefined
false
null === null
true
```

packages/core/src/command.ts
thboop: Those functions only || with an empty string if the value passed is falsey. Booleans, ints, and objects with no `.replace` function crash when passed into those functions because `123 || ""` will return 123.

packages/core/src/core.ts
thboop: Setting process.env in exportVariable makes that hard
https://github.com/actions/toolkit/pull/405/files#diff-396527b4d62db15ebbfb2845654a161fR43

ericsciple: Ah good point regarding process.env.

Is command.ts internal only? If so then it should assume type is always string. Otherwise should update types to allow any.

packages/core/src/core.ts
thboop: I enabled this for warning. I don't think we need to enable it for info or debug at this time, but if we get feature requests for it we can.

packages/core/src/command.ts
```diff
@@ -61,27 +64,42 @@ class Command {
             } else {
               cmdStr += ','
             }
-
+            // safely escape the property - avoid blowing up when attempting to
```
ericsciple, line 67: it looks like these comments can be deleted now: `safely escape [...]`

now the type is `any` so `for some reason` language not needed anymore

jasonkarns:
Late to the party here ( 🎉 for resolving!!)... but shouldn't the `Error` stringification use `e.message` instead of `e.toString()`? The `toString` method prepends `"Error: "`, to the message which is likely redundant and not desirable.

PR #408, opened 2020-04-08
@actions/artifact package updates

konradpabjan:
## Overview

- Prepare for 0.3.0 release
    - Update `RELEASES.md` with changelog info
    - Bump 0.2.0 -> 0.3.0 in `package.json` and `package-lock.json`
-  Storage Quota Error
    - During artifact upload, if a user has hit their storage quota, a `403` might be returned when trying to create a new container for an artifact. There is now a clear error message that lets the user know they can't upload any more artifacts because of storage. For more info, see https://github.com/actions/upload-artifact/issues/62#issuecomment-608313373
    -  Extra test has been added to `upload.test.ts` to test a `403` response code along with an extra test in `util.test.ts`
- Empty file downloads
    - The `v1` action of `download-artifact` has some extra logic when dealing with empty files. You can find it [here](https://github.com/actions/runner/blob/6c70d53eead402ba5d53676d6ed649a04e219c9b/src/Runner.Plugins/Artifact/FileContainerServer.cs#L128). The updates in the PR largely aim to replicate this behavior so there are no extra HTTP calls made when downloading empty files.
     - `download-specification` has been updated with a new `emptyFilesToCreate` array that keeps track of empty files that should be created
     -  `util.ts` has been updated with a new method that creates empty files
     -  tests have been added to make sure the `download-specification` correctly picks up empty files and that empty files actually get created
- Misc
     - Some asyncification of some previous sync methods that were used in tests
     - Small spelling error fixes

## Testing

- Unit tests added/updated for all the changes
- End-To-End tests (with some empty files being uploaded , you can see some runs here:
     - https://github.com/konradpabjan/artifact-test/actions/runs/73901004
     - https://github.com/konradpabjan/artifact-test/actions/runs/73900876

### DIFF
--- a/packages/artifact/__tests__/download-specification.test.ts
+++ b/packages/artifact/__tests__/download-specification.test.ts
@@ -21,7 +21,8 @@ function getPartialContainerEntry(): ContainerEntry {
     lastModifiedBy: '82f0bf89-6e55-4e5a-b8b6-f75eb992578c',
     itemLocation: 'ADD_INFORMATION',
     contentLocation: 'ADD_INFORMATION',
-    contentId: ''
+    contentId: '',
+    fileLength: 100
   }
 }
 
@@ -73,12 +74,12 @@ function createContentLocation(relativePath: string): string {
                     /dir4
                         file4.txt
                         file5.txt
-
+                        file6.txt (empty file)
     /my-artifact-extra
         /file1.txt
 */
 
-// main artfact
+// main artifact
 const file1Path = path.join(artifact1Name, 'file1.txt')
 const file2Path = path.join(artifact1Name, 'file2.txt')
 const dir1Path = path.join(artifact1Name, 'dir1')
@@ -88,6 +89,7 @@ const dir3Path = path.join(dir2Path, 'dir3')
 const dir4Path = path.join(dir3Path, 'dir4')
 const file4Path = path.join(dir4Path, 'file4.txt')
 const file5Path = path.join(dir4Path, 'file5.txt')
+const file6Path = path.join(dir4Path, 'file6.txt')
 
 const rootDirectoryEntry = createDirectoryEntry(artifact1Name)
 const directoryEntry1 = createDirectoryEntry(dir1Path)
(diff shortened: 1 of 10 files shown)


packages/artifact/src/internal/utils.ts
```diff
@@ -296,3 +304,11 @@ export async function createDirectoriesForArtifact(
     })
   }
 }
+
+export async function createEmptyFilesForArtifact(
```
konradpabjan, line 308: I followed the guidance [here](https://stackoverflow.com/questions/12809068/create-an-empty-file-in-node-js) on how to create an empty file in node.

packages/artifact/__tests__/util.test.ts
```diff
@@ -216,12 +227,43 @@ describe('Utils', () => {
...
+
+    const emptyFile1 = path.join(root, 'emptyFile1')
+    const directoryToCreate = path.join(root, 'folder1')
+    const emptyFile2 = path.join(directoryToCreate, 'emptyFile2')
+
+    // empty files should only be created after the directory structure is fully setup
+    // ensure they are first created by using the createDirectoriesForArtifact method
+    const directoryStructure = [root, directoryToCreate]
+    await utils.createDirectoriesForArtifact(directoryStructure)
+    await expect(fs.promises.access(root)).resolves.toEqual(undefined)
```
konradpabjan, line 251: For asynchronously checking if files exist, I used [this](https://stackoverflow.com/questions/4482686/check-synchronously-if-file-directory-exists-in-node-js) for guidance

packages/artifact/__tests__/download-specification.test.ts
```diff
@@ -99,6 +101,9 @@ const fileEntry2 = createFileEntry(file2Path)
 const fileEntry3 = createFileEntry(file3Path)
 const fileEntry4 = createFileEntry(file4Path)
 const fileEntry5 = createFileEntry(file5Path)
+const fileEntry6 = createFileEntry(file6Path)
+fileEntry6.fileLength = 0 // empty file path
+fileEntry5.fileLength = undefined // one file does not have a fileLength
```
joshmgross, line 106: Could we name these file entries to be more descriptive? Something like:
```suggestion
const missingLengthFileEntry = createFileEntry(file5Path)
missingLengthFileEntry.fileLength = undefined // one file does not have a fileLength

const emptyFileEntry = createFileEntry(file6Path)
emptyFileEntry.fileLength = 0 // empty file length
```

packages/artifact/__tests__/util.test.ts
joshmgross: What's wrong with `fs.existsSync`?

konradpabjan: `fs.existsSync` is perfectly fine however I'm trying to use async/promises wherever possible. In general any sync operation will block the event loop until the operation is complete so asyncification is usually a good thing.

packages/artifact/__tests__/download-specification.test.ts
konradpabjan: 👍 Had to update the names in a few more places

PR #410, opened 2020-04-10
Update eslint to 2.2.7

### LINKED ISSUE #291: Eslint Version is not compatible with Typescript version
**Describe the bug**
See the output from the build logs below:
```
WARNING: You are currently running a version of TypeScript which is not officially supported by typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.2.1 <3.5.0

YOUR TYPESCRIPT VERSION: 3.7.4
```

Eslint creates warnings on typescript features that were introducted in v3.5.0 or later. We should bump the linter to its latest version.

thboop:
Resolves #291 

Updates EsLint to get rid of the unsupported typescript warning every time we run it!

To do this, we had to create a separate tsconfig for the linter, as there was a [breaking change](https://github.com/typescript-eslint/typescript-eslint/issues/853#issuecomment-521523708).

### DIFF
--- a/packages/github/__tests__/github.test.ts
+++ b/packages/github/__tests__/github.test.ts
@@ -11,7 +11,7 @@ describe('@actions/github', () => {
 
   beforeAll(async () => {
     // Start proxy server
-    proxyServer = proxy() as http.Server
+    proxyServer = proxy()
     await new Promise(resolve => {
       const port = Number(proxyUrl.split(':')[2])
       proxyServer.listen(port, () => resolve())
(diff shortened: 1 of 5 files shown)


packages/github/__tests__/github.test.ts
```diff
@@ -11,7 +11,7 @@ describe('@actions/github', () => {
 
   beforeAll(async () => {
     // Start proxy server
-    proxyServer = proxy() as http.Server
```
thboop, line 14: This failed the linter rule `@typescript-eslint/no-unnecessary-type-assertion`

PR #418, opened 2020-04-20
Overwrite files when calling unzip from the tool-cache package

svartalf:
While Windows implementation is executed in the non-interactive mode, `unzip` can prompt user in a many different situations, one of them if any of the extracted files already exists in the destination path,
see [@actions-rs/tool-cache#5](https://github.com/actions-rs/tool-cache/issues/5) as a bug example.

This patch adds `-o` flag to the `unzip` invocation, allowing non-interactive file overwrites as it stated in the `unzip` help:

```
unzip modifiers:
  -o   Overwrite existing files without prompting.  Useful with -f.  Use with
         care.
```

### DIFF
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -295,7 +295,7 @@ async function extractZipWin(file: string, dest: string): Promise<void> {
 
 async function extractZipNix(file: string, dest: string): Promise<void> {
   const unzipPath = await io.which('unzip')
-  await exec(`"${unzipPath}"`, [file], {cwd: dest})
+  await exec(`"${unzipPath}"`, ['-o', file], {cwd: dest})
 }
 
 /**


svartalf:
Failed unittest seems to unrelated to the change:

```

FAIL packages/github/__tests__/github.test.ts (7.405s)
  @actions/github
    √ basic REST client (235ms)
    √ basic REST client with custom auth (328ms)
    √ basic REST client with proxy (236ms)
    √ basic GraphQL client (218ms)
    × basic GraphQL client with custom auth (5011ms)
    √ basic GraphQL client with proxy (188ms)

  ● @actions/github › basic GraphQL client with custom auth

    thrown: "Exceeded timeout of 5000ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

      117 |   })
      118 | 
    > 119 |   it('basic GraphQL client with custom auth', async () => {
          |   ^
      120 |     const token = getToken()
      121 |     if (!token) {
      122 |       return

      at packages/github/__tests__/github.test.ts:119:3
```

Veetaha:
@jclem could you please look at this? This bug is currently a big blocker for us... I would appreciate if you get this merged

thboop:
Ideally, we should have the same behavior across all OS's. Windows does not currently overwrite files. 

However, for a CI system I think there's value in doing this by default. @bryanmacfarlane, what do you think? 

If we decide to do so, we should make the same change in windows. This would require having to swap to the `Expand-Archive` function with `-force`, which requires >= powershell 5). There might be other subtle changes in behavior we would have to investigate.

Veetaha:
I don't think this is reasonable to let perfect be the enemy of good. Having this issue PR floating unmerged doesn't help, I'd suggest having a fix for what currently already seems ridiculous (falling into interactive mode on CI).
cc @thboop, @bryanmacfarlane

thboop:
Closing in favor of https://github.com/actions/toolkit/pull/807/files

PR #420, opened 2020-04-20
@actions/artifact 0.3.1 update

konradpabjan:
# Overview

Small updates
- Bump package version to `0.3.1` (`RELEASES.MD`, `package.json` and `package-lock.json`)
   - This is super small so it doesn't feel right to go straight to `0.4.0`
- Updates to temporary files during artifact upload
   - Explicitly call `cleanup` when we are done using temporary files
   - Cleanup code a little so that some ugly `.then()` statements aren't needed
   - Update comments and some documentation
- Remove spaces as forbidden characters when uploading artifacts
   - For some reason I added this earlier, in `v1` this is not a restriction. See the existing code [here](https://github.com/actions/runner/blob/9f78ad3b3428322131eff596e1e5b027f8db6c14/src/Runner.Plugins/Artifact/PublishArtifact.cs#L43)
   - When we push `v2-preview` into `master` for `upload-artifact` this extra restriction could break some people so we don't want to do that
   - Update tests and docs to reflect this change

# Upload compression/flow

Reminder of what happens during upload (slightly updated to mention when temp files get deleted):

image

# Testing

End to end test: https://github.com/konradpabjan/artifact-test/runs/602161502?check_suite_focus=true

### DIFF
--- a/packages/artifact/src/internal/upload-http-client.ts
+++ b/packages/artifact/src/internal/upload-http-client.ts
@@ -313,9 +313,7 @@ export class UploadHttpClient {
           // successfully uploaded so the server may report a different size for what was uploaded
           isUploadSuccessful = false
           failedChunkSizes += chunkSize
-          core.warning(
-            `Aborting upload for ${parameters.file} due to failure`
-          )
+          core.warning(`Aborting upload for ${parameters.file} due to failure`)
           abortFileUpload = true
         }
       }
@@ -325,9 +323,9 @@ export class UploadHttpClient {
 
       // only after the file upload is complete and the temporary file is deleted, return the UploadResult
       return {
-          isSuccess: isUploadSuccessful,
-          successfulUploadSize: uploadFileSize - failedChunkSizes,
-          totalSize: totalFileSize
+        isSuccess: isUploadSuccessful,
+        successfulUploadSize: uploadFileSize - failedChunkSizes,
+        totalSize: totalFileSize
       }
     }
   }
--- a/packages/artifact/src/internal/upload-http-client.ts
+++ b/packages/artifact/src/internal/upload-http-client.ts
@@ -318,10 +318,11 @@ export class UploadHttpClient {
         }
       }
 
-      // delete the temporary file that was created as part of the upload
-      await tempFile.cleanup()
+      // Delete the temporary file that was created as part of the upload. Even though this operation is async, we don't need to
+      // await this at it unnecessarily could block other uploads from starting faster. If the temp file does not get manually deleted by
+      // calling cleanup, it gets removed when the node process exits. For more info see: https://www.npmjs.com/package/tmp-promise#about
+      tempFile.cleanup()
 
-      // only after the file upload is complete and the temporary file is deleted, return the UploadResult
       return {
         isSuccess: isUploadSuccessful,
         successfulUploadSize: uploadFileSize - failedChunkSizes,
--- a/packages/artifact/src/internal/upload-http-client.ts
+++ b/packages/artifact/src/internal/upload-http-client.ts
@@ -248,91 +248,87 @@ export class UploadHttpClient {
       }
     } else {
       // the file that is being uploaded is greater than 64k in size, a temporary file gets created on disk using the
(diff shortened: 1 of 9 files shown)


packages/artifact/src/internal/upload-http-client.ts
```diff
@@ -248,91 +248,86 @@ export class UploadHttpClient {
...
-              start,
-              end,
-              uploadFileSize,
-              isGzip,
-              totalFileSize
-            )
+      // Delete the temporary file that was created as part of the upload. Even though this operation is async, we don't need to
+      // await this at it unnecessarily could block other uploads from starting faster. If the temp file does not get manually deleted by
+      // calling cleanup, it gets removed when the node process exits. For more info see: https://www.npmjs.com/package/tmp-promise#about
+      tempFile.cleanup()
```
konradpabjan, line 324: Most important line in this PR is here. We could call `await ` since it's async but I don't see a point since it just prevents other uploads from potentially starting faster

joshmgross: I think we should `await` this, especially with the recent concerns around disk size it seems like it would be better to clean up as we go. Also, I think it's just good practice to `await` any async code.

Do you see any noticeable performance differences when you *don't* await this?

yacaovsnc: `await` shouldn't block right?  The thread will just put this on the event loop and move on to the next upload?  I also think we will spend most of the time waiting on network transferring, so my guess is this won't cause measurable delays.

konradpabjan: Await will block. I did this because there actually is a measurable delay, check out this small snippet
```
import * as tmp from 'tmp-promise'
import {performance} from 'perf_hooks'

let startTime1
let startTime2

export async function test(): Promise<void> {
    await mockCleanupNoAwait()
    var endTime1 = performance.now()
    console.log("Call with no await " + (endTime1 - startTime1) + " milliseconds.")

    await mockCleanupAwait()
    var endTime2 = performance.now()
    console.log("Call with await " + (endTime2 - startTime2) + " milliseconds.")
}

async function mockCleanupNoAwait(): Promise<void>{
    const tempFile = await tmp.file()
    console.log(tempFile)
    startTime1 = performance.now()
    tempFile.cleanup()
    return
}

async function mockCleanupAwait(): Promise<void>{
    const tempFile = await tmp.file()
    console.log(tempFile)
    startTime2 = performance.now()
    await tempFile.cleanup()
    return
}

test()
```

If you run this, this is the result:
image

I tried a few variations of this and with the `await` it's always slower by a significant margin. For this package this means that the next file upload won't start until the `return` is hit. 

I'm not concerned about storage at all. I think it's better to start the next upload ASAP.

joshmgross: I don't think that timing is taking into account the time spent by node later cleaning up the temp file. I would expect the overall perf of an artifact upload to not differ much.

konradpabjan: The timing doesn't take into account when it was actually deleted, it's only measuring when the promise returns and resolves which is the point. We don't start uploading another file until we return from the previous upload. The reasoning behind not having the await is to let the cleanup finish behind the scenes without it blocking the start of the next upload. 

Regardless of the actual time implications, I added back the await.

yacaovsnc: Well, this test is a simplified version of what's really going on right?  :) In this test it does block because there is nothing else to do.  However in the real action, we have plenty stuff to do all the time, e.g. write logs, upload chunks, and other async code.  

`async / await` by definition means don't block, but do it in background right?  So if you got stuff to do, it won't increase any time, if you don't, then yeah, you will wait for it to finish.

konradpabjan: Yeah the test is a little simplified. I did a larger large scale test locally and it's pretty hard to draw any conclusions as it starts to vary significantly once I start adding `await Promise.All` and a bunch of other things that this package actually does 🤷‍♂️

PR #421, opened 2020-04-21
E2E tests for the @actions/artifact package

konradpabjan:
Super hacky workflow that will end-to-end test artifact upload and download. It felt really dirty writing some of this stuff 😂 but code coverage feels good!

It will make it very quick to test and verify the basic scenarios that the package supports. This will also give us way more confidence when accepting any user submitted pull requests.

Will only run when changes are made under `packages/artifact/**`

Normally it is not possible to upload/download artifacts from a normal shell, but I created another action to get the necessary env variables and then I set them before trying to call each of the function in `@actions/artifact`. You can find this here: https://github.com/konradpabjan/Get-Actions-Env-Variable-From-Node-Context/blob/master/src/index.ts


.github/workflows/artifact-tests.yml
```diff
@@ -0,0 +1,144 @@
...
+      uses: konradpabjan/Get-Actions-Env-Variable-From-Node-Context@4a3142ee3e
+      with:
+        EnvName: GITHUB_RUN_ID
+      id: get_run_id
+
+    # The env variables that were retrieved in a different step need to be set for each subsequent shell or upload/download will fail
+    # We're using node -e to call the functions directly available in the @actions/artifact package
+    - name: Upload artifacts using uploadArtifact()
+      run: |
+        export ACTIONS_RUNTIME_TOKEN="${{ steps.get_runtime_token.outputs.EnvValue }}"
```
konradpabjan, line 72: `${{ github.token }}` does not work unfortunately so I'm having to use this

.github/workflows/artifact-tests.yml
```diff
@@ -0,0 +1,144 @@
...
+    - name: Create files that will be uploaded
+      run: |
+        mkdir artifact-path
+        echo "hello" > artifact-path/world.txt
+        echo "Some large amount of text that has a compression ratio that is greater than 100%. If greater than 100%, gzip is used to upload the file" > artifact-path/gzip.txt
+
+    # Certain env variables are needed in order to make the necessary API calls. These env variables are not available in a standard shell but they
+    # are available for actions written in node. We're using a hack here to get the necessary env variables from a different context
+    - name: Get ACTIONS_RUNTIME_TOKEN env variable
+      uses: konradpabjan/Get-Actions-Env-Variable-From-Node-Context@4a3142ee3e
```
konradpabjan, line 51: I do not intend to ever publish this action (there is no `v1` tag or anything) so I'm just pinning it to a commit instead of branch that could regress

.github/workflows/artifact-tests.yml
```diff
@@ -0,0 +1,144 @@
...
+        - 'packages/artifact/**'
+
+jobs:
+
+  build:
+    name: Build
+
+    strategy:
+      matrix:
+        runs-on: [ubuntu-latest]
```
konradpabjan, line 17: Only testing on` ubuntu-latest` for now, I'm thinking one env should be sufficient

`macOS-latest` doesn't have `tsc` by default in the shell and with Windows some adjustments would have to be made to how env variables are set so.

.github/workflows/artifact-tests.yml
konradpabjan: I also tried permanently setting the env variables in a separate step for the machine so it's global but that didn't work so I'm having to repeat the same 3 steps multiple times

.github/workflows/artifact-tests.yml
```diff
@@ -0,0 +1,144 @@
...
+      id: get_run_id
+
+    # The env variables that were retrieved in a different step need to be set for each subsequent shell or upload/download will fail
+    # We're using node -e to call the functions directly available in the @actions/artifact package
+    - name: Upload artifacts using uploadArtifact()
+      run: |
+        export ACTIONS_RUNTIME_TOKEN="${{ steps.get_runtime_token.outputs.EnvValue }}"
+        export ACTIONS_RUNTIME_URL="${{ steps.get_runtime_url.outputs.EnvValue }}"
+        export GITHUB_RUN_ID="${{ steps.get_run_id.outputs.EnvValue }}"
+        node -e "Promise.resolve(require('./packages/artifact/lib/artifact-client').create().uploadArtifact('my-artifact-1',['artifact-path/world.txt'], '${{ github.workspace }}'))"
```
konradpabjan, line 75: I'm using `-e` for running node function in a script, for more info see:https://stackoverflow.com/questions/30782693/run-function-in-script-from-command-line-node-js

Also: https://nodejs.org/api/cli.html#cli_e_eval_script

.github/workflows/artifact-tests.yml
```diff
@@ -0,0 +1,144 @@
...
+
+    # Need root node_modules because certain npm packages like jest are configured for the entire repository and it won't be possible
+    # without these to just compile the artifacts package
+    - name: Install root npm packages
+      run: npm install
+
+    - name: Compile artifact package
+      run: |
+        npm install
+        tsc
```
joshmgross, line 39: We should use `npm run tsc` for this so that we know we're using the correct Typescript version from `node_modules` rather than the `tsc` on the path.

.github/workflows/artifact-tests.yml
```diff
@@ -0,0 +1,144 @@
...
+      run: |
+        export ACTIONS_RUNTIME_TOKEN="${{ steps.get_runtime_token.outputs.EnvValue }}"
+        export ACTIONS_RUNTIME_URL="${{ steps.get_runtime_url.outputs.EnvValue }}"
+        export GITHUB_RUN_ID="${{ steps.get_run_id.outputs.EnvValue }}"
+        node -e "Promise.resolve(require('./packages/artifact/lib/artifact-client').create().uploadArtifact('my-artifact-1',['artifact-path/world.txt'], '${{ github.workspace }}'))"
+        node -e "Promise.resolve(require('./packages/artifact/lib/artifact-client').create().uploadArtifact('my-artifact-2',['artifact-path/gzip.txt'], '${{ github.workspace }}'))"
+
+    - name: Download artifacts using downloadArtifact()
+      run: |
+        export ACTIONS_RUNTIME_TOKEN="${{ steps.get_runtime_token.outputs.EnvValue }}"
```
joshmgross, line 80: You should be able to use step level `env` values rather than manually calling `export`

.github/workflows/artifact-tests.yml
joshmgross: Can we combine these into one call? And do we need an external action to do this?

Seems like a quick node script that we run as a local action would be more concise and checked in to this repo:


```js
# index.js
console.log(`::set-output name=ACTIONS_RUNTIME_URL::${process.env.ACTIONS_RUNTIME_URL}`)
console.log(`::set-output name= ACTIONS_RUNTIME_TOKEN::${process.env. ACTIONS_RUNTIME_TOKEN}`)
...
```

```yaml
# action.yml
runs:
  using: 'node12'
  main: 'index.js'
```

.github/workflows/artifact-tests.yml
joshmgross: You should be able to use https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable to do this

.github/workflows/artifact-tests.yml
```diff
@@ -0,0 +1,144 @@
...
+        export ACTIONS_RUNTIME_URL="${{ steps.get_runtime_url.outputs.EnvValue }}"
+        export GITHUB_RUN_ID="${{ steps.get_run_id.outputs.EnvValue }}"
+        mkdir multi-artifact-directory
+        node -e "Promise.resolve(require('./packages/artifact/lib/artifact-client').create().downloadAllArtifacts('multi-artifact-directory'))"
+
+    - name: Verify downloadAllArtifacts() 
+      run: |
+        $file1 = "multi-artifact-directory/my-artifact-1/artifact-path/world.txt"
```
joshmgross, line 123: We should include these as separate script files in the repo rather than having the entire thing in the workflow file

.github/workflows/artifact-tests.yml
joshmgross: We should test on all environments, use `tsc` from your `node_modules` that you install and there are helpers in `core` and workflow commands to set environment variables.

.github/workflows/artifact-tests.yml
konradpabjan: What benefit does putting the scripts in separate files have here? Is the entire script in a single workflow always a bad thing?

If the scripts start to get really really long (or if they need to be reused by other workflows) then I can see it being desired, but in this scenario it would just make it hard to maintain without a real benefit. Having worked with some of our larger canaries that used a lot of templates and script references, it gets really annoying to have to dig through different files. Here if a run fails, you can just look at the `WorkFlow file` tab and you have all your code.

Also, think about the artifact names, contents and paths that we are checking here (what we specify in other steps). We would have to cross-reference the file paths, contents and the artifact names in the script files and the YAML. Considering the simplicity of what we're trying to do, I absolutely think that separate script files are not warranted.

.github/workflows/artifact-tests.yml
konradpabjan: Forgot about this, it does the trick! I tried a bunch of things in bash and powershell to globally set it but it would never work for some reason.

.github/workflows/artifact-tests.yml
joshmgross: The large script blocks make it hard to parse the workflow, and you're duplicating a lot of the checking logic.

For example, looking at checkout's E2E tests, it's very easy to read the workflow and understand what's being validated: https://github.com/actions/checkout/blob/master/.github/workflows/test.yml#L175

The scripts themselves are very short (https://github.com/actions/checkout/blob/master/__test__/verify-basic.sh) but the actual script content isn't necessary to understand what's happening in the workflow.

>We would have to cross-reference the file paths, contents and the artifact names in the script files and the YAML. 

These could all be passed in as environment variables or command-line arguments instead of being hardcoded in the scripts

.github/workflows/artifact-tests.yml
konradpabjan: I aggregated everything into a local action, cleans up things a little more.

Originally I was using a separate action so that it could be reused, but I don't expect this ever to be needed outside of this specific testing scenario 😂

.github/workflows/artifact-tests.yml
konradpabjan: Took a bit to figure out why mac wasn't working. I had to add `typescript` to the dev dependencies for the artifact package because of the whole shell issue.

.github/workflows/artifact-tests.yml
konradpabjan: Took a wee bit of work, but it's done

packages/artifact/package-lock.json
```diff
@@ -4,6 +4,11 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@actions/core": {
```
thboop, line 7: @konradpabjan lerna doesn't typically check in sibling projects. This will get removed when other developers run commands, and they will have to manually update or fail the build. for example : https://github.com/actions/toolkit/pull/428/files

Can we run `lerna bootstrap` rather then npm ci here: https://github.com/actions/toolkit/pull/421/files#diff-c0a0a8d36f11bf33d1a7ea8e7cda953dR36

.github/workflows/artifact-tests.yml
```diff
@@ -0,0 +1,79 @@
...
+        echo "::set-env name=gzip-artifact-content::Some large amount of text that has a compression ratio that is greater than 100%. If greater than 100%, gzip is used to upload the file"
+
+    - name: Create files that will be uploaded
+      run: |
+        mkdir artifact-path 
+        echo ${{ env.non-gzip-artifact-content }} > artifact-path/world.txt
+        echo ${{ env.gzip-artifact-content }} > artifact-path/gzip.txt
+
+    # We're using node -e to call the functions directly available in the @actions/artifact package
+    - name: Upload artifacts using uploadArtifact()
```
hashtagchris, line 52: It would be good to test a multi-file artifact upload. say the artifact-path dir itself.

.github/workflows/artifact-tests.yml
```diff
@@ -0,0 +1,79 @@
...
+    - name: Set artifact file contents
+      run: |
+        echo "::set-env name=non-gzip-artifact-content::hello"
+        echo "::set-env name=gzip-artifact-content::Some large amount of text that has a compression ratio that is greater than 100%. If greater than 100%, gzip is used to upload the file"
+
+    - name: Create files that will be uploaded
+      run: |
+        mkdir artifact-path 
+        echo ${{ env.non-gzip-artifact-content }} > artifact-path/world.txt
+        echo ${{ env.gzip-artifact-content }} > artifact-path/gzip.txt
```
hashtagchris, line 49: I recommend:

1. Using a file/file comparison instead of a file/string comparison for your verification. You could pass these `artifact-path` file paths to `test-artifact-file.sh`, or possibly `cmp` directly, below.

2. Skipping the use of environment variables here - just use the following:

```
        echo "hello" > artifact-path/world.txt
        echo "Some large amount of text that has a compression ratio that is greater than 100%. If greater than 100%, gzip is used to upload the file" > artifact-path/gzip.txt
```

packages/artifact/package-lock.json
konradpabjan: Thanks for the heads up @thboop !

I was wondering why sometimes locally `package-lock.json` would change but then when running here it wouldn't. That explains it. I'll work on addressing this in a subsequent PR

.github/workflows/artifact-tests.yml
konradpabjan: The file comparison is a good idea, same with the multi-file artifact. I'll work on that in a subsequent PR. 

Env variables were used to cut down on copy and pasting since later on they're used to check the contents of the downloaded files

hashtagchris: If we switch to file comparisons, then we'll no longer need the environment variables. 🎉

PR #422, opened 2020-04-23
Add architecture filter to findFromManifest

maxim-lobanov:
For some actions, like `setup-python` we want to install tool with different architecture.
For example, Python x86 can be installed to Windows Server 2019 so it makes sense to provide ability to override architecture

### DIFF
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -526,13 +526,15 @@ export async function getManifestFromRepo(
 export async function findFromManifest(
   versionSpec: string,
   stable: boolean,
-  manifest: IToolRelease[]
+  manifest: IToolRelease[],
+  archFilter: string = os.arch()
 ): Promise<IToolRelease | undefined> {
   // wrap the internal impl
   const match: mm.IToolRelease | undefined = await mm._findMatch(
     versionSpec,
     stable,
-    manifest
+    manifest,
+    archFilter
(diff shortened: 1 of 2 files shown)


packages/tool-cache/src/tool-cache.ts
```diff
@@ -526,13 +526,15 @@ export async function getManifestFromRepo(
 export async function findFromManifest(
   versionSpec: string,
   stable: boolean,
-  manifest: IToolRelease[]
+  manifest: IToolRelease[],
+  archFilter: string = os.arch()
```
maxim-lobanov, line 530: I implemented the easiest solution and just added one more parameter. But may be it makes sense to use some kind of object here, like :
```
interface IManifestMatchFilter {
  version: string;
  stable:? boolean;
  archFilter:? string;
  platFilter:? string;
}

export async function findFromManifest(
  manifest: IToolRelease[],
  filter: IManifestMatchFilter;
}
```
It will allow us to extend this interface in future if we need. If it sounds good to you, I can rework this pull-request.

If you prefer to keep current parameters, I would suggest to change order of parameters:
```
manifest,
versionSpec,
stable,
archFilter
```
It will be more logical

bryanmacfarlane: I think another param is fine for now.  If we ever do a v2 where we break compat, we can take a pass and put all options in options objs.

PR #436, opened 2020-04-29
Add secret option to getInput

clarkbw:
This is a convenience method but I think it has value to mark the inputs which are secrets or passwords at the point where they are captured such that the `setSecret` function call isn't forgotten later.

Here's an example of usage:
```javascript
const username = core.getInput('username', { required: true });
const password = core.getInput('password', { required: true, secret: true });
```

And here's how you currently achieve this:
```javascript
const username = core.getInput('username', { required: true });
const password = core.getInput('password', { required: true });
core.setSecret(password);
```

### DIFF
--- a/packages/core/__tests__/core.test.ts
+++ b/packages/core/__tests__/core.test.ts
@@ -109,6 +109,11 @@ describe('@actions/core', () => {
     )
   })
 
+  it('getInput masks secret inputs', () => {
+    expect(core.getInput('My InPuT', { secret: true })).toBe('val')
+    assertWriteCalls([`::add-mask::val${os.EOL}`])
+  })
+
   it('setOutput produces the correct command', () => {
     core.setOutput('some output', 'some value')
     assertWriteCalls([`::set-output name=some output::some value${os.EOL}`])
--- a/packages/core/__tests__/core.test.ts
+++ b/packages/core/__tests__/core.test.ts
@@ -110,7 +110,7 @@ describe('@actions/core', () => {
   })
 
   it('getInput masks secret inputs', () => {
-    expect(core.getInput('My InPuT', { secret: true })).toBe('val')
+    expect(core.getInput('My InPuT', {secret: true})).toBe('val')
     assertWriteCalls([`::add-mask::val${os.EOL}`])
   })
 
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -9,6 +9,8 @@ import * as path from 'path'
 export interface InputOptions {
   /** Optional. Whether the input is required. If required and not present, will throw. Defaults to false */
   required?: boolean
+  /** Optional. If the input is a secret and should be auto-masked from logs */
+  secret?: boolean
 }
 
 /**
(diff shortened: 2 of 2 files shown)


thboop:
Hey @clarkbw , thank you for the contribution!

I'd like to understand the use case a little better. Generally I would expect the `password` input to get [set as a secret](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#creating-encrypted-secrets) in the repository. 

For secret values that are getting set as the output of an action , and later consumed by another action, I would expect that the original action would set any sensitive materials they are outputting as a secret.

Could you elaborate on the use case more, or provide an example workflow yaml file?

clarkbw:
> I'd like to understand the use case a little more. Generally I would expect the `password` input to get [set as a secret](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#creating-encrypted-secrets) in the repository.

Perhaps I didn't see this in the docs but if I'm understanding you correctly any value set in the secret store is automatically masked within Actions and there's no need for an Action to call `setSecret` on it?


My use case is that given an Action which requires a password to log into some kind of account like this one that logs you into a Docker registry:

```javascript
const username = core.getInput('username', { required: true });
const password = core.getInput('password', { required: true, secret: true });
const registry = core.getInput('registry');

const options = {
  input: Buffer.from(password)
};
await exec.exec(
    'docker',
    ['login', registry, '-u', username, '--password-stdin'],
    options
  );
```

A workflow file would pass values like the following:

```yml
name: Docker Image CI

on:
  push:

jobs:

  build:

    runs-on: ubuntu-latest

    steps:
    - uses: actions/checkout@v2
    - uses: clarkbw/setup-docker@master
      with:
        username: $GITHUB_ACTOR
        password: ${{ secrets.REGISTRY_TOKEN }}
    - name: Build the Docker image
      run: docker build . --tag docker.pkg.github.com/github/container-registry-playground/container-registry-playground:latest
    - name: Publish the Docker image
      run: docker push docker.pkg.github.com/github/container-registry-playground/container-registry-playground:latest
```

I want to make sure the password being consumed is hidden no matter what it is. And currently that means making a separate `setSecret` call on the password entry to ensure it gets masked.  Perhaps `masked` would be a more descriptive option.

I haven't heard other use cases for masking things like usernames yet.

thboop:
> Perhaps I didn't see this in the docs but if I'm understanding you correctly any value set in the secret store is automatically masked within Actions and there's no need for an Action to call setSecret on it?

That is correct. There's no real harm in doing so, but in your example, calling `setSecret` on `password` is essentially a no-op. It is already registered, so it will get ignored in the [actions/runner](https://github.com/actions/runner).

`setSecret` is really intended for the use case where your action generates a secret value, and uses that secret or sets it as an output for another action to use.

We have a bit [here](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#about-encrypted-secrets) about automatically redacting secrets printed to the log, but I would welcome your feedback on how to improve the docs!

clarkbw:
Ah cool.  Ok, no need for this then. Thanks!



> `setSecret` is really intended for the use case where your action generates a secret value, and uses that secret or sets it as an output for another action/step to use.

I'd say this snippet inline with the `setSecret` function documentation would be 💯

PR #441, opened 2020-05-01
artifact header cleanup

konradpabjan:
# Overview

~- Update `npm` packages for `@actions/artifact` to their latest versions~
  ~- I want to frequently do this so that we don't get any security alerts~
  ~- [ncu](https://www.npmjs.com/package/npm-check-updates) makes this super easy and painless~
- Clarifications around headers and request options
   - replace `requestOptions` in most places with `headers`
   - This has no impact on any logic or control flow, this is purely for clarification and cleanup
   - [requestOptions](https://github.com/actions/http-client/blob/5f61f72a73224ee5541c6961f864e09e9f4cf6d7/interfaces.ts#L79) in the http-client are different from [headers](https://github.com/actions/http-client/blob/5f61f72a73224ee5541c6961f864e09e9f4cf6d7/interfaces.ts#L3)

# Testing

- We have E2E CI, and it's looking good 😀
- Extra sanity check with some larger artifact upload and downloads: https://github.com/konradpabjan/artifact-test/runs/636175436?check_suite_focus=true

### DIFF
--- a/packages/artifact/package.json
+++ b/packages/artifact/package.json
@@ -36,10 +36,10 @@
     "url": "https://github.com/actions/toolkit/issues"
   },
   "dependencies": {
-    "@actions/core": "^1.2.1",
-    "@actions/http-client": "^1.0.7",
-    "@types/tmp": "^0.1.0",
-    "tmp": "^0.1.0",
+    "@actions/core": "^1.2.4",
+    "@actions/http-client": "^1.0.8",
+    "@types/tmp": "^0.2.0",
+    "tmp": "^0.2.1",
     "tmp-promise": "^2.0.2"
   },
   "devDependencies": {
--- a/packages/artifact/package.json
+++ b/packages/artifact/package.json
@@ -36,10 +36,10 @@
     "url": "https://github.com/actions/toolkit/issues"
   },
   "dependencies": {
-    "@actions/core": "^1.2.4",
-    "@actions/http-client": "^1.0.8",
-    "@types/tmp": "^0.2.0",
-    "tmp": "^0.2.1",
+    "@actions/core": "^1.2.1",
+    "@actions/http-client": "^1.0.7",
+    "@types/tmp": "^0.1.0",
+    "tmp": "^0.1.0",
     "tmp-promise": "^2.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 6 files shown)


packages/artifact/package.json
```diff
@@ -36,10 +36,10 @@
...
   },
   "dependencies": {
-    "@actions/core": "^1.2.1",
-    "@actions/http-client": "^1.0.7",
-    "@types/tmp": "^0.1.0",
-    "tmp": "^0.1.0",
+    "@actions/core": "^1.2.4",
+    "@actions/http-client": "^1.0.8",
+    "@types/tmp": "^0.2.0",
+    "tmp": "^0.2.1",
```
joshmgross, line 42: What's the benefit of doing this?

If there's a security issue with any of these, the alert should be enough of a notice. It feels like we're just introducing extra code changes proactively without needing the updated packages.

konradpabjan: It's recommended to regularly update them. From: https://docs.npmjs.com/updating-packages-downloaded-from-the-registry

> We recommend regularly updating the local packages your project depends on to improve your code as improvements to its dependencies are made.

Possibly fewer security alerts is just one benefit.  Updating these packages doesn't break compat since there in ways that packages define the min and max versions that they can use, so by doing this we are regularly pulling in improvements. 

---

Aside as to why i prefer [ncu over npm update](https://stackoverflow.com/questions/57069622/command-npm-update-vs-package-npm-check-updates)

konradpabjan: Reverted after some offline discussions, only thing left in this PR is the clarifications around headers.

packages/artifact/src/internal/utils.ts
```diff
@@ -135,7 +133,7 @@ export function getContentRange(
  * @param {string} acceptType the type of content that we can accept
  * @returns appropriate request options to make a specific http call during artifact download
```
joshmgross, line 134:
```suggestion
 * @returns appropriate headers to make a specific http call during artifact download
```

packages/artifact/src/internal/utils.ts
```diff
@@ -174,7 +172,7 @@ export function getDownloadRequestOptions(
  * @param {string} contentRange the range of the content that is being uploaded
  * @returns appropriate request options to make a specific http call during artifact upload
```
joshmgross, line 173:
```suggestion
 * @returns appropriate headers to make a specific http call during artifact upload
```

PR #444, opened 2020-05-03
Simplify mkdirP implementation

LinusU:
less code = fewer bugs ☺️ 

The `recursive` option have been available since Node.js 10.12.0 ([ref](https://nodejs.org/api/fs.html#fs_fs_mkdir_path_options_callback)), and since GitHub Actions runs on Node.js 12.x this should always be available.

### DIFF
--- a/packages/io/__tests__/io.test.ts
+++ b/packages/io/__tests__/io.test.ts
@@ -3,7 +3,6 @@ import {promises as fs} from 'fs'
 import * as os from 'os'
 import * as path from 'path'
 import * as io from '../src/io'
-import * as ioUtil from '../src/io-util'
 
 describe('cp', () => {
   it('copies file with no flags', async () => {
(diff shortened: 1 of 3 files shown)


LinusU:
(force pushed a lint fix)

mgd966:
في أحد، 3 مايو، 2020 في 2:48 م، كتب Linus Unnebäck <notifications@github.com

LinusU:
ping @damccorm, would you be able to review this? ☺️

damccorm:
@hross could you take a look?

bryanmacfarlane:
I would also like to point out that while more code, the current code has been working in azure pipelines and github for 5 years+.  changing the code can also introduce bugs and subtle behavioral differences ;).  We should confirm 100% that the error case behaviors are the exact same.  e.g. see the note about optimistic behaviors of the existing function.  If needed, we should introduce more tests to prove the behaviors are the same in all cases.

bryanmacfarlane:
@thboop for thoughts.

LinusU:
> I would also like to point out that while more code, the current code has been working in azure pipelines and github for 5 years+. changing the code can also introduce bugs and subtle behavioral differences ;)

Fair point, although the built-in one have been shipped with every copy of Node.js since 2018, so it has probably been quite battle tested itself ☺️ 

> e.g. see the note about optimistic behaviors of the existing function

Node's code also tries to mkdir the full path first, and then traverses up one step at a time:

https://github.com/nodejs/node/blob/d3a8a23089af06bb047bf9bad7531fbfc70f6314/src/node_file.cc#L1304-L1339

> If needed, we should introduce more tests to prove the behaviors are the same in all cases.

I'd be happy to help with this if there is any specific behaviours that you think needs testing 👨‍💻

thboop (approved):
LGTM thanks for the contribution! @LinusU

PR #448, opened 2020-05-06
Initial commit to create @actions/cache package

aiqiaoy:
Initial commit to create the @actions/cache package. ADR is [here](https://github.com/github/pe-actions-service/blob/master/docs/adrs/0516-cache-action-package.md). 

The original code can be found under the cache repo [here](https://github.com/actions/cache/tree/master/src). Differences between cache actions and the node package are: 
* Keys, paths and concurrency settings are input and environment variables in cache action. Here they are function parameters. 
* For save and restore failures, cache action log warnings while the node package throw errors and allow consumers to decide what to do.

Last commit included in this PR is: bac1a40c8194ae3194f8e2f134c086b56d17eaf4

Addresses: 
https://github.com/github/c2c-actions-service/issues/533
https://github.com/actions/cache/issues/55


packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -0,0 +1,339 @@
...
+
+  return new HttpClient(
+    'actions/cache',
+    [bearerCredentialHandler],
+    getRequestOptions()
+  )
+}
+
+export function getCacheVersion(
+  inputPath: string,
```
aiqiaoy, line 87: `inputPath` here is the `path` input variable for our first party cache action. We use the input string directly to compute version [here](https://github.com/actions/cache/blob/ce9276c90e52c28f187062530b2d7a43fd7973a5/src/cacheHttpClient.ts#L89). 

This is not ideal for the cache package. Action authors might want to implement their cache action differently and use different variables. Perhaps we should pass paths as an array to `getCacheVersion` and bump up the version salt.

packages/cache/src/cache.ts
```diff
@@ -0,0 +1,169 @@
...
+}
+
+/**
+ * Saves a file with the specified key
+ *
+ * @param path a string representing files to be cached
+ * @param key an explicit key for restoring the cache
+ * @returns number returns cacheId if the cache was saved successfully, otherwise return -1
+ */
+export async function saveCache(path: string, key: string): Promise<number> {
```
joshmgross, line 113: We should support multiple paths for this input, we don't have to rely on splitting `path` by a newline in the package itself.
```suggestion
export async function saveCache(paths: string[], key: string): Promise<number> {
```

packages/cache/src/cache.ts
```diff
@@ -0,0 +1,169 @@
...
+      )
+      return -1
+    }
+
+    core.debug(`Saving Cache (ID: ${cacheId})`)
+    await cacheHttpClient.saveCache(cacheId, archivePath)
+
+    return cacheId
+  } catch (error) {
+    utils.logWarning(error.message)
```
joshmgross, line 166: We should either make this configurable or always throw an error. The package consumer can control whether they want to catch the error or allow it to fail the step.

packages/cache/src/internal/cacheHttpClient.ts
joshmgross: I think it would make sense to allow action authors to optionally specify a `cacheVersion` along with their key and we could use this function to generate one if not specified 

> I would like to change this to pass paths as an array to getCacheVersion but it would break existing caches.

There are no consumers of this package yet, so we can make that breaking change. It will be breaking when we start consuming it in `actions/cache` but that's acceptable since this will go along with the breaking v2 major version.

aiqiaoy:
I'm still investigating the test failures but the PR is ready for review.

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -0,0 +1,339 @@
...
+  archivePath: string
+): Promise<void> {
+  // Upload Chunks
+  const fileSize = fs.statSync(archivePath).size
+  const resourceUrl = getCacheApiUrl(`caches/${cacheId.toString()}`)
+  const fd = fs.openSync(archivePath, 'r')
+
+  const concurrency = parseEnvNumber('CACHE_UPLOAD_CONCURRENCY') ?? 4 // # of HTTP requests in parallel
+  const MAX_CHUNK_SIZE =
+    parseEnvNumber('CACHE_UPLOAD_CHUNK_SIZE') ?? 32 * 1024 * 1024 // 32 MB Chunks
```
aiqiaoy, line 275: Do we want to add upload concurrency and chunk size as optional parameters?

joshmgross: Yeah let's make these optional inputs instead of environment variables.

We should do something similar to [`globOptions`](https://github.com/actions/toolkit/blob/master/packages/glob/src/glob.ts#L14) so that we don't have to change the function signature everytime we add a new option.

README.md
```diff
@@ -81,6 +83,15 @@ $ npm install @actions/artifact --save
 ```
 <br/>
 
+:dart: [@actions/cache](packages/cache) 
+
+Provides functions to cache dependencies and build outputs to improve workflow execution time. Read more [here](packages/cache)
+
+```bash
+$ npm install @actions/cache --save
```
thboop, line 91: -- save is deprecated and done by default, we can remove it!

.github/workflows/cache-tests.yml
```diff
@@ -0,0 +1,61 @@
+name: cache-unit-tests
+on: push
```
konradpabjan, line 2: I would recommend path filters here. Certain things like `.md` files can be ignored: https://github.com/actions/toolkit/blob/83dd3ef0f1e5bc93c5ab7072e1edf1715a01ba9d/.github/workflows/unit-tests.yml#L7

aiqiaoy:
@joshmgross @dhadka @thboop Finally got all tests to pass. PR is ready for review 😄

packages/cache/__tests__/cacheUtils.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+test('unlinkFile unlinks file', async () => {
+  const testDirectory = await fs.mkdtemp('unlinkFileTest')
+  const testFile = path.join(testDirectory, 'test.txt')
+  await fs.writeFile(testFile, 'hello world')
+
+  await cacheUtils.unlinkFile(testFile)
+
+  // This should throw as testFile should not exist
+  await expect(fs.stat(testFile)).rejects.toThrow()
```
konradpabjan, line 34: I would add a test to check if it doesn't throw before the unlink

packages/cache/__tests__/cacheUtils.test.ts
```diff
@@ -0,0 +1,37 @@
...
+import * as cacheUtils from '../src/internal/cacheUtils'
+
+jest.mock('@actions/core')
+jest.mock('os')
+
+function getTempDir(): string {
+  return path.join(__dirname, '_temp', 'cacheUtils')
+}
+
+afterAll(async () => {
```
konradpabjan, line 13: Are these needed? I don't think there is anything created in the TempDir with these 2 tests. With `fs.mkdtemp` that is used below, I think it uses some other directory

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+import * as path from 'path'
+import * as util from 'util'
+import {v4 as uuidV4} from 'uuid'
+import {CacheFilename, CompressionMethod} from './constants'
+
+// From https://github.com/actions/toolkit/blob/master/packages/tool-cache/src/tool-cache.ts#L23
+export async function createTempDirectory(): Promise<string> {
+  const IS_WINDOWS = process.platform === 'win32'
+
+  let tempDirectory: string = process.env['RUNNER_TEMP'] || ''
```
konradpabjan, line 15: If the env variable isn't set, is the temp directory expected to be created in the current working directory? There is an [extra check](https://github.com/actions/toolkit/blob/36e732155e521874637327a845b689ddc7a20645/packages/tool-cache/src/tool-cache.ts#L557) in the `tool-cache` package

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+    : CompressionMethod.Gzip
+}
+
+export function getCacheFileName(compressionMethod: CompressionMethod): string {
+  return compressionMethod === CompressionMethod.Zstd
+    ? CacheFilename.Zstd
+    : CacheFilename.Gzip
+}
+
+export async function useGnuTar(): Promise<boolean> {
```
konradpabjan, line 96: Wouldn't a better name be `isGnuTarSupported` or something similar since this is checking if `gnu tar` is supported?

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+  } catch (err) {
+    core.debug(err.message)
+  }
+
+  versionOutput = versionOutput.trim()
+  core.debug(versionOutput)
+  return versionOutput
+}
+
+export async function getCompressionMethod(): Promise<CompressionMethod> {
```
konradpabjan, line 83: I would add a comment or something about the default behavior. If zstd is not supported then we default to gzip (I like documentation and comments 😀).

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -0,0 +1,411 @@
...
+    const expectedLength = parseInt(contentLengthHeader)
+    const actualLength = utils.getArchiveFileSize(archivePath)
+
+    if (actualLength !== expectedLength) {
+      throw new Error(
+        `Incomplete download. Expected file size: ${expectedLength}, actual file size: ${actualLength}`
+      )
+    }
+  } else {
+    core.debug('Unable to validate download, no Content-Length header')
```
konradpabjan, line 253: Just curious, what are the implications of not properly validating the download? If this is serious then maybe `core.warning` should be used?

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -0,0 +1,411 @@
...
+  const httpClient = createHttpClient()
+  const version = getCacheVersion(paths, options?.compressionMethod)
+  const resource = `cache?keys=${encodeURIComponent(
+    keys.join(',')
+  )}&version=${version}`
+
+  const response = await retryTypedResponse('getCacheEntry', async () =>
+    httpClient.getJson<ArtifactCacheEntry>(getCacheApiUrl(resource))
+  )
+  if (response.statusCode === 204) {
```
konradpabjan, line 194: Not needed for this PR, but 204 could be added to the [HTTP Codes](https://github.com/actions/http-client/blob/5f61f72a73224ee5541c6961f864e09e9f4cf6d7/index.ts#L8) in our http-client

packages/cache/src/cache.ts
```diff
@@ -0,0 +1,175 @@
+import * as core from '@actions/core'
+import * as path from 'path'
+import * as utils from './internal/cacheUtils'
+import * as cacheHttpClient from './internal/cacheHttpClient'
+import {createTar, extractTar} from './internal/tar'
+import {UploadOptions} from './options'
+
+export class ValidationError extends Error {
```
konradpabjan, line 8: This is a pretty cool pattern! 👍

packages/cache/src/cache.ts
```diff
@@ -0,0 +1,175 @@
...
+  paths: string[],
+  key: string,
+  options?: UploadOptions
+): Promise<number> {
+  checkPaths(paths)
+  checkKey(key)
+
+  const compressionMethod = await utils.getCompressionMethod()
+
+  core.debug('Reserving Cache')
```
konradpabjan, line 135: Just an observation, `core.debug` is used pretty frequently, but I think there should be some uses of `core.info` so users see more logs and information.

Most users won't have step-debugging turned on and I've noticed that a lot of our first party actions don't have the greatest output/logging.

konradpabjan:
Side note: I really like documentation and visuals 🙂 For the artifacts package I added some charts to help make it easier for new users to understand what is going on (and maintainers when they need a refresh in the future). You can see it here: https://github.com/actions/toolkit/blob/master/packages/artifact/docs/implementation-details.md#uploadcompression-flow

Absolutely not needed for this PR, but I think it would be pretty nice to have something similar for this package. Even some basic charts about how the decision is made between `zstd` and `gzip` or how the retry logic is handled. Even a few things should make it easier for open-source contributors to understand things quicker.

README.md
```diff
@@ -82,6 +84,15 @@ $ npm install @actions/artifact --save
 ```
 <br/>
 
+:dart: [@actions/cache](packages/cache) 
+
+Provides functions to cache dependencies and build outputs to improve workflow execution time. Read more [here](packages/cache)
+
+```bash
+$ npm install @actions/cache
```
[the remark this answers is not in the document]
joshmgross: `--save` is deprecated, we should remove it from all of these examples in a separate PR. See https://github.com/actions/toolkit/pull/448#discussion_r422217165

packages/cache/src/cache.ts
```diff
@@ -0,0 +1,175 @@
...
+  }
+}
+
+/**
+ * Restores cache from keys
+ *
+ * @param paths a list of file paths to restore from the cache
+ * @param primaryKey an explicit key for restoring the cache
+ * @param restoreKeys an optional ordered list of keys to use for restoring the cache if no cache hit occurred for key
+ * @returns string returns the key for the cache hit, otherwise return undefined
```
joshmgross, line 50:
```suggestion
 * @returns string returns the key for the cache hit, otherwise returns undefined
```

packages/cache/src/internal/cacheUtils.ts
joshmgross: See below, if `tempDIrectory` is empty we create our own temp directory. Note this is an edge case, as `RUNNER_TEMP` should always be defined.

packages/cache/src/internal/cacheUtils.ts
joshmgross: I think it's a personal preference, `useGnuTar` seems fine here to me

packages/cache/src/options.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+export interface UploadOptions {
+  /**
+   * Number of parallel cache upload
+   *
+   * @default 4
+   */
+  uploadConcurrency?: number
+  /**
+   * Maximum chunk size for cache upload
```
joshmgross, line 12:
```suggestion
   * Maximum chunk size in bytes for cache upload
```

Let's clarify bytes here, it might be confusing if users think the default is `32` instead of 32 * 1024 * 1024

packages/cache/src/cache.ts
joshmgross: Package users/Action Authors might not want that verbose logging from a package they consume. I'd say leave as-is and if we get feedback we can add an option to the package to add verbose logging rather than verbose by default

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+import {v4 as uuidV4} from 'uuid'
+import {CacheFilename, CompressionMethod} from './constants'
+
+// From https://github.com/actions/toolkit/blob/master/packages/tool-cache/src/tool-cache.ts#L23
+export async function createTempDirectory(): Promise<string> {
+  const IS_WINDOWS = process.platform === 'win32'
+
+  let tempDirectory: string = process.env['RUNNER_TEMP'] || ''
+
+  if (!tempDirectory) {
```
thboop, line 17: should we create a `@actions/core` function for this, we use it in other places? We can create an issue for it and solve it sometime in the future if you want :) https://github.com/actions/toolkit/blob/57d20b4db494c25af8d2f3d9323650044610e531/packages/tool-cache/src/tool-cache.ts#L555

Is there a reason we expect `RUNNER_TEMP` to not be defined? IIRC, it should always be defined, it may be okay to error if it isn't.

packages/cache/README.md
```diff
@@ -0,0 +1,41 @@
...
+const cache = require('@actions/cache');
+const paths = [
+    'node_modules',
+    'packages/*/node_modules/'
+]
+const key = 'npm-foobar-d5ea0750'
+const cacheId = await cache.saveCache(paths, key)
+```
+
+## Additional Documentation
```
joshmgross, line 39: Once `actions/cache` starts to consume this package, we can update this doc to point to it as an example implementation.

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -0,0 +1,411 @@
...
+  return retryableStatusCodes.includes(statusCode)
+}
+
+function getCacheApiUrl(resource: string): string {
+  // Ideally we just use ACTIONS_CACHE_URL
+  const baseUrl: string = (
+    process.env['ACTIONS_CACHE_URL'] ||
+    process.env['ACTIONS_RUNTIME_URL'] ||
+    ''
+  ).replace('pipelines', 'artifactcache')
```
thboop, line 59: What is this replace for?

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -0,0 +1,411 @@
...
+    HttpCodes.GatewayTimeout
+  ]
+  return retryableStatusCodes.includes(statusCode)
+}
+
+function getCacheApiUrl(resource: string): string {
+  // Ideally we just use ACTIONS_CACHE_URL
+  const baseUrl: string = (
+    process.env['ACTIONS_CACHE_URL'] ||
+    process.env['ACTIONS_RUNTIME_URL'] ||
```
thboop, line 57: Do we need a backup? Or is this for back-compat we can remove now?

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -0,0 +1,411 @@
...
+export function getCacheVersion(
+  paths: string[],
+  compressionMethod?: CompressionMethod
+): string {
+  const components = paths.concat(
+    compressionMethod === CompressionMethod.Zstd ? [compressionMethod] : []
+  )
+
+  // Add salt to cache version to support breaking changes in cache entry
+  components.push(versionSalt)
```
thboop, line 103: 👍

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -0,0 +1,411 @@
...
+}
+
+export async function saveCache(
+  cacheId: number,
+  archivePath: string,
+  options?: UploadOptions
+): Promise<void> {
+  const httpClient = createHttpClient()
+
+  core.debug('Upload cache')
```
thboop, line 397: Great debugging all around 🏅

packages/cache/src/internal/tar.ts
```diff
@@ -0,0 +1,86 @@
...
+  archivePath: string,
+  compressionMethod: CompressionMethod
+): Promise<void> {
+  // Create directory to extract tar into
+  const workingDirectory = getWorkingDirectory()
+  await io.mkdirP(workingDirectory)
+  // --d: Decompress.
+  // --long=#: Enables long distance matching with # bits. Maximum is 30 (1GB) on 32-bit OS and 31 (2GB) on 64-bit.
+  // Using 30 here because we also support 32-bit self-hosted runners.
+  const args = [
```
thboop, line 44: Should we change the logging of tar to be more verbose if core.isDebug is set?

packages/cache/src/internal/tar.ts
```diff
@@ -0,0 +1,86 @@
...
+    path.join(archiveFolder, manifestFilename),
+    sourceDirectories.join('\n')
+  )
+  // -T#: Compress using # working thread. If # is 0, attempt to detect and use the number of physical CPU cores.
+  // --long=#: Enables long distance matching with # bits. Maximum is 30 (1GB) on 32-bit OS and 31 (2GB) on 64-bit.
+  // Using 30 here because we also support 32-bit self-hosted runners.
+  const workingDirectory = getWorkingDirectory()
+  const args = [
+    ...(compressionMethod === CompressionMethod.Zstd
+      ? ['--use-compress-program', 'zstd -T0 --long=30']
```
thboop, line 75: Should we change the logging of tar to be more verbose if core.isDebug is set?

.github/workflows/cache-tests.yml
```diff
@@ -0,0 +1,69 @@
+name: cache-unit-tests
```
thboop, line 1: Awesome 🎖️

packages/cache/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+See ["Caching dependencies to speed up workflows"](https://help.github.com/github/automating-your-workflow-with-github-actions/caching-dependencies-to-speed-up-workflows).
```
thboop, line 41: I would lead with a link to this. Gives a great explanation of caching and how keys work., then go into what functions this exposes.

thboop, line 41: We may want to consider explicitly calling out usage limits as well:
https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy

README.md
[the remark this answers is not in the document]
konradpabjan: Gotcha 👍 Didn't see that since it was marked as resolved

packages/cache/src/internal/tar.ts
aiqiaoy: I would like to make as little function change in this PR as possible but I think this is a good idea. I'll create an issue to add this.

packages/cache/src/internal/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export enum CompressionMethod {
+  Gzip = 'gzip',
+  Zstd = 'zstd'
+}
+
+// Socket timeout in milliseconds during download.  If no traffic is received
+// over the socket during this period, the socket is destroyed and the download
+// is aborted.
+export const SocketTimeout = 5000
```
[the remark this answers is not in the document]
aiqiaoy: I actually made the same comment in @dhadka's original PR that added socket timeout. I think in this case the comment above the variable does the job. I'll leave it as is.

packages/cache/README.md
```diff
@@ -0,0 +1,41 @@
...
+const restoreKeys = [
+    'npm-foobar-',
+    'npm-'
+]
+const cacheKey = await cache.restoreCache(paths, key, restoreKeys)
+```
+
+#### Save Cache
+
+Saves a cache containing the files in `paths` using the `key` provided. Function returns the cache id if the cache was save succesfully.
```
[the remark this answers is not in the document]
aiqiaoy: Path are handled by the glob library we don't really do any path manipulation in cache

packages/cache/RELEASES.md
```diff
@@ -0,0 +1,5 @@
+# @actions/cache Releases
+
+### 1.0.0
```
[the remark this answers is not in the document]
aiqiaoy: Any guideline on how to do beta release? Set the release to 0.0.0?

packages/cache/src/internal/cacheUtils.ts
aiqiaoy: It should always be defined see Josh's comment above but I think we just want to make sure it doesn't fail cache if the folder is for some reason deleted (maybe in self-hosted runner).

packages/cache/src/internal/cacheHttpClient.ts
aiqiaoy: It's for when ACTIONS_CACHE_URL doesn't exist and we have to use the runtime url. ACTIONS_RUNTIME_URL is targeting the pipelines service and this replace change it to target the cache service.

packages/cache/RELEASES.md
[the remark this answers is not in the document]
konradpabjan: With artifacts, we had a `0.1.0` release: https://github.com/actions/toolkit/blob/master/packages/artifact/RELEASES.md

packages/cache/__tests__/cacheUtils.test.ts
aiqiaoy: Added

packages/cache/RELEASES.md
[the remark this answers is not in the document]
thboop: If you set the version to `1.0.0-beta.0` we can push a beta tag using `npm publish --tag beta`. Just make sure whoever is doing the first publish (I can) knows.

Since we are re-using a lot of this code and feel confident about the behavior and interfaces, we can also opt to skip a beta release and just release the package.

packages/cache/src/cache.ts
konradpabjan: Fine for now, just be aware that once `actions/cache` switches over to using this package, the same level of logging will apply (unless there is an option)

packages/cache/src/internal/cacheHttpClient.ts
aiqiaoy: I think we still do but I'll follow up on this.

PR #449, opened 2020-05-06
set base URL for GHES

ericsciple:
Related to ADR for setting GHE URLs (in c2c-actions)

The toolkit should use the GITHUB_API_URL and GITHUB_GRAPHQL_URL to connect to GHES or Dotcom.

### DIFF
--- a/packages/github/src/github.ts
+++ b/packages/github/src/github.ts
@@ -61,6 +61,9 @@ export class GitHub extends Octokit {
     const token = args[0]
     const options = {...args[1]} // Shallow clone - don't mutate the object provided by the caller
 
+    // Base URL - GHES or Dotcom
+    options.baseUrl = options.baseUrl || this.getApiBaseUrl()
+
     // Auth
     const auth = GitHub.getAuthString(token, options)
     if (auth) {
@@ -68,7 +71,7 @@ export class GitHub extends Octokit {
     }
 
     // Proxy
-    const agent = GitHub.getProxyAgent(options)
+    const agent = GitHub.getProxyAgent(options.baseUrl, options)
     if (agent) {
       // Shallow clone - don't mutate the object provided by the caller
       options.request = options.request ? {...options.request} : {}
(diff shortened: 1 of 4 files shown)


packages/github/src/github.ts
```diff
@@ -61,14 +61,19 @@ export class GitHub extends Octokit {
     const token = args[0]
     const options = {...args[1]} // Shallow clone - don't mutate the object provided by the caller
 
+    // Base URL - GHES or Dotcom
```
ericsciple, line 64: I'll look before merging whether any unit test can be added. Might be able to verify the expected baseUrl is set

ericsciple:
notes from offline review:
- bump version, add release notes

packages/github/src/github.ts
ericsciple: unfortunately doesnt appear to be a way to get the options after it's constructed

packages/github/src/github.ts
```diff
@@ -61,14 +61,19 @@ export class GitHub extends Octokit {
...
+    if (!options.baseUrl) {
+      options.baseUrl = this.getApiBaseUrl()
+    }
```
bryanmacfarlane, line 67: other way to write but if you're tested up, no worries.
```javascript 
options.baseUrl = options.baseUrl || this.getApiBaseUrl()
```

packages/github/src/github.ts
```diff
@@ -119,16 +125,31 @@ export class GitHub extends Octokit {
...
+  private static getApiBaseUrl(): string {
+    return process.env['GITHUB_API_URL'] || 'https://api.github.com'
+  }
+
+  private static getGraphQLBaseUrl(): string {
+    let url =
+      process.env['GITHUB_GRAPHQL_URL'] || 'https://api.github.com/graphql'
+
+    // Remove trailing "/graphql"
+    if (url.toUpperCase().endsWith('/GRAPHQL')) {
```
bryanmacfarlane, line 150: There's no possibility of a trailing slash is there?  Alternative https://stackoverflow.com/a/16750711/775184

ericsciple: shouldnt be a trailing slash, i'll trim on launch side when setting the urls

ericsciple: also not bad to add resiliency here, i added code to trim trailing slash if detected

ericsciple:
Going ahead and merging since failure due to audit service being down. Retried twice, still failed. Passed in a previous run and not changing dependencies with this PR.

Issue #402, opened 2020-04-02
@actions/github v2.1.1 missing `deleteDeployment`


maxcountryman:
It seems that `deleteDeployment` is missing from the latest release, v2.1.1:

```
$ node
> const { GitHub } = require('@actions/github');
> const github = new GitHub("secret");
> github.repos.deleteDeployment
undefined
```

thboop:
Should have been resolved with the linked pr, let me know if you see any more issues!

PR #453, opened 2020-05-13
@actions/github v3 using Octokit/core

### LINKED ISSUE #468: Update @octokit dependencies
The @octokit dependencies that this action uses are a little out of date (particularly, @octokit/rest).

In order to reflect API deprecations, new APIs, and get some much-needed bug fixes, we should bump those dependencies to their latest versions.

thboop:
@actions/github is stuck on an older major version of `@octokit/rest` due to the way we extended the class. Let's move away from extending the Octokit class so we don't encounter this again, and move towards the plugin/defaults model all the JS libraries have been using.

This allows us to customize the octokit is a variety of ways, and reuse existing plugins easily.
In particular,  we currently were using `@octokit/rest` which is just @octokit/core with [3 plugins layered on top](https://github.com/octokit/rest.js/blob/master/src/index.ts):

- @octokit/plugin-paginate-rest - provides the paginate function to paginate rest calls
- @octokit/plugin-rest-endpoint-methods - provides the bulk of the rest apis accessible via functions
- @octokit/plugin-request-log - provides additional logging for rest calls

We don't really need the latter, so we can use `@octokit/core` to set up a custom octokit with the paginate and rest endpoint plugins and set sane defaults using environmental variables to automatically handle proxy and GHES url scenarios. 

We can also choose to implement our own custom plugins as needed going forward, or reuse some of the existing plugins, like [GHES api plugins](https://www.npmjs.com/package/@octokit/plugin-enterprise-rest)

Why don't we just keep using `@octokit/rest`?
The core library is really intended to be used if you are [extending the plugin model](https://github.com/octokit/core.js#corejs). Type information does not carry well over multiple exports when using plugins, and this offers far more flexibility (at the cost of having to update more packages)

Resolves #468 , #402

### DIFF
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -19,7 +19,7 @@ async function run() {
 
     const octokit = github.getOctokit(token)
 
-    // You can also pass in additional options as a second parameter to getOptions
+    // You can also pass in additional options as a second parameter to getOctokit
     // const octokit = github.getOctokit(myToken, {userAgent: "MyActionVersion1"});
 
     const { data: pullRequest } = await octokit.pulls.get({
@@ -80,7 +80,7 @@ if (github.context.eventName === 'push') {
 For example, using the `@octokit/plugin-enterprise-server` you can now access enterprise admin apis on GHES instances.
 
 ```ts
-import { GitHub, getOptions } from '@actions/github/utils'
+import { GitHub, getOctokitOptions } from '@actions/github/lib/utils'
 import { enterpriseServer220Admin } from '@octokit/plugin-enterprise-server'
 
 const octokit = GitHub.plugin(enterpriseServer220Admin)
@@ -88,7 +88,7 @@ const octokit = GitHub.plugin(enterpriseServer220Admin)
 // const octokit = GitHub.plugin(enterpriseServer220Admin).defaults({userAgent: "MyNewUserAgent"})
 
 const myToken = core.getInput('myToken');
-const myOctokit = new octokit(getOptions(token))
+const myOctokit = new octokit(getOctokitOptions(token))
 // Create a new user
 myOctokit.enterpriseAdmin.createUser({
   username: "testuser",
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -91,7 +91,7 @@ const myToken = core.getInput('myToken');
 const myOctokit = new octokit(getOctokitOptions(token))
 // Create a new user
 myOctokit.enterpriseAdmin.createUser({
-  username: "testuser",
+  login: "testuser",
   email: "testuser@test.com",
 });
 ```
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -18,8 +18,13 @@ async function run() {
     const myToken = core.getInput('myToken');
 
     const octokit = new github.GitHub(github.getOptions(myToken));
-    // or you may set the authentication on the options yourself
-    // const octokit = new github.GitHub({auth: `token ${myToken}`});
+
+    // You can set the authentication on the options yourself instead
+    // const octokit = new github.GitHub({auth: `token ${myToken}`, userAgent: "MyActionVersion1"});
+
+    // You can also pass in additional options as a second parameter to getOptions
+    // const octokit = new github.GitHub(github.getOptions(myToken), {userAgent: "MyActionVersion1"});
+
     const { data: pullRequest } = await octokit.pulls.get({
         owner: 'octokit',
         repo: 'rest.js',
@@ -34,8 +39,6 @@ async function run() {
 
 run();
 ```
-
-You can pass client options, as specified by [Octokit](https://github.com/octokit/core.js#options), as a second argument getOptions function.
 
 You can also make GraphQL requests. See https://github.com/octokit/graphql.js for the API.
 
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -4,7 +4,7 @@
 
 ## Usage
 
-Returns an authenticated Octokit client that follows the machine [proxy settings](https://help.github.com/en/actions/hosting-your-own-runners/using-a-proxy-server-with-self-hosted-runners). See https://octokit.github.io/rest.js for the API.
+Returns an authenticated Octokit client that follows the machine [proxy settings](https://help.github.com/en/actions/hosting-your-own-runners/using-a-proxy-server-with-self-hosted-runners) and correctly sets GHES base urls. See https://octokit.github.io/rest.js for the API.
 
 ```js
 const github = require('@actions/github');
@@ -17,8 +17,9 @@ async function run() {
     // https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token#about-the-github_token-secret
     const myToken = core.getInput('myToken');
 
-    const octokit = new github.GitHub(myToken);
-
+    const octokit = new github.GitHub(github.getOptions(myToken));
+    // or you may set the authentication on the options yourself
+    // const octokit = new github.GitHub({auth: `token ${myToken}`});
     const { data: pullRequest } = await octokit.pulls.get({
         owner: 'octokit',
         repo: 'rest.js',
@@ -34,7 +35,7 @@ async function run() {
 run();
 ```
 
-You can pass client options, as specified by [Octokit](https://octokit.github.io/rest.js/), as a second argument to the `GitHub` constructor.
+You can pass client options, as specified by [Octokit](https://github.com/octokit/core.js#options), as a second argument getOptions function.
 
 You can also make GraphQL requests. See https://github.com/octokit/graphql.js for the API.
 
--- a/packages/github/README.md
+++ b/packages/github/README.md
@@ -17,13 +17,13 @@ async function run() {
     // https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token#about-the-github_token-secret
     const myToken = core.getInput('myToken');
 
-    const octokit = new github.GitHub(github.getOptions(myToken));
+    const octokit = github.getOctokit(token)
+
+    // You can also pass in additional options as a second parameter to getOptions
+    // const octokit = github.getOctokit(myToken, {userAgent: "MyActionVersion1"});
 
     // You can set the authentication on the options yourself instead
     // const octokit = new github.GitHub({auth: `token ${myToken}`, userAgent: "MyActionVersion1"});
-
-    // You can also pass in additional options as a second parameter to getOptions
-    // const octokit = new github.GitHub(github.getOptions(myToken), {userAgent: "MyActionVersion1"});
 
     const { data: pullRequest } = await octokit.pulls.get({
         owner: 'octokit',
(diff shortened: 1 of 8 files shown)


packages/github/__tests__/github.proxy.test.ts
```diff
@@ -0,0 +1,109 @@
+import * as http from 'http'
```
thboop, line 1: We now load proxy settings when the module is imported because of the way `defaults` works. This required a new test suite for proxy tests.

packages/github/__tests__/github.test.ts
```diff
@@ -85,23 +84,6 @@ describe('@actions/github', () => {
     expect(failed).toBeTruthy()
   })
 
-  it('basic REST client with proxy', async () => {
```
thboop, line 88: moved to new test suite

packages/github/src/github.ts
```diff
@@ -34,125 +25,39 @@ export class GitHub extends Octokit {
...
-    token: string | Octokit.Options,
-    opts?: Octokit.Options
-  ): [string, Octokit.Options] {
-    return [
-      typeof token === 'string' ? token : '',
-      typeof token === 'object' ? token : opts || {}
-    ]
-  }
-
-  private static getOctokitOptions(
```
thboop, line 58: utilities moved to a separate class

packages/github/src/github.ts
```diff
@@ -34,125 +25,39 @@ export class GitHub extends Octokit {
...
-    return url
+const baseUrl = Utils.getApiBaseUrl()
+const defaults = {
+  baseUrl,
+  request: {
+    agent: Utils.getProxyAgent(baseUrl)
   }
 }
+
+export const GitHub = GitHubCore.plugin(
```
thboop, line 60: We can eventually export an Enterprise version using the enterprise plugins instead of `restEndpointMethods` as well to support GHES

export const GitHubEnterprise = GitHubCore.plugin(paginateRest, require("@octokit/plugin-enterprise-server")).defaults(defaults)

packages/github/src/internal/utils.ts
```diff
@@ -0,0 +1,111 @@
...
+export function getProxyAgent(
+  destinationUrl: string,
+  options?: OctokitOptions
+): http.Agent | undefined {
+  if (!options || !options.request || !options.request.agent) {
+    if (httpClient.getProxyUrl(destinationUrl)) {
+      const hc = new httpClient.HttpClient()
+      return hc.getAgent(destinationUrl)
+    }
+  } else if (options && options.request && options.request.agent) {
```
thboop, line 87: This `else if` is the only new. We used to not set the agent if a user specified an agent, but we never actually used the agent specified.

packages/github/src/internal/utils.ts
```diff
@@ -0,0 +1,111 @@
...
+  ]
+}
+
+export function getOctokitOptions(
+  args: [string, OctokitOptions]
+): OctokitOptions {
+  const token = args[0]
+  const options = {...args[1]} // Shallow clone - don't mutate the object provided by the caller
+
+  // Auth
```
thboop, line 33: removed setting up agent and baseurl as that is now done by defaults.

ericsciple:
Should we provide a factory method, to help users mock. Might buy us flexibility also (abstraction) instead of inheriting.

For example:

```js
export function createOctokit(token: string, opts?: OctokitOptions): Octokit {
}
```

Would the types get exported correctly?

Would we need to overload to enable custom list of plugins?

packages/github/src/github.ts
```diff
@@ -1,158 +1,36 @@
...
-    }
-
-    return undefined
-  }
-
-  private static getApiBaseUrl(): string {
-    return process.env['GITHUB_API_URL'] || 'https://api.github.com'
-  }
-
-  private static getGraphQLBaseUrl(): string {
```
thboop, line 143: This was removed, octokit takes care of it: https://github.com/octokit/core.js/issues/65

packages/github/src/github.ts
```diff
@@ -1,158 +1,36 @@
...
-      // Set the agent
-      options.request.agent = agent
-    }
-
-    return options
+const baseUrl = Utils.getApiBaseUrl()
+const defaults = {
+  baseUrl,
+  request: {
+    agent: Utils.getProxyAgent(baseUrl)
```
thboop, line 16: TODO: set default user agent for debugging

packages/github/src/github.ts
```diff
@@ -1,158 +1,36 @@
...
-      url = url.substr(0, url.length - 1)
-    }
-
-    // Remove trailing "/graphql"
-    if (url.toUpperCase().endsWith('/GRAPHQL')) {
-      url = url.substr(0, url.length - '/graphql'.length)
-    }
-    return url
-  }
+export const GitHub = Octokit.plugin(
```
thboop, line 20: Plugins make this really simple. @octokit/rest, which is what we are using in production now, only implements 3 plugins, the two listed here and an additional logging plugin. [source](https://github.com/octokit/rest.js/blob/master/src/index.ts)

We want to save package size, so we don't implement the logging plugin as its default log level doesn't even log, but the `paginateRest` adds the paginate function which is helpful, and `restEndpointMethods` adds nearly all of the rest calls (stuff like `octokit.repos.getBranch`. These two plugins brings us to feature parity with `@actions/github` v2, except for the registerEndpoint method which was deprecated in the newest version of @octokit/rest, and a few other legacy apis.

Default values just set the defaults for a certain field unless a user overrides it. We use that for proxy settings and

packages/github/src/github.ts
```diff
@@ -1,158 +1,36 @@
...
+const baseUrl = Utils.getApiBaseUrl()
+const defaults = {
+  baseUrl,
+  request: {
+    agent: Utils.getProxyAgent(baseUrl)
   }
+}
 
-  private static getGraphQL(args: [string, Octokit.Options]): GraphQL {
-    const defaults: GraphQLRequestParameters = {}
```
thboop, line 87: Octokit now comes with an attached Graphql instance, so this is no longer needed and has been removed.

thboop:
> Should we provide a factory method, to help users mock. Might buy us flexibility also (abstraction) instead of inheriting.
> 
> For example:
> 
> ```js
> export function createOctokit(token: string, opts?: OctokitOptions): Octokit {
> }
> ```
> 
> Would the types get exported correctly?
> 
> Would we need to overload to enable custom list of plugins?

I exported a helper function called `getOptions(token: string, opts?: OctokitOptions)` which supports this scenario nicely.However, it is a little verbose as you have to pass the output of that function into the constructor like
```
import {getOptions, GitHub} from `@actions/github`
import {getInput} from `@actions/core`
const token = getInput("token")

const octokit = new GitHub(getOptions(token))

// or pass in octokit options as well
const octokit = new GitHub(getOptions(token{ timeZone: "America/Los_Angeles"}))

// you may also choose to just format your octokit options correctly
const octokit = new github.GitHub({auth: `token ${myToken}`});
```

But `getOptions` also allows for people to expand on our Instance as needed. 

```
import {getOptions, GitHub} from `@actions/github`
import {getInput} from `@actions/core`
import {myPlugin} from `my-custom-plugin`
const myGitHub = GitHub.plugin(myPlugin)

const token = getInput("token")
const octokit = new myGitHub(getOptions(token), {//whatever octokit options you want})
// use your new octokit with your plugins, or a community plugin
```

packages/github/package.json
```diff
@@ -38,8 +38,9 @@
   },
   "dependencies": {
     "@actions/http-client": "^1.0.3",
-    "@octokit/graphql": "^4.3.1",
-    "@octokit/rest": "^16.43.1"
+    "@octokit/core": "^2.5.1",
+    "@octokit/plugin-paginate-rest": "^2.2.0",
+    "@octokit/plugin-rest-endpoint-methods": "^3.10.0"
```
ericsciple, line 43: Curious whether there is a big impact to overall size on disk?

packages/github/README.md
```diff
@@ -17,7 +17,13 @@ async function run() {
     // https://help.github.com/en/actions/automating-your-workflow-with-github-actions/authenticating-with-the-github_token#about-the-github_token-secret
     const myToken = core.getInput('myToken');
 
-    const octokit = new github.GitHub(myToken);
+    const octokit = new github.GitHub(github.getOptions(myToken));
```
ericsciple, line 20: have you considered an interface like:

```javascript
export function newOctokit(token: string, options?: OctokitOptions, plugin: SOME_TYPE[]): Octokit
```

Advantages:
- Simplify construction
  - `const octokit = github.newOctokit(myToken)`<br />instead of<br />`const octokit = new github.GitHub(github.getOptions(myToken))`
- Simplify by eliminating inheritance (i dont think we add or override any functions?)
- Support plugins?
  - Honestly dont understand how the plugin stuff works with typings. If allowed optional `plugins` arg (assume could use [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) to pass to the plugins function), would the consumer need to cast to whatever type they expect?

ericsciple: Also would enable customers to more easily mock

ericsciple: Sorry just noticed the comments on the conversation tab... ok didnt consider `.plugin` example you had:

```
const myGitHub = GitHub.plugin(myPlugin)
[...]
const octokit = new myGitHub(getOptions(token), {//whatever octokit options you want})
```

In that case, the defaults wouldnt be set though, correct? Late right now. Let's touch base in the morning.

packages/github/package.json
thboop: There shouldn't be, @octokit/rest imported all 3 of these modules. i'll check but it should be less

packages/github/README.md
thboop: The defaults will get set in the plugin scenario. You could also override the default and/or add additional defaults
```
// Always use the baseurl for dotcom, set a default timezone as well.
const myGitHub = GitHub.plugin(myplugin).defaults({
  baseUrl: "https://github.com/api",
  timeZone: "America/Los_Angeles"
})

const myInstance = new myGitHub(getOptions(token))
```

getInstance has a hard time returning the type, because the type is generated when you do `octokit.plugin` or `octokit.defaults`

So a function like

```
export function getInstance(
  token: string,
  options?: OctokitOptions,
  plugins?: OctokitPlugin[]
): Octokit {
  var myClass;
  if (plugins != null){
    myClass = Octokit.plugin(restEndpointMethods, paginateRest, ...plugins).defaults(defaults)
  }
  else {
    myClass = Octokit.plugin(restEndpointMethods, paginateRest).defaults(defaults)
  }
  return new myClass(getOptions(token, options))
}
```

Doesn't have great type information attached. For example, you can run 
```
const myOctokit = getInstance(token)
myOctokit.repos.getBranch({
        owner: 'actions',
        repo: 'toolkit',
        branch: 'master'
})
```

But you we don't get the type information to support it, so consumers will just have to know what functions exist which isn't a great UX.

There might be a way to return typeOf GitHub, but the returned object still would not have the type information associated with the plugins they pass in, so I don't really think we should go down this route.

ericsciple: Consider optimizing for simple case (assumption this is the mainline):

```javascript
// simple case
export function createOctokit(token: string, options?: OctokitOptions)
```

And guidance for advanced users:

```javascript
const octokitType = Octokit.plugin([...])
  .defaults(github.createDefaults())
const octokit = new octokitType(github.createOptions(token, opts))
```

thboop: Done!

PR #454, opened 2020-05-13
tool-cache 1.3.5 release

thboop:


packages/tool-cache/package.json
```diff
@@ -36,7 +36,7 @@
     "url": "https://github.com/actions/toolkit/issues"
   },
   "dependencies": {
-    "@actions/core": "^1.2.0",
+    "@actions/core": "^1.2.3",
```
thboop, line 39: required for isDebug: [Make extract functions quiet by default](https://github.com/actions/toolkit/pull/206)

PR #458, opened 2020-05-14
Correctly reset chunk during artifact upload on retry

konradpabjan:
Fixes weird retry behavior reported here: https://github.com/actions/upload-artifact/issues/71

If there is a connection error or timeout, the readstream should be successfully reset to the beginning so the subsequent retries should not give the same error.

`actions/cache` recently fixed this issue in a PR so this is largely copying over the same fix: https://github.com/actions/cache/pull/305

### DIFF
--- a/.github/workflows/artifact-tests.yml
+++ b/.github/workflows/artifact-tests.yml
@@ -1,5 +1,12 @@
 name: artifact-unit-tests
-on: push
+push:
+  branches:
+    - master
+  paths-ignore:
+    - '**.md'
+pull_request:
+  paths-ignore:
+    - '**.md'
 
 jobs:
   build:
@@ -7,7 +14,7 @@ jobs:
 
     strategy:
       matrix:
-        runs-on: [ubuntu-latest, windows-latest, macOS-latest]
+        runs-on: [ubuntu-latest, windows-latest, macos-latest]
       fail-fast: false
 
     runs-on: ${{ matrix.runs-on }}
--- a/.github/workflows/artifact-tests.yml
+++ b/.github/workflows/artifact-tests.yml
@@ -1,12 +1,13 @@
 name: artifact-unit-tests
-push:
-  branches:
-    - master
-  paths-ignore:
-    - '**.md'
-pull_request:
-  paths-ignore:
-    - '**.md'
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    paths-ignore:
+      - '**.md'
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


.github/workflows/artifact-tests.yml
```diff
@@ -1,13 +1,20 @@
...
+    - '**.md'
 
 jobs:
   build:
     name: Build
 
     strategy:
       matrix:
-        runs-on: [ubuntu-latest, windows-latest, macOS-latest]
+        runs-on: [ubuntu-latest, windows-latest, macos-latest]
```
konradpabjan, line 17: All of our public documentation has `macos` so I'm changing it just for consistency

https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

https://github.com/actions/virtual-environments#available-environments

.github/workflows/artifact-tests.yml
```diff
@@ -1,13 +1,21 @@
 name: artifact-unit-tests
-on: push
+on:
```
konradpabjan, line 2: Also updating these to match our filters in [unit-tests.yml](https://github.com/actions/toolkit/blob/master/.github/workflows/unit-tests.yml)

.github/workflows/artifact-tests.yml
joshmgross: These are case-insensitive and `macOS` is more correct than `macos` based on how Apple names it

packages/artifact/src/internal/upload-http-client.ts
```diff
@@ -208,25 +208,30 @@ export class UploadHttpClient {
...
+        openUploadStream = () => fs.createReadStream(parameters.file)
         isGzip = false
         uploadFileSize = totalFileSize
       } else {
         // create a readable stream using a PassThrough stream that is both readable and writable
-        const passThrough = new stream.PassThrough()
-        passThrough.end(buffer)
-        uploadStream = passThrough
+        openUploadStream = () => {
+          const passThrough = new stream.PassThrough()
```
joshmgross, line 224: I realize this was the existing code, but why do we need a stream that is both readable and writable for upload?

konradpabjan: It's an implementation detail, we have an in-memory buffer and we have to convert somehow to a readable stream. A passthrough stream is used to get the job done without any extra pipping or secondary streams. During the actual upload, it's treated only as a readable stream so it makes no difference.

I think this is where I originally got the idea from: https://stackoverflow.com/questions/16038705/how-to-wrap-a-buffer-as-a-stream2-readable-stream

I recall experimenting with a bunch of other techniques but they all ended up being considerably more complex.

joshmgross: Could we document that in https://github.com/actions/toolkit/blob/master/packages/artifact/docs/implementation-details.md?

konradpabjan: Done: https://github.com/actions/toolkit/blob/296a75104cc117a87b80bab8a03b7f22450b6ba4/packages/artifact/docs/implementation-details.md#uploadcompression-flow

PR #460, opened 2020-05-15
support any auth header

ericsciple:
1) Change downloadTool to support any auth header
2) Change getManifestFromRepo to accept any auth header (consistency within the module)

### DIFF
--- a/packages/tool-cache/package.json
+++ b/packages/tool-cache/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@actions/tool-cache",
-  "version": "1.5.3",
+  "version": "1.5.4",
   "description": "Actions tool-cache lib",
   "keywords": [
     "github",
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -30,12 +30,13 @@ const userAgent = 'actions/tool-cache'
  *
  * @param url       url of tool to download
  * @param dest      path to download tool
+ * @param auth      authorization header
  * @returns         path to downloaded tool
  */
 export async function downloadTool(
   url: string,
   dest?: string,
-  token?: string
+  auth?: string
 ): Promise<string> {
   dest = dest || path.join(_getTempDirectory(), uuidV4())
   await io.mkdirP(path.dirname(dest))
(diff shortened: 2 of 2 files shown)


packages/tool-cache/src/tool-cache.ts
```diff
@@ -501,15 +502,16 @@ interface GitHubTree {
 export async function getManifestFromRepo(
   owner: string,
   repo: string,
-  token: string,
+  auth?: string,
```
ericsciple, line 505: Technically could have left getManifestFromRepo as is... Ended up changing for consistency within the module (downloadTool). Also consistency within the scenario (call getManifestFromRepo then downloadTool).

lmk and i can revert

packages/tool-cache/src/tool-cache.ts
```diff
@@ -30,12 +30,13 @@ const userAgent = 'actions/tool-cache'
...
  * @param url       url of tool to download
  * @param dest      path to download tool
+ * @param auth      authorization header
  * @returns         path to downloaded tool
  */
 export async function downloadTool(
   url: string,
   dest?: string,
-  token?: string
+  auth?: string
```
ericsciple, line 39: Nit pick change since downloadTool is generic (shouldn't assume `token` auth scheme)

packages/tool-cache/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/tool-cache",
-  "version": "1.5.3",
+  "version": "1.5.4",
```
konradpabjan, line 3: `package-lock.json` also looks like it needs to be bumped up

packages/tool-cache/src/tool-cache.ts
```diff
@@ -501,15 +503,17 @@ interface GitHubTree {
...
   branch = 'master'
 ): Promise<IToolRelease[]> {
   let releases: IToolRelease[] = []
   const treeUrl = `https://api.github.com/repos/${owner}/${repo}/git/trees/${branch}`
 
   const http: httpm.HttpClient = new httpm.HttpClient('tool-cache')
-  const headers: IHeaders = {
-    authorization: `token ${token}`
+  const headers: IHeaders = {}
+  if (auth) {
```
ericsciple, line 514: this change was required to download unauthenticated depending on scenario

LocalCragDirtBag (commented):
users/ericsciple/m262auth

LocalCragDirtBag (commented):
`packages/tool-cache/package.json `

LocalCragDirtBag (commented):
`packages/tool-cache/src/tool-cache.ts `

LocalCragDirtBag (commented):
`packages/tool-cache/src/tool-cache.ts` #988

PR #467, opened 2020-05-18
Update Save State Docs

thboop:
Now that the official docs for pre and post actions are up, we should link those in our save state docs!


docs/commands.md
```diff
@@ -114,12 +114,14 @@ echo "::remove-matcher owner=eslint-compact::"
 `remove-matcher` removes a Problem Matcher by owner
 ### Save State
 
-Save state to be used in the corresponding wrapper (finally) post job entry point.
+Save state to be used in the main or post action.
```
TingluoHuang, line 117: pre, main, post?

Issue #462, opened 2020-05-17
cache.saveCache fails for ubuntu-16.04 with zstd enabled


svartalf:
**Describe the bug**

`cache.saveCache` does not work for `ubuntu-16.04` runner, because of the wrong arguments passed to `zstd` CLI utility.

**To Reproduce**

**ubuntu-16.04 log**:

```
##[debug]Checking zstd --version
##[debug]*** zstd command line interface 64-bits v1.3.1, by Yann Collet ***
##[debug]Reserving Cache
##[debug]Resource Url: https://artifactcache.actions.githubusercontent.com/AaeRKueYTJn4WPKwvkhaJ6oa7IeFHpBzq8k8zgGOgH0PDfVyFw/_apis/artifactcache/caches
##[debug]Cache ID: 31
##[debug]implicitDescendants 'false'
##[debug]followSymbolicLinks 'true'
##[debug]implicitDescendants 'false'
##[debug]omitBrokenSymbolicLinks 'true'
##[debug]Search path '/home/runner/work/install/install/target/actions-rs-install/sccache'
##[debug]Matched: target/actions-rs-install/sccache
##[debug]Cache Paths:
##[debug]["target/actions-rs-install/sccache"]
##[debug]Archive Path: /home/runner/work/_temp/fa26d111-0ed0-4633-8777-669233bfbd28/cache.tzst
/bin/tar --use-compress-program zstd -T0 --long=30 -cf cache.tzst -P -C /home/runner/work/install/install --files-from manifest.txt
Incorrect parameters
Usage : 
      zstd [args] [FILE(s)] [-o file] 

FILE    : a filename 
          with no FILE, or when FILE is - , read standard input
Arguments : 
 -#     : # compression level (1-19, default:3) 
 -d     : decompression 
 -D file: use `file` as Dictionary 
 -o file: result stored into `file` (only if 1 input file) 
 -f     : overwrite output without prompting and (de)compress links 
--rm    : remove source file(s) after successful de/compression 
 -k     : preserve source file(s) (default) 
 -h/-H  : display help/long help and exit 
##[error]Tar failed with error: The process '/bin/tar' failed with exit code null
```

**Expected behavior**

Same code on **ubuntu-18.04**:

```
##[debug]Checking zstd --version
##[debug]*** zstd command line interface 64-bits v1.3.3, by Yann Collet ***
##[debug]Reserving Cache
[...]

PR #469, opened 2020-05-18
Fix two issues related to using zstd compression

aiqiaoy:
Addresses: https://github.com/actions/toolkit/issues/462

* Long range mode was added to zstd in [v1.3.2](https://github.com/facebook/zstd/releases?after=v1.3.3) but we are still using v1.3.1 in ubuntu-16.04, so only include `--long` filter if zstd version is greater than v1.3.2

* Change to only use zstd on windows when GnuTar is installed due to bug https://github.com/actions/cache/issues/301

### DIFF
--- a/packages/cache/RELEASES.md
+++ b/packages/cache/RELEASES.md
@@ -3,3 +3,6 @@
 ### 0.1.0
 
 - Initial release
+
+### 0.2.0
+- Fixes two issues around using zstd compression algorithm
--- a/packages/cache/src/internal/cacheUtils.ts
+++ b/packages/cache/src/internal/cacheUtils.ts
@@ -86,14 +86,20 @@ export async function getCompressionMethod(): Promise<CompressionMethod> {
   if (process.platform === 'win32' && !isGnuTarInstalled()) {
     // Disable zstd due to bug https://github.com/actions/cache/issues/301
     return CompressionMethod.Gzip
+  }
+
+  const versionOutput = await getVersion('zstd')
+  const version = semver.clean(versionOutput)
+
+  if (!versionOutput.toLowerCase().includes('zstd command line interface')) {
+    // zstd is not installed
+    return CompressionMethod.Gzip
+  } else if (!version || semver.lt(version, 'v1.3.2')) {
+    // zstd is installed but using a version earlier than v1.3.2
+    // v1.3.2 is required to use the `--long` options in zstd
+    return CompressionMethod.ZstdWithoutLong
   } else {
-    const versionOutput = await getVersion('zstd')
-    const version = semver.clean(versionOutput)
-    return !versionOutput.toLowerCase().includes('zstd command line interface')
-      ? CompressionMethod.Gzip
-      : !version || semver.lt(version, 'v1.3.2')
-      ? CompressionMethod.ZstdOld
-      : CompressionMethod.Zstd
+    return CompressionMethod.Zstd
   }
 }
 
--- a/packages/cache/src/internal/cacheUtils.ts
+++ b/packages/cache/src/internal/cacheUtils.ts
@@ -83,8 +83,8 @@ async function getVersion(app: string): Promise<string> {
 
 // Use zstandard if possible to maximize cache performance
 export async function getCompressionMethod(): Promise<CompressionMethod> {
-  if (process.platform === 'win32') {
-    // Disable zstd on windows due to bug https://github.com/actions/cache/issues/301
+  if (process.platform === 'win32' && !isGnuTarInstalled()) {
+    // Disable zstd due to bug https://github.com/actions/cache/issues/301
     return CompressionMethod.Gzip
   } else {
     const versionOutput = await getVersion('zstd')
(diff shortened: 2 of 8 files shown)


packages/cache/src/internal/tar.ts
```diff
@@ -66,14 +84,24 @@ export async function createTar(
...
     sourceDirectories.join('\n')
   )
+  const workingDirectory = getWorkingDirectory()
+
   // -T#: Compress using # working thread. If # is 0, attempt to detect and use the number of physical CPU cores.
   // --long=#: Enables long distance matching with # bits. Maximum is 30 (1GB) on 32-bit OS and 31 (2GB) on 64-bit.
   // Using 30 here because we also support 32-bit self-hosted runners.
-  const workingDirectory = getWorkingDirectory()
+  // Long range mode is added to zstd in v1.3.2 release, so we will not use --long in older version of zstd.
+  function getProg(): string[] {
```
joshmgross, line 93: nit: `getProg` isn't very descriptive. Maybe `getCompressionArguments`?

packages/cache/src/internal/constants.ts
```diff
@@ -5,6 +5,7 @@ export enum CacheFilename {
 
 export enum CompressionMethod {
   Gzip = 'gzip',
+  ZstdOld = 'zstd-old',
```
joshmgross, line 8: Could we add a comment or be more descriptive in the enum name for what's considered an "Old" version of zstd?

The latest version of zstd is [v1.4.4](https://github.com/facebook/zstd/releases/tag/v1.4.4), so both of the versions we're dealing with could be considered old.

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -82,19 +83,27 @@ async function getVersion(app: string): Promise<string> {
...
+    // Disable zstd due to bug https://github.com/actions/cache/issues/301
+    return CompressionMethod.Gzip
+  } else {
+    const versionOutput = await getVersion('zstd')
+    const version = semver.clean(versionOutput)
+    return !versionOutput.toLowerCase().includes('zstd command line interface')
+      ? CompressionMethod.Gzip
+      : !version || semver.lt(version, 'v1.3.2')
+      ? CompressionMethod.ZstdOld
+      : CompressionMethod.Zstd
```
joshmgross, line 96: This block of quote is a little difficult to parse. It might be better to break this up and add some comments about why we care about `v1.3.2`

```suggestion
    if (!versionOutput.toLowerCase().includes('zstd command line interface')) {
      return CompressionMethod.Gzip
    }
    
    // v1.3.2 is required to use the `--long` options in ZSTD
    if (!version || semver.lt(version, 'v1.3.2') {
      return CompressionMethod.ZstdOld
    }
    
    return CompressionMethod.Zstd
```

aiqiaoy:
> Is there a drastic performance when using the `--long` filter? Would it be simpler to just remove that option from our compression arguments?

I don’t know much about compression algorithms so I can only say based on @kcgen’s [analysis](https://github.com/actions/cache/issues/184#issuecomment-591617222) and zstd long range mode [release note](https://github.com/facebook/zstd/releases/tag/v1.3.2) that the `--long` filter significantly improves the compression ratio if the files have a lot of matches at a large distance. Compression and decompression speed is also dependent on the amount of long matches.

My understanding is that `--long` is good for caching for example node modules and different versions of same library but bad for random files.

packages/cache/src/internal/tar.ts
aiqiaoy: I changed it to `getCompressionProgram()`

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -82,19 +83,33 @@ async function getVersion(app: string): Promise<string> {
 
 // Use zstandard if possible to maximize cache performance
 export async function getCompressionMethod(): Promise<CompressionMethod> {
+  if (process.platform === 'win32' && !isGnuTarInstalled()) {
```
joshmgross, line 86: I'm fine with not messing with zstd on Windows if we want to be consistent across Windows versions.

packages/cache/RELEASES.md
```diff
@@ -2,4 +2,7 @@
 
 ### 0.1.0
 
-- Initial release
\ No newline at end of file
+- Initial release
+
+### 0.2.0
+- Fixes two issues around using zstd compression algorithm
```
joshmgross, line 8:
```suggestion
- Fixes issues with the zstd compression algorithm on Windows and Ubuntu 16.04 [#469](https://github.com/actions/toolkit/pull/469)
```

We might want to link the PRs in the RELEASES notes to help with discoverability

gitfool:
> The latest version of zstd is v1.4.4, so both of the versions we're dealing with could be considered old.

PS. zstd [v1.4.5](https://github.com/facebook/zstd/releases/tag/v1.4.5) recently dropped; it's been 6 months since 1.4.4 so probably worth upgrading the virtual environments.

lazka:
Does anyone here know of an example workflow (or has any hints) on how to make actions/cache use gnu tar and thus zstd on Windows?

PR #471, opened 2020-05-19
Fix a bug with getCompressionMethod and minor edit to release note

aiqiaoy:


aiqiaoy:
@joshmgross Included a bug fix with this PR 🤦

packages/cache/RELEASES.md
```diff
@@ -4,5 +4,5 @@
 
 - Initial release
 
-### 0.2.0
-- Fixes two issues around using zstd compression algorithm
\ No newline at end of file
+### 0.2.1
```
joshmgross, line 7: We should still keep the release notes for 0.2.0 and add a section for 0.2.1

aiqiaoy: Done

PR #475, opened 2020-05-26
Use posix archive format

smorimoto:
Related issue: https://github.com/actions/cache/issues/326

### DIFF
--- a/packages/cache/src/internal/tar.ts
+++ b/packages/cache/src/internal/tar.ts
@@ -101,6 +101,7 @@ export async function createTar(
     }
   }
   const args = [
+    '--posix',
     ...getCompressionProgram(),
     '-cf',
     cacheFileName.replace(new RegExp(`\\${path.sep}`, 'g'), '/'),
--- a/packages/cache/__tests__/tar.test.ts
+++ b/packages/cache/__tests__/tar.test.ts
@@ -135,6 +135,7 @@ test('zstd create tar', async () => {
   expect(execMock).toHaveBeenCalledWith(
     `"${tarPath}"`,
     [
+      '--posix',
       '--use-compress-program',
       'zstd -T0 --long=30',
       '-cf',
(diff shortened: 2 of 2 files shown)


imbsky:
I've checked that this works on my Windows machine, so there should be no major problem.

imbsky:
@joshmgross @aiqiaoy friendly ping

packages/cache/src/internal/tar.ts
```diff
@@ -62,6 +62,7 @@ export async function extractTar(
     }
   }
   const args = [
+    '--posix',
```
joshmgross, line 65: What happens if `--posix` is called on an archive that wasn't created with the `--posix` flag?

imbsky: I will check it.

OlegYch: `--posix` is not allowed when extracting
```
>tar --posix -xzf qwe.tgz qwe                   
tar: Option --format is not permitted in mode -x
```

imbsky: Oops, all right. I will update this PR.

imbsky:
Status?

smorimoto:
Hi @bryanmacfarlane. Can you check this?

smorimoto:
If necessary, please let me know so that I can help you with the maintenance of the cache action related projects.

aiqiaoy:
I'm gonna pull the branch and test on my mac (mac and linux runner is using `zstd` tar)

smorimoto:
Sorry. I will be in an area with an unstable Internet connection until the 29th. Of course, I can work on this later. But, if you don't mind, this branch is set to be editable by you, so feel free to push those changes.

smorimoto:
I lost the repository because of an accident, so I will close this and open another PR later.

smorimoto:
This was merged as https://github.com/actions/toolkit/pull/533.

PR #492, opened 2020-06-04
Explicitly exclude DS store files

### LINKED ISSUE #491: .DS_Store in @actions/github 3.0.0
When upgrading to `"@actions/github": "^3.0.0"` a MacOS `.DS_Store` file was included.
I am reporting this so packages remain clean ;)
image

thboop:
Related to this npm issue: 
https://github.com/npm/cli/issues/440#issuecomment-552069243

We published a release of `@actions/github` that included a DS_STORE file. It appears that the default .npmignore is overwritten by the files list. So we can add !.DS_Store to this list to ensure we don't publish any more DS_STORE files. 

Tested locally using `npm publish --dry-run` and `npm pack --dry-run`
Resolves #491

### DIFF
--- a/packages/artifact/package.json
+++ b/packages/artifact/package.json
@@ -17,7 +17,8 @@
     "test": "__tests__"
   },
   "files": [
-    "lib"
+    "lib",
+    "!.DS_Store"
   ],
   "publishConfig": {
     "access": "public"
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -17,7 +17,8 @@
     "test": "__tests__"
   },
   "files": [
-    "lib"
+    "lib",
+    "!.DS_Store"
   ],
   "publishConfig": {
     "access": "public"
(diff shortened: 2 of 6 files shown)


thboop:
> Looks good to me!
> 
> I briefly thought about explicitly specifying the file types (something like this)
> 
> ```
> lib/*.js
> lib/*.js.map
> lib/*.d.ts
> ```
> 
> This would prevent other types of files types from showing up accidentally (not just `.DS_Store`). But I don't think this is necessary. Thoughts?

There are some other file types [automatically ignored](https://docs.npmjs.com/using-npm/developers.html#keeping-files-out-of-your-package) that could also get added. We could prevent those by explicitly stating the file types we support, but looking through that list nothing really sticks out to me as something that we would accidentally include.

Hopefully this is just a short term fix until an npm bugfix is in, so I think we can avoid specifying explicit file types for now.

konradpabjan:
> Hopefully this is just a short term fix until an npm bugfix is in, so I think we can avoid specifying explicit file types for now.

Agree, the chances of other types of files accidentally showing up is pretty small so this is fine. Once the upstream issue is fixed, we can get rid of this.

Ship it! 🚀

PR #493, opened 2020-06-04
Adds option to download caches using AzCopy

dhadka:
Adds option to download caches using AzCopy.  This improves download speed and helps reduce flakiness.  AzCopy downloads the file in 4 MB chunks, so it has more granular control over retries and parallelism.

This download option is selected when
1. The cache URL is from Azure blob storage
2. The `DISABLE_AZCOPY` env var is *not* set (used to test the http-client download path)
3. `azcopy` is available on the runner

For hosted runners, `azcopy` is available on Ubuntu and Mac, but not Windows.  On Ubuntu, we use the `azcopy10` alias since `azcopy` refers to a much earlier version (7.x.x).

### Stats

The table below shows the average time to download a 1 GB file from an EastUS storage account to a VM hosted in EastUS2 or WestUS:

```
Client           EastUS2 VM          WestUS VM
---------------  ------------------  ----------------
http-client      43.6 s              141 s
azcopy           7.5 s               22.4 s
```

### TODO
- [ ] Get AzCopy installed on Windows runners (https://github.com/actions/virtual-environments/issues/1004)

### DIFF
--- a/packages/cache/src/internal/cacheUtils.ts
+++ b/packages/cache/src/internal/cacheUtils.ts
@@ -116,12 +116,14 @@ export async function isGnuTarInstalled(): Promise<boolean> {
 
 export async function getAzCopyCommand(): Promise<string | undefined> {
   // Always prefer the azcopy10 alias first, which is the correct version on Ubuntu.
-  if ((await getVersion('azcopy10')).toLowerCase().startsWith('azcopy version')) {
+  let versionOutput = await getVersion('azcopy10')
+
+  if (versionOutput.toLowerCase().startsWith('azcopy version')) {
     return 'azcopy10'
   }
 
   // Fall back to any azcopy that is version 10 or newer.
-  const versionOutput = await getVersion('azcopy')
+  versionOutput = await getVersion('azcopy')
 
   if (versionOutput.toLowerCase().startsWith('azcopy version')) {
     const version = versionOutput.substring(15)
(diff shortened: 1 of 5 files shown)


packages/cache/src/internal/cacheUtils.ts
```diff
@@ -113,3 +113,28 @@ export async function isGnuTarInstalled(): Promise<boolean> {
   const versionOutput = await getVersion('tar')
   return versionOutput.toLowerCase().includes('gnu tar')
 }
+
+export async function getAzCopyCommand(): Promise<string | undefined> {
+  // Always prefer the azcopy10 alias first, which is the correct version on Ubuntu.
```
aiqiaoy, line 118: What do you mean by correct version? What happens if we just do the version 10 or newer check?

dhadka: `azcopy` on Ubuntu is version 7.x.x, which uses a completely different CLI.  So we need to use the `azcopy10` alias on Ubuntu to get to version 10.

This is different on a Mac, where `azcopy` is v10.  Windows currently does not have AzCopy installed, but I already filed an issue to fix that.

I'm not sure why it's setup this way, but this logic is needed to ensure we get the correct version on Ubuntu.

dhadka:
Closing, PR to use the NodeJS library for Azure storage used instead of AzCopy.

PR #497, opened 2020-06-08
Use Azure storage SDK to download cache

dhadka:
Use the Azure Storage SDK for Node (https://github.com/Azure/azure-storage-node) when downloading cache content hosted on Azure blob storage.  This is intended to improve reliability and performance because the download call, `downloadToBuffer`, downloads the file in 4 MB chunks which can be independently retried and parallelized.

In testing, we observed download speeds reduced by about 50% using this approach and a significant improvement in reliability.  Additionally, this follows the suggestions made by Azure in our investigation into download flakiness (https://github.com/actions/cache/issues/267).

One limitation of `downloadToBuffer` is Node only supports buffers up to ~1 GB on 32-bit systems and ~2 GBs on 64-bit systems.  While the vast majority of caches are < 1 GB, there are a good number of caches > 1 GB and we technically allow up to 5 GBs.  As a result, any download that exceeds the buffer limit will be split up and downloaded in multiple calls to `downloadToBuffer`.

Proxies are still supported using the existing `http_proxy` and `https_proxy` environment variables.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   },
   "devDependencies": {
     "@types/jest": "^24.0.11",
-    "@types/node": "^11.13.5",
+    "@types/node": "^12.12.47",
     "@types/signale": "^1.2.1",
     "@typescript-eslint/parser": "^2.2.7",
     "concurrently": "^4.1.0",
--- a/packages/cache/RELEASES.md
+++ b/packages/cache/RELEASES.md
@@ -9,3 +9,6 @@
 
 ### 0.2.1
 - Fix to await async function getCompressionMethod
+
+### 0.3.0
+- Downloads Azure-hosted caches using AzCopy when available
--- a/packages/cache/RELEASES.md
+++ b/packages/cache/RELEASES.md
@@ -11,4 +11,4 @@
 - Fix to await async function getCompressionMethod
 
 ### 0.3.0
-- Downloads Azure-hosted caches using AzCopy when available
+- Downloads Azure-hosted caches using the Azure SDK for speed and reliability
--- a/packages/cache/RELEASES.md
+++ b/packages/cache/RELEASES.md
@@ -10,5 +10,5 @@
 ### 0.2.1
 - Fix to await async function getCompressionMethod
 
-### 0.3.0
+### 1.0.0
 - Downloads Azure-hosted caches using the Azure SDK for speed and reliability
--- a/packages/cache/RELEASES.md
+++ b/packages/cache/RELEASES.md
@@ -12,3 +12,5 @@
 
 ### 1.0.0
 - Downloads Azure-hosted caches using the Azure SDK for speed and reliability
+- Includes changes that break compatibility with earlier versions, including:
+  - `retry`, `retryTypedResponse`, and `retryHttpClientResponse` moved from `cacheHttpClient` to `requestUtils`
--- a/packages/cache/src/internal/cacheHttpClient.ts
+++ b/packages/cache/src/internal/cacheHttpClient.ts
@@ -1,21 +1,13 @@
 import * as core from '@actions/core'
-import {HttpClient, HttpCodes} from '@actions/http-client'
+import {HttpClient} from '@actions/http-client'
 import {BearerCredentialHandler} from '@actions/http-client/auth'
-import {
-  IHttpClientResponse,
-  IRequestOptions,
-  ITypedResponse
-} from '@actions/http-client/interfaces'
-import {BlockBlobClient} from '@azure/storage-blob'
-import * as buffer from 'buffer'
+import {IRequestOptions, ITypedResponse} from '@actions/http-client/interfaces'
 import * as crypto from 'crypto'
 import * as fs from 'fs'
-import * as stream from 'stream'
 import {URL} from 'url'
-import * as util from 'util'
 
 import * as utils from './cacheUtils'
-import {CompressionMethod, SocketTimeout} from './constants'
+import {CompressionMethod} from './constants'
 import {
   ArtifactCacheEntry,
   InternalCacheOptions,
@@ -23,35 +15,15 @@ import {
   ReserveCacheRequest,
   ReserveCacheResponse
 } from './contracts'
+import {downloadCacheHttpClient, downloadCacheStorageSDK} from './downloadUtils'
 import {DownloadOptions, UploadOptions} from '../options'
+import {
+  isSuccessStatusCode,
+  retryHttpClientResponse,
+  retryTypedResponse
+} from './requestUtils'
 
 const versionSalt = '1.0'
-
-function isSuccessStatusCode(statusCode?: number): boolean {
-  if (!statusCode) {
-    return false
-  }
-  return statusCode >= 200 && statusCode < 300
-}
-
-function isServerErrorStatusCode(statusCode?: number): boolean {
-  if (!statusCode) {
-    return true
-  }
-  return statusCode >= 500
-}
-
-function isRetryableStatusCode(statusCode?: number): boolean {
-  if (!statusCode) {
-    return false
-  }
-  const retryableStatusCodes = [
-    HttpCodes.BadGateway,
-    HttpCodes.ServiceUnavailable,
-    HttpCodes.GatewayTimeout
-  ]
-  return retryableStatusCodes.includes(statusCode)
-}
 
 function getCacheApiUrl(resource: string): string {
   // Ideally we just use ACTIONS_CACHE_URL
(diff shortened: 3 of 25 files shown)


packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -256,6 +260,64 @@ export async function downloadCache(
...
+      core.debug(`Downloading segment at offset ${offset} with length ${segmentSize}...`)
+      
+      const result = await client.downloadToBuffer(offset, segmentSize, { concurrency: os.cpus().length * 8 })
+      fs.writeFileSync(fd, result)
+
+      core.debug(`Finished segment at offset ${offset}`)
+      offset += segmentSize
+    }
+
+    fs.closeSync(fd)
```
joshmgross, line 298: Should we need to put this in a `try / finally` block in case `downloadToBuffer` throws an error? We want to ensure we close the `fd` regardless of any errors

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -256,6 +260,64 @@ export async function downloadCache(
...
+  }
+}
+
+export async function downloadCache(
+  archiveLocation: string,
+  archivePath: string
+): Promise<void> {
+  const archiveUrl = new URL(archiveLocation)
+  const disableAzureSdk = process.env['DISABLE_AZURE_SDK'] ?? ''
```
joshmgross, line 307: Instead of environment variables, we should prefer having this as an option on `downloadCache`. Environment variables aren't friendly to action authors. If we want a way to disable this from a workflow, we could look for this environment variable in the cache action directly instead.

To make compat easier and enable future options we could do:
```ts
export async function downloadCache(
  archiveLocation: string,
  archivePath: string,
  options?: DownloadOptions
): Promise<void> {
```

with 
```ts
export interface DownloadOptions {
  /**
   * Indicates whether to use the Azure Blob SDK to download caches that are stored with Azure Blob Storage.
   * This option provides better reliability and performance, etc...
   *
   * @default true
   */
  useAzureSDK?: boolean

}
```
`@actions/glob` has a good example of this: https://github.com/actions/toolkit/blob/master/packages/glob/src/internal-glob-options.ts

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -256,6 +260,64 @@ export async function downloadCache(
...
+
+  if (
+    archiveUrl.hostname.endsWith('.blob.core.windows.net') &&
```
joshmgross, line 310: Do we have any guarantees that all blob storage downloads will be at this address?

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -256,6 +260,64 @@ export async function downloadCache(
...
+    const maxSegmentSize = buffer.kMaxLength
+    let offset = 0
+
+    const fd = fs.openSync(archivePath, 'w')
+
+    while (offset < contentLength) {
+      const segmentSize = Math.min(maxSegmentSize, contentLength-offset)
+      core.debug(`Downloading segment at offset ${offset} with length ${segmentSize}...`)
+      
+      const result = await client.downloadToBuffer(offset, segmentSize, { concurrency: os.cpus().length * 8 })
```
joshmgross, line 291: Where did this concurrency calculation come from?

packages/cache/src/internal/cacheHttpClient.ts
dhadka: Fixed, added try-finally block

packages/cache/src/internal/cacheHttpClient.ts
dhadka: Fixed, moved to DonwloadOptions

packages/cache/src/internal/cacheHttpClient.ts
dhadka: Previous versions of AzCopy created 8*ncpu download threads.  However, any improvement seems to be marginal going beyond 8, so I've made it configurable with a default of 8.

packages/cache/src/internal/cacheHttpClient.ts
dhadka: Guarantee? No. This is what they've always used and I don't see it changing anytime soon.  If it were to change, we would just fall back to http-client.

dhadka:
@joshmgross Any idea why the toolkit `npm test` fails with:

```
    packages/cache/src/internal/cacheHttpClient.ts:290:35 - error TS2339: Property 'constants' does not exist on type 'typeof import("buffer")'.

    290     const maxSegmentSize = buffer.constants.MAX_LENGTH
```

When the action runs fine in the cache tests?

joshmgross:
> @joshmgross Any idea why the toolkit `npm test` fails with:
> 
> ```
>     packages/cache/src/internal/cacheHttpClient.ts:290:35 - error TS2339: Property 'constants' does not exist on type 'typeof import("buffer")'.
> 
>     290     const maxSegmentSize = buffer.constants.MAX_LENGTH
> ```
> 
> When the action runs fine in the cache tests?

Looks like it was just a type issue with `@types/node`. Since all actions run on the LTS version of Node 12, I installed the latest version of `@types/node@12`

package.json
```diff
@@ -14,7 +14,7 @@
   },
   "devDependencies": {
     "@types/jest": "^24.0.11",
-    "@types/node": "^11.13.5",
+    "@types/node": "^12.12.47",
```
joshmgross, line 17: @thboop @ericsciple FYI I bumped this package version since all actions run on Node 12 and it was causing issues in the test, see https://github.com/actions/toolkit/pull/497#issuecomment-644996955

dhadka:
@joshmgross Can you please take one more pass?  My last commit added some testing.  In order to make it work with `jest`, I had to move around some methods, including adding the following:

* `requestUtils.ts` - Contains the retry logic and methods that check HTTP status codes
* `downloadUtils.ts` - Contains the actual download logic

If you're good, then this should be ready to merge (I don't have write permission on this repo so I can't merge).

packages/cache/src/internal/downloadUtils.ts
```diff
@@ -0,0 +1,134 @@
...
+export async function downloadCacheStorageSDK(
+  archiveLocation: string,
+  archivePath: string,
+  options?: DownloadOptions
+): Promise<void> {
+  const client = new BlockBlobClient(archiveLocation, undefined, {
+    retryOptions: {
+      // Override the timeout used when downloading each 4 MB chunk
+      // The default is 2 min / MB, which is way too slow
+      tryTimeoutInMs: options?.timeoutInMs ?? 30000
```
joshmgross, line 88: I think we should do something similar to https://github.com/actions/toolkit/blob/master/packages/glob/src/internal-glob-options-helper.ts so that we can initialize a default options object and keep all the defaults in a single place.

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -110,75 +85,6 @@ export function getCacheVersion(
     .digest('hex')
 }
 
-export async function retry<T>(
```
joshmgross, line 113: Unfortunately, I think this is a breaking change. If a user was importing `retry` or any other exported function from this file, their imports will break as they have to now import from `downloadUtils` or `requestUtils`

This change makes sense to me though, and I like the extra testing it allows. We can bump the major version and list these breaking changes in the release.

packages/cache/src/internal/downloadUtils.ts
dhadka: Done.  The one issue I hit with this approach is that even though we fill in the default value if it's undefined, the interface still shows the properties as optional, so we still need some check if it's undefined.  I ended up adding an `assertDefined` method that throws if the property is undefined, but we should never hit that issue...it's just to make TypeScript happy.

If you know of a better way to do this, please let me know.

dhadka: (I also tried using the non-null operator `!`, but lint doesn't like it)

packages/cache/src/internal/cacheHttpClient.ts
dhadka: Bumped the version to `1.0.0`

packages/cache/RELEASES.md
```diff
@@ -8,4 +8,7 @@
 - Fixes issues with the zstd compression algorithm on Windows and Ubuntu 16.04 [#469](https://github.com/actions/toolkit/pull/469)
 
 ### 0.2.1
-- Fix to await async function getCompressionMethod
\ No newline at end of file
+- Fix to await async function getCompressionMethod
+
+### 1.0.0
+- Downloads Azure-hosted caches using the Azure SDK for speed and reliability
```
joshmgross, line 14: Could we list the breaking changes here?

dhadka: Done.  Are there any other breaking changes we would want to make now?

dhadka: I'm working on a follow-on PR to display progress.  Let's not release 1.0.0 just yet.

dhadka: @joshmgross Here's the follow-on PR - https://github.com/dhadka/toolkit/pull/2

packages/cache/src/internal/requestUtils.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+  return statusCode >= 500
+}
+
+export function isRetryableStatusCode(statusCode?: number): boolean {
+  if (!statusCode) {
+    return false
+  }
+  const retryableStatusCodes = [
+    HttpCodes.BadGateway,
```
yaananth, line 27: Do we get throttled here? Should we retry on `429`?

packages/cache/src/options.ts
```diff
@@ -15,3 +17,93 @@ export interface UploadOptions {
...
+ */
+export function getDownloadOptions(copy?: DownloadOptions): DownloadOptions {
+  const result: DownloadOptions = {
+    useAzureSdk: true,
+    downloadConcurrency: 8,
+    timeoutInMs: 30000
+  }
+
+  if (copy) {
+    if (typeof copy.useAzureSdk === 'boolean') {
```
yaananth, line 91: any reason for checking these types? we are already in typescript, we should have good typechecking. I guess are these sent by customers and they can send anything they want? Why don't we validate them and fail if that's the case

packages/cache/src/internal/requestUtils.ts
yacaovsnc: I don't think we should retry on throttling - it probably will make it worse, no?  :D

packages/cache/src/options.ts
dhadka: I think the purpose of this if condition is to check if the value is set (not undefined and not null).  So I think this is just an easier expression.  (This pattern is used elsewhere, such as https://github.com/actions/toolkit/blob/1a2c5929039617047c4b37b17b1c4cd6de0b8333/packages/glob/src/internal-glob-options-helper.ts#L15)

packages/cache/src/internal/requestUtils.ts
dhadka: (Note that I didn't add this code, just moved it to a different file for testing purposes)

I believe it's this way because a 429 would generally mean we need to wait for some period of time. The question then becomes whether it's better for us to wait and retry, or just "cache miss" and let the user rebuild the cache content.  These retries are only used when accessing the ArtifactCache API, so this would mean RU is throttling the host and that host would likely remain throttled for some time.

PR #498, opened 2020-06-08
Setup Weekly Automation to Update @actions/github

thboop:
This automation will create a PR when any of the prod dependencies in actions/github have been updated. This should simplify the process of keeping the action up to date with octokit releases.

Also included is a bump to the versions of these dependencies, created by the  automation.

It runs every sunday at 18:00 UTC

### DIFF
--- a/.github/workflows/update-github.yaml
+++ b/.github/workflows/update-github.yaml
@@ -22,8 +22,8 @@ jobs:
       run: | 
         if [[ "$(git status --porcelain)" != "" ]]; then 
         echo "::set-output name=createPR::true"
-        git config --global user.email "bot@github.com"
-        git config --global user.name "Dependency Update Bot"
+        git config --global user.email "github-actions[bot]@users.noreply.github.com"
+        git config --global user.name "github-actions[bot]"
         git checkout -b bots/updateGitHubDependencies-${{github.run_number}}
         git add .
         git commit -m "Update Dependencies"
--- a/.github/workflows/update-github.yaml
+++ b/.github/workflows/update-github.yaml
@@ -1,7 +1,6 @@
 name: "UpdateOctokit"
 
 on:
-  push:
   schedule:
     - cron: '0 18 * * 0'
 
--- a/.github/workflows/update-github.yaml
+++ b/.github/workflows/update-github.yaml
@@ -15,7 +15,8 @@ jobs:
     - name: Update Octokit
       working-directory: packages/github 
       run: |
-        npm update
+        npx npm-check-updates -u --dep prod
+        npm install
     - name: Check Status
       id: status
       working-directory: packages/github 
--- a/.github/workflows/update-github.yaml
+++ b/.github/workflows/update-github.yaml
@@ -1,0 +1,15 @@
+name: "Update Actions GitHub Octokit"
+
+on:
+  push:
+  schedule:
+    - cron: '0 0 * * 0'
+
+jobs:
+  Update Actions GitHub:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
--- a/.github/workflows/update-github.yaml
+++ b/.github/workflows/update-github.yaml
@@ -1,13 +1,14 @@
 name: "UpdateOctokit"
 
 on:
+  push:
   schedule:
     - cron: '0 18 * * 0' # sunday at 18 UTC
 
 jobs:
   UpdateOctokit:
     runs-on: ubuntu-latest
-    if: ${{ github.repository_owner == 'actions' }}
+    # if: ${{ github.repository_owner == 'actions' }}
     steps:
     - name: Checkout repository
       uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/update-github.yaml
```diff
@@ -0,0 +1,45 @@
...
+        npx npm-check-updates -u --dep prod
+        npm install
+    - name: Check Status
+      id: status
+      working-directory: packages/github 
+      run: | 
+        if [[ "$(git status --porcelain)" != "" ]]; then 
+        echo "::set-output name=createPR::true"
+        git config --global user.email "bot@github.com"
+        git config --global user.name "Dependency Update Bot"
```
peaceiris, line 26:
```suggestion
        git config --global user.email "github-actions[bot]@users.noreply.github.com"
        git config --global user.name "github-actions[bot]"
```

I am using this way in [peaceiris/actions-gh-pages: GitHub Actions for GitHub Pages](https://github.com/peaceiris/actions-gh-pages).

peaceiris:
Why don't you use the Dependabot? Just curious.

.github/workflows/update-github.yaml
```diff
@@ -0,0 +1,45 @@
...
+        git checkout -b bots/updateGitHubDependencies-${{github.run_number}}
+        git add .
+        git commit -m "Update Dependencies"
+        git push --set-upstream origin bots/updateGitHubDependencies-${{github.run_number}}; fi
+    - name: Create PR
+      if: ${{steps.status.outputs.createPR}}
+      uses: actions/github-script@v2
+      with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
```
ericsciple, line 35: fyi, it looks like you could omit `github-token: ...` since `${{ github.token }}` is [the default](https://github.com/actions/github-script/blob/v2/action.yml#L13)

.github/workflows/update-github.yaml
```diff
@@ -0,0 +1,45 @@
...
+      working-directory: packages/github 
+      run: | 
+        if [[ "$(git status --porcelain)" != "" ]]; then 
+        echo "::set-output name=createPR::true"
+        git config --global user.email "bot@github.com"
+        git config --global user.name "Dependency Update Bot"
+        git checkout -b bots/updateGitHubDependencies-${{github.run_number}}
+        git add .
+        git commit -m "Update Dependencies"
+        git push --set-upstream origin bots/updateGitHubDependencies-${{github.run_number}}; fi
```
ericsciple, line 30: should `fi` be on it's own line? (strange style, since multiline block)

.github/workflows/update-github.yaml
```diff
@@ -0,0 +1,45 @@
...
+  UpdateOctokit:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Update Octokit
+      working-directory: packages/github 
+      run: |
+        npx npm-check-updates -u --dep prod
```
ericsciple, line 17: - consider `--upgrade` instead of `-u` (long form for readability in scripts, short form for interactive terminal)
- is it fragile to assume `npm-check-updates` is installed globally?
- curious whether considered running npm update commands instead?

.github/workflows/update-github.yaml
```diff
@@ -0,0 +1,45 @@
+name: "UpdateOctokit"
+
+on:
+  schedule:
+    - cron: '0 18 * * 0'
+
+jobs:
+  UpdateOctokit:
```
ericsciple, line 8: will forks inherit this cron workflow? Should we add a job-condition so it doesnt run on forks?

.github/workflows/update-github.yaml
```diff
@@ -0,0 +1,45 @@
+name: "UpdateOctokit"
+
+on:
+  schedule:
+    - cron: '0 18 * * 0'
```
ericsciple, line 5: can you add a comment in code, what this cron syntax means? According to [`https://crontab.guru/#0_18_*_*_0`](https://crontab.guru/#0_18_*_*_0) it means: "At 18:00 on Sunday"

.github/workflows/update-github.yaml
thboop: - Great Idea, I'll update it
- npx will install items if they don't [exist in path](https://www.npmjs.com/package/npx)
- Update respect semver, which is nice, but ` npm-check-updates` will create a new pr will notify for new major versions.  Even if the PR isn't perfect (the new version isn't compatible), the pr will act as a notification that we need to do something!

packages/github/package.json
```diff
@@ -38,10 +38,10 @@
...
   },
   "dependencies": {
-    "@actions/http-client": "^1.0.3",
-    "@octokit/core": "^2.5.1",
-    "@octokit/plugin-paginate-rest": "^2.2.0",
-    "@octokit/plugin-rest-endpoint-methods": "^3.10.0"
+    "@actions/http-client": "^1.0.8",
+    "@octokit/core": "^3.0.0",
+    "@octokit/plugin-paginate-rest": "^2.2.3",
+    "@octokit/plugin-rest-endpoint-methods": "^4.0.0"
```
thboop, line 44: Lots of changes to actions api here, going to release a new major version:
https://github.com/octokit/plugin-rest-endpoint-methods.js/releases/tag/v4.0.0

Issue #371, opened 2020-03-08
@actions/github should populate Run Number and Run ID


joshmgross:
### Background
The `@actions/github` package populates a GitHub `Context` similar to [`github` context](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) available in workflows, see https://github.com/actions/toolkit/blob/master/packages/github/src/context.ts

### Issue
This context is missing the Run ID and Run Number, available as `GITHUB_RUN_NUMBER` and `GITHUB_RUN_ID`, see https://help.github.com/en/actions/configuring-and-managing-workflows/using-environment-variables#default-environment-variables

We should add these values to the package's context and ensure that the workflow `github` context matches what's available in this package.

Related: https://github.com/actions/toolkit/issues/65

gitwdel:
#371

PR #499, opened 2020-06-14
Add execution state information

8398a7:
I created a PR because I wanted this feature too.

I've been dealing with the following workaround, but it would be nice to be able to use it in the library.
https://github.com/8398a7/action-slack/blob/v3.1.0/src/client.ts#L143

Resolves: https://github.com/actions/toolkit/issues/371

### DIFF
--- a/packages/github/src/context.ts
+++ b/packages/github/src/context.ts
@@ -15,8 +15,8 @@ export class Context {
   workflow: string
   action: string
   actor: string
-  runNumber: string
-  runId: string
+  runNumber: number
+  runId: number
 
   /**
    * Hydrate the context from the environment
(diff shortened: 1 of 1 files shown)


8398a7:
I would like to add `GITHUB_JOB` as well, if that's OK with you.
https://github.com/8398a7/action-slack/blob/v3.1.0/src/client.ts#L150

packages/github/src/context.ts
```diff
@@ -15,6 +15,8 @@ export class Context {
   workflow: string
   action: string
   actor: string
+  runNumber: string
```
thboop, line 18: Hey @8398a7 , `runNumber` and `runId` are numbers in the [rest api](
https://developer.github.com/v3/actions/workflow-runs/#list-workflow-runs) , we should match that here.

8398a7: Thanks for the review.
I fixed it in 728bcb5.

thboop:
> I would like to add GITHUB_JOB as well, if that's OK with you.

Sounds great! This PR looks good, do you want to open a separate pr for this, or amend this one?

8398a7:
> Sounds great! This PR looks good, do you want to open a separate pr for this, or amend this one?

Thank you. I've made some changes to this PR.

I made a mistake in the template for the commit message. I did a force push to fix it. Sorry.

thboop:
Going to include this in the 4.0.0 release. Thank you for the contribution!

https://github.com/actions/toolkit/pull/507

PR #513, opened 2020-07-13
Display download progress when using Azure Storage SDK

dhadka:
Creates a timer to display download progress every second, including the # of bytes received and the average download speed.  This is intended to provide better visibility into the download, especially for larger caches.

Here's an example restoring a 500 MB cache:

```
##[debug]Archive Path: /home/runner/work/_temp/7ad5a251-5ec2-455f-ad81-d9968ddcad82/cache.tzst
##[debug]Use Azure SDK: true
##[debug]Download concurrency: 8
##[debug]Request timeout (ms): 30000
##[debug]Downloading segment at offset 0 with length 524301052...
Received 171966464 of 524301052 (32.8%), 164.0 MBs/sec
Received 268435456 of 524301052 (51.2%), 127.9 MBs/sec
Received 486539264 of 524301052 (92.8%), 154.6 MBs/sec
Received 524301052 of 524301052 (100.0%), 143.6 MBs/sec
Cache Size: ~500 MB (524301052 B)
/bin/tar --use-compress-program zstd -d -xf /home/runner/work/_temp/7ad5a251-5ec2-455f-ad81-d9968ddcad82/cache.tzst -P -C /home/runner/work/cache-canary/cache-canary
```

### DIFF
--- a/packages/cache/src/internal/downloadUtils.ts
+++ b/packages/cache/src/internal/downloadUtils.ts
@@ -120,12 +120,12 @@ export class DownloadProgress {
   }
 
   /**
-   * Callback used to update the progress.
-   *
-   * @param progress the progress event
+   * Returns a function used to handle TransferProgressEvents.
    */
-  onProgressCallback(progress: TransferProgressEvent): void {
-    this.setReceivedBytes(progress.loadedBytes)
+  onProgress(): (progress: TransferProgressEvent) => void {
+    return (progress: TransferProgressEvent) => {
+      this.setReceivedBytes(progress.loadedBytes)
+    }
   }
 
   /**
@@ -261,9 +261,7 @@ export async function downloadCacheStorageSDK(
           segmentSize,
           {
             concurrency: options.downloadConcurrency,
-            onProgress: downloadProgress.onProgressCallback.bind(
-              downloadProgress
-            )
+            onProgress: downloadProgress.onProgress()
           }
         )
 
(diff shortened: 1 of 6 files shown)


packages/cache/src/internal/downloadUtils.ts
```diff
@@ -107,27 +241,36 @@ export async function downloadCacheStorageSDK(
...
+      while (!downloadProgress.isDone()) {
+        const segmentSize = Math.min(
+          maxSegmentSize,
+          contentLength - downloadProgress.segmentOffset
         )
 
-        const result = await client.downloadToBuffer(offset, segmentSize, {
-          concurrency: options.downloadConcurrency
-        })
+        downloadProgress.nextSegment(segmentSize)
```
aiqiaoy, line 257: Shouldn't this be after the download call?

packages/cache/src/internal/downloadUtils.ts
```diff
@@ -26,6 +27,139 @@ async function pipeResponseToStream(
...
+    }
+  }
+
+  /**
+   * Callback used to update the progress.
+   *
+   * @param progress the progress event
+   */
+  onProgressCallback(progress: TransferProgressEvent): void {
+    this.setReceivedBytes(progress.loadedBytes)
```
aiqiaoy, line 128: Nit: Maybe `return (progress: TransferProgressEvent) => { this.setReceivedBytes(progress.loadedBytes) }` and then we can do
``` 
       const result = await client.downloadToBuffer(
          downloadProgress.segmentOffset,
          segmentSize,
          {
            concurrency: options.downloadConcurrency,
            onProgress: downloadProgress.onProgress()
          }
```

dhadka: Fixed.

packages/cache/src/internal/downloadUtils.ts
dhadka: No, it needs to be called before so `downloadProgress` knows the current `segmentSize`.

aiqiaoy:
toolkit-audit check failure is a known issue: https://github.slack.com/archives/D010RN1245D/p1594736728009600

PR #515, opened 2020-07-13
Pipe audit results to a json file so lerna does not overflow

thboop:
Our audit command is currently failing due to a lerna overflow issue. There's an open issue in lerna [here](https://github.com/lerna/lerna/issues/2384). For now, we should pipe the stdout to a json file, and have stderr display the results of the run.

### DIFF
--- a/packages/artifact/package.json
+++ b/packages/artifact/package.json
@@ -29,7 +29,7 @@
     "directory": "packages/artifact"
   },
   "scripts": {
-    "audit-moderate": "npm install && npm audit --audit-level=moderate --json > audit.json",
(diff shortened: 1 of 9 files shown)


packages/artifact/package.json
```diff
@@ -29,7 +29,7 @@
     "directory": "packages/artifact"
   },
   "scripts": {
-    "audit-moderate": "npm install && npm audit --audit-level=moderate",
+    "audit-moderate": "npm install && npm audit --audit-level=moderate --json > audit.json",
```
ericsciple, line 32: super nit: usually flags fist

PR #520, opened 2020-07-15
Fix Issue where we can no longer create zip files on windows during cli tests.

thboop:
The new virtual environment shipped for windows today which resulted in a change where our powershell script could no longer create the directory on windows. We should just move to pwsh (powershell core) as there is more support for that shell.


packages/tool-cache/__tests__/tool-cache.test.ts
```diff
@@ -553,7 +553,7 @@ describe('@actions/tool-cache', function() {
       if (IS_WINDOWS) {
         const escapedStagingPath = stagingDir.replace(/'/g, "''") // double-up single quotes
         const escapedZipFile = zipFile.replace(/'/g, "''")
-        const powershellPath = await io.which('powershell', true)
```
ericsciple, line 556: consider `await io.which('pwsh', false) || await io.which('powershell', true)`

mishu017:
[]()

packages/tool-cache/__tests__/tool-cache.test.ts
mishu017: #- [__]()

mishu017: No []()

PR #527, opened 2020-07-20
Set main as the default branch

thboop:
Resolves: #514

### DIFF
--- a/docs/action-types.md
+++ b/docs/action-types.md
@@ -32,14 +32,14 @@ jobs:
         os: [ubuntu-16.04, windows-2019]
     runs-on: ${{matrix.os}}
     actions:
-    - uses: actions/setup-node@main
+    - uses: actions/setup-node@v1
       with:
         version: ${{matrix.node}}
     - run: | 
         npm install
     - run: |
         npm test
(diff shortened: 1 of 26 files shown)


docs/action-types.md
```diff
@@ -32,14 +32,14 @@ jobs:
         os: [ubuntu-16.04, windows-2019]
     runs-on: ${{matrix.os}}
     actions:
-    - uses: actions/setup-node@master
+    - uses: actions/setup-node@main
```
ericsciple, line 35: would be better to pin to v1 or v2

packages/tool-cache/src/tool-cache.ts
```diff
@@ -551,7 +551,7 @@ export async function getManifestFromRepo(
   owner: string,
   repo: string,
   auth?: string,
-  branch = 'master'
+  branch = 'main'
```
ericsciple, line 554: check w/ @bryanmacfarlane

thboop: Going to pull this change out and address it in another issue. We should probably default to the default branch of a repo, or force a user to specify, but I don't want to make breaking changes here right now.

PR #530, opened 2020-07-23
Allow specifying arbitrary headers when downloading tools to the tool cache.

chrisgavin:
This change allows specifying arbitrary headers when downloading items to the toolcache.

This is useful for some URLs that require things like a particular `Accept` header to be present to deliver the right content.


thboop (approved):
LGTM thanks for the contribution @chrisgavin

PR #533, opened 2020-07-29
Use posix archive format

smorimoto:
See #475 for details.

Closes: https://github.com/actions/cache/issues/326

### DIFF
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@actions/cache",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "preview": true,
   "description": "Actions cache lib",
   "keywords": [
--- a/packages/cache/src/internal/tar.ts
+++ b/packages/cache/src/internal/tar.ts
@@ -101,6 +101,7 @@ export async function createTar(
     }
   }
   const args = [
+    '--posix',
     ...getCompressionProgram(),
     '-cf',
     cacheFileName.replace(new RegExp(`\\${path.sep}`, 'g'), '/'),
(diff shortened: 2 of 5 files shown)


smorimoto:
@aiqiaoy This should now be in a mergeable state. Could you review it?

smorimoto:
CC: @joshmgross

smorimoto:
Related issue: https://github.com/actions/cache/issues/326

packages/cache/src/internal/tar.ts
```diff
@@ -101,6 +101,7 @@ export async function createTar(
     }
   }
   const args = [
+    '--posix',
```
konradpabjan, line 104: I looked over the previous PR, I _think_ this won't cause any compat issues. No objections from me 👍

packages/cache/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/cache",
-  "version": "1.0.1",
+  "version": "1.1.0",
```
konradpabjan, line 3: This is a pretty small change so I think a patch bump to version `1.0.2` is sufficient

Could you also update `package-lock.json`? There is a version number there as well.

smorimoto: Oh indeed.

smorimoto: It was changed with the latest commit.

packages/cache/src/internal/tar.ts
smorimoto: Yeah, this shouldn't cause any problems.

smorimoto:
@konradpabjan Thanks for your review.

aiqiaoy:
Yeah sorry I was on-call when I reviewed the original PR and completely forgot about following up. Looks good!

konradpabjan:
Version `1.0.2` has been published: https://www.npmjs.com/package/@actions/cache/v/1.0.2

smorimoto:
@konradpabjan Thanks again for your work :)

smorimoto:
@aiqiaoy No worries! I sometimes do something similar to you.

PR #535, opened 2020-07-30
Better Artifact Telemetry + Increase Upload chunk size

konradpabjan:
When making any http calls during artifact upload or download, the user-agent string can be used to help differentiate between http-calls that get logged as part of internal telemetry. This PR changes the user-agent string to `@actions/artifact-upload` for any call that gets made during artifact upload and `@actions/artifact-download` for any call that gets made during download.

I really wanted to leverage the package version from `package.json` and then have the user-agent string be something like `@actions/artifact-0.3.2-upload` however there are a host of issues reading the version number from `package.json` since it's not included in the `rootDir` folder. More information here: https://stackoverflow.com/a/53836076

I'm also bumping up the upload chunk size from 4MB to 8MB to help improve artifact upload in GHES scenarios


micha1le617:
Audit

PR #540, opened 2020-08-04
@actions/artifact Retry on 413 response

konradpabjan:
Client side fix to: https://github.com/github/c2c-actions-service/issues/1337


packages/artifact/src/internal/utils.ts
```diff
@@ -74,7 +74,8 @@ export function isRetryableStatusCode(statusCode?: number): boolean {
     HttpCodes.BadGateway,
     HttpCodes.ServiceUnavailable,
     HttpCodes.GatewayTimeout,
-    HttpCodes.TooManyRequests
+    HttpCodes.TooManyRequests,
+    413 // Payload Too Large
```
konradpabjan, line 78: Hardcoded for now. Will update in the future once a new version of `actions/http-client` is out: https://github.com/actions/http-client/pull/31

PR #548, opened 2020-08-06
Add License.md to all npm packages

### LINKED ISSUE #546: Packages are published without license text
Lerna does not appear to currently distribute the license to each individual package, ex: https://github.com/lerna/lerna/issues/1213.

A lot of our packages are being published without the license file included, we should add this file to each package to ensure we correctly publish the license file.

thboop:
Resolves #546

### DIFF
--- a/packages/artifact/LICENSE.md
+++ b/packages/artifact/LICENSE.md
@@ -1,0 +1,7 @@
+Copyright 2019 GitHub
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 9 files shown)


packages/artifact/LICENSE.md
```diff
@@ -0,0 +1,7 @@
+Copyright 2019 GitHub
```
konradpabjan, line 1: Shouldn't there be a line with `The MIT License (MIT)` at the top of each file?

Our licenses for some other repos have it included: https://github.com/actions/download-artifact/blob/c3f5d00c8784369c43779f3d2611769594a61f7a/LICENSE#L2

thboop: This should be the license text, it is taken from the root license file in this repo.

https://opensource.org/licenses/MIT

There's no harm in adding that extra line, but I don't see a reason to.

konradpabjan (approved):

Marudhas (approved):

Marudhas (approved):

Jircs1 (commented):
Approve change reviews

Issue #552, opened 2020-08-10
Select GNU tar for caching if available on hosted runners


dhadka:
**Describe the bug**

BSD tar has been problematic for the caching action for several reasons:

1. `bsdtar` does not handle recursive directory structures created by junction points - https://github.com/actions/cache/issues/315

2. `bsdtar` does not work with `zstd` compression, causing the process to hang - https://github.com/actions/cache/issues/301

On hosted runners, `tar.exe` is available in several locations locations:

1. BSD tar - `C:\Windows\System32\tar.exe`  (Windows 2019 only)
2. GNU tar - `C:\Program Files\Git\usr\bin\tar.exe`
3. GNU tar - `C:\msys64\usr\bin\tar.exe`

Caching correctly selects GNU tar on Windows 2016 runners, since (2) shows up on the PATH.  Windows 2019 runners use BSD tar because (1) shows up first on the PATH.  (3) does not appear on the system PATH so it is never selected.

Given the problems with `bsdtar`, it seems we should either ensure GNU tar appears first on the system PATH or improve the lookup logic in the caching module to find GNU tar. Right now, a workaround is to explicitly set the system PATH:

```
    - name: Use GNU Tar from Msys
      if: ${{ runner.os == 'Windows' }}
      run: |
        echo "::add-path::C:\msys64\usr\bin"
```

**Expected behavior**
GNU tar is selected if it is available.

**Desktop (please complete the following information):**
 - OS: Windows

eregon:
In https://github.com/ruby/setup-ruby/issues/102 we noticed that the MSYS2 tar doesn't seem to work with the given arguments.
Probably, the list of paths needs to be `/` paths and not `\` paths.

PR #553, opened 2020-08-10
Prefer using GNU tar

dhadka:
Fixes a variety of [actions/cache](https://github.com/actions/cache) issues related to differences between BSD and GNU tar.

The old `which` function already scanned the system path for the executable, but would return the first hit.  This change moves most of this functionality into a new method, `findInPath`, which will return a list of all matches on the system path.  This allows the cache module to scan all available `tar` implementations to find GNU tar.

Fixes https://github.com/actions/toolkit/issues/552
Fixes https://github.com/actions/cache/issues/362 - OS-agnostic caches miss on Windows due to different compression (caused by zstd issue)
Fixes https://github.com/actions/cache/issues/315 - infinite recursion in `bsdtar` caused by junction points on windows
Fixes https://github.com/actions/cache/issues/301 - `zstd` hangs when used with `bsdtar`

### DIFF
--- a/packages/cache/src/internal/cacheUtils.ts
+++ b/packages/cache/src/internal/cacheUtils.ts
@@ -64,7 +64,7 @@ async function getVersion(app: string): Promise<string> {
   core.debug(`Checking ${app} --version`)
   let versionOutput = ''
   try {
-    await exec.exec(`${app} --version`, [], {
+    await exec.exec(`"${app}" --version`, [], {
       ignoreReturnCode: true,
       silent: true,
       listeners: {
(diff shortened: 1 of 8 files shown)


packages/io/src/io.ts
```diff
@@ -192,69 +188,89 @@ export async function which(tool: string, check?: boolean): Promise<string> {
...
+    return ''
+  } catch (err) {
+    throw new Error(`which failed with message ${err.message}`)
+  }
+}
 
-      if (filePath) {
-        return filePath
-      }
+/**
```
dhadka, line 208: Most of the `which` method is now in `findInPath`.  Most of the changes are just indentation...use the "hide whitespace changes" option for easier viewing.

smorimoto:
I love this change 🙂

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -64,7 +64,7 @@ async function getVersion(app: string): Promise<string> {
   core.debug(`Checking ${app} --version`)
   let versionOutput = ''
   try {
-    await exec.exec(`${app} --version`, [], {
+    await exec.exec(`"${app}" --version`, [], {
```
smorimoto, line 67:
```suggestion
    await exec.exec(app, ['--version'], {
```

smorimoto:
@dhadka Is there any progress on this? If you are busy, I can take over this.

dhadka:
Closing for https://github.com/actions/toolkit/pull/610

PR #556, opened 2020-08-13
Add workspace to GitHub context

### LINKED ISSUE #555: Missing github context `workspace` property
**Describe the bug**

The github context's `workspace` property is not exposed from context.ts:

https://github.com/actions/toolkit/blob/1cc56db0ff126f4d65aeb83798852e02a2c180c3/packages/github/src/context.ts#L10-L20

**Expected behavior**

Per [the docs](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) I expect to be able to access the `workspace` property on a github context.

AArnott:
Fixes #555

### DIFF
--- a/packages/github/src/context.ts
+++ b/packages/github/src/context.ts
@@ -18,6 +18,7 @@ export class Context {
   job: string
   runNumber: number
   runId: number
+  workspace: string
 
   /**
    * Hydrate the context from the environment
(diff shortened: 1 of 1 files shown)


AArnott:
Note: I have *not* tested this.

thboop:
This value isn't really that useful in the `@actions/github` module. It may be better to expose it in `core` or `io`, so we don't need to import the `actions/github` lib just to get to this value when doing file operations. The `actions/github` context is really just for items that you may want to use in the octokit apis, a local file path doesn't really fit in.

 Thanks for the pr, but I don't think this is the correct approach. I'm going to close this out.

Issue #537, opened 2020-08-02
"Cache already exists" due to concurrency


Vampire:
If you have a look at this workflow run: https://github.com/Vampire/setup-wsl/runs/938421512?check_suite_focus=true you see the problem.
Here the timeline of the problem:

```
22:30:04 Job `Test 'Debian' distribution` / Step ` Execute action`: ##[debug]Resolved Keys
22:30:04 Job `Test 'Debian' distribution` / Step ` Execute action`: Cache not found => download and later add to cache
22:30:08 Job `Test additional packages for 'Debian' distribution` / Step `Execute action`: ##[debug]Resolved Keys:
22:30:08 Job `Test additional packages for 'Debian' distribution` / Step `Execute action`: Cache not found => download and later add to cache
22:30:09 Job `Test 'Debian' distribution` / Step ` Execute action`: ##[debug]Reserving Cache
22:30:11 Job `Test additional packages for 'Debian' distribution` / Step ` Execute action`: ##[debug]Reserving Cache
22:30:11 Job `Test additional packages for 'Debian' distribution` / Step ` Execute action`: ##[error]reserveCache failed
22:30:14 Job `Test 'Debian' distribution` / Step ` Execute action`: Cache saved successfully
```

The problem is clear, the typical problem you have with lazy initialization if you don't do the double-checked locking properly.
Both jobs see the cache is not present, prepare the thing to be cached and then try to cache it, while the second fails at caching.

Is there a clean way to solve this, or can something be added like for example a method `cache.exists()` that only checks whether the cache exists? Doing a full `restore` before doing the `save` seems a bit overkill just to check whether in the meantime another job already filled the cache. And just catching the exception feels like "exception for flow control" if there could maybe be a simple test method to check up front.

JojOatXGME:
I created my own action and I seem to get a wanring caused by this error quite reliably on every build (if the cache was not already present before I started it). It is annoying.

    ##[warning]Error: reserveCache failed: Cache already exists. Scope: refs/heads/***, Key: ***, Version: ***

If we look at the `actions/cache` action, they already have implemented a special case to prevent such warnings. I did the same in my action.

```js
// https://github.com/actions/cache/blob/d29c1df198dd38ac88e0ae23a2881b99c2d20e68/src/save.ts#L43
} catch (error) {
    if (error.name === cache.ValidationError.name) {
        throw error;
    } else if (error.name === cache.ReserveCacheError.name) {
        core.info(error.message);
    } else {
[...]

PR #558, opened 2020-08-17
Handle errors representing non-successful http responses in retry logic

dhadka:
Fixes cache to read status code contained in an `HttpClientError` thrown by `http-client` and handle retries correctly.  Fixes https://github.com/actions/toolkit/issues/537

### Why?

If using one of the standard `get`, `put`, `post`, etc. methods in `http-client`, it returns an `IHttpClientResponse` that contains a status code.  To check if the request was successful or not, we must check the value of the status code.

If using one of the JSON methods (`getJson`, `putJson`, `postJson`, etc.), it will return an `ITypedResponse<T>` if successful.  If the status code is >= 300, it will throw an Error...except for 404 which will return an `ITypedResponse<T>` with a `null` result.

At some point, we updated the cache to use the JSON methods for several calls, but never updated the logic to expect an `Error` instead of a status code.  As a result, the retry logic is not correctly handling errors.  Specifically, errors such as a 409 conflict are not retryable.  However, we treat a generic `Error` as retryable, as it could indicate a socket / network issue, and ultimately retry calls that will never succeed.

### Fix

This PR adds an optional `onError` handler to the `retry` method, which allows it to convert the `Error` object it receives from `http-client` back into a `ITypedResponse<T>`.  The `ITypedResponse<T>` is populated with the status code contained in the `Error` object.  All other fields are left empty (`null`).  This lets the existing `isSuccessfulStatusCode` checks used by the cache module to work as expected.

### DIFF
--- a/packages/cache/src/internal/requestUtils.ts
+++ b/packages/cache/src/internal/requestUtils.ts
@@ -4,6 +4,7 @@ import {
   IHttpClientResponse,
   ITypedResponse
 } from '@actions/http-client/interfaces'
+import {RetryDelay} from './constants'
 
 export function isSuccessStatusCode(statusCode?: number): boolean {
   if (!statusCode) {
@@ -31,11 +32,16 @@ export function isRetryableStatusCode(statusCode?: number): boolean {
   return retryableStatusCodes.includes(statusCode)
 }
 
+async function sleep(milliseconds: number): Promise<void> {
+  return new Promise(resolve => setTimeout(resolve, milliseconds))
+}
+
 export async function retry<T>(
   name: string,
   method: () => Promise<T>,
   getStatusCode: (arg0: T) => number | undefined,
   maxAttempts = 2,
+  delay = RetryDelay,
   onError: ((arg0: Error) => T | undefined) | undefined = undefined
 ): Promise<T> {
   let errorMessage = ''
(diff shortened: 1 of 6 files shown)


packages/cache/tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "baseUrl": "./",
     "outDir": "./lib",
     "rootDir": "./src",
-    "lib": ["es6", "dom"]
+    "lib": ["es6", "dom"],
+    "strict": false
```
dhadka, line 8: This is needed to read the `statusCode` property contained in errors:

```
error['statusCode']
```

packages/cache/src/internal/requestUtils.ts
```diff
@@ -31,32 +32,48 @@ export function isRetryableStatusCode(statusCode?: number): boolean {
...
+
     try {
       response = await method()
+    } catch (error) {
+      if (onError) {
+        response = onError(error)
+      }
+
+      isRetryable = true
+      errorMessage = error.message
```
aiqiaoy, line 63: Nit: Should we log this errorMessage here since it gets converted to a generic `Cache service responded with ${statusCode}` below if statusCode is set?

aiqiaoy (approved):
👍

joshmgross (approved):

3e311 (approved):

PR #571, opened 2020-09-11
Add File Commands

thboop:


labqurzor1:
Fusionando

docs/commands.md
```diff
@@ -155,8 +124,90 @@ function setCommandEcho(enabled: boolean): void {}
...
+```javascript
+export function exportVariable(name: string, val: string): void {}
+```
+
+### PATH Manipulation
+
+To prepend a string to PATH write to the file located at `GITHUB_PATH` or use the equivalent `actions/core` function
+
+```sh
+echo "foo=bar" >> $GITHUB_PATH
```
ericsciple, line 197: consider something that looks more like a real path, for example `echo '/some/directory/path' >> $GITHUB_PATH`

packages/core/src/core.ts
```diff
@@ -39,7 +41,15 @@ export enum ExitCode {
 export function exportVariable(name: string, val: any): void {
   const convertedVal = toCommandValue(val)
   process.env[name] = convertedVal
-  issueCommand('set-env', {name}, convertedVal)
+
+  const filePath = process.env['GITHUB_ENV'] || ''
+  if (filePath && filePath.length > 0) {
```
ericsciple, line 46: nit: truthy check is sufficient (ensures length \> 0)

docs/commands.md
```diff
@@ -155,8 +124,73 @@ function setCommandEcho(enabled: boolean): void {}
...
+
 CMD processes the `"` character differently from other shells when echoing. In CMD, the above snippets should have the `"` characters removed in order to correctly process. For example, the set output command would be:
 ```cmd
 echo ::set-output name=FOO::BAR
 ```
+
+
+# File Commands
+
+During the execution of a workflow, the runner generates temporary files that you can write to to perform certain actions. The path to these files are exposed via environment variables. You will need to use the `utf-8` encoding when writing to these files to ensure proper processing of the commands. Multiple commands can be written to the same file, seperated by newlines.
```
ericsciple, line 137: consider changing from

> to to

to 😉

> the runner generates temporary files that can be used to perform certain actions.

ericsciple, line 137: sp: \"seperated\" -\> \"separated\"

docs/commands.md
```diff
@@ -155,8 +124,73 @@ function setCommandEcho(enabled: boolean): void {}
...
-### Command Prompt 
+### Command Prompt
+
 CMD processes the `"` character differently from other shells when echoing. In CMD, the above snippets should have the `"` characters removed in order to correctly process. For example, the set output command would be:
 ```cmd
 echo ::set-output name=FOO::BAR
 ```
+
+
+# File Commands
```
ericsciple, line 135: nit: should we call these something like \"Environment files\"

packages/core/src/command.ts
```diff
@@ -76,19 +77,6 @@ class Command {
...
 }
 
-/**
- * Sanitizes an input into a string so it can be passed into issueCommand safely
- * @param input input to sanitize into a string
- */
-export function toCommandValue(input: any): string {
-  if (input === null || input === undefined) {
-    return ''
-  } else if (typeof input === 'string' || input instanceof String) {
```
akkim31, line 86: []()

akkim31 (approved):

akkim31 (approved):
run formet

641210 (commented):
- ****

Hbutlercapone (commented):
Added File Commands

Tuneabq (commented):
✅

PR #575, opened 2020-09-16
Add an option to specify retention days for artifacts

yacaovsnc:
Introducing an option to reduce the default retention period for uploaded artifacts.

Today all artifacts have a 90 days retention value and there is no knob to reduce the value.  We are revamping the experience so we can set custom expiry on individual artifact.

The new minimum value is `1` day.

### DIFF
--- a/packages/artifact/README.md
+++ b/packages/artifact/README.md
@@ -39,6 +39,10 @@ Method Name: `uploadArtifact`
     - If set to `false`, and an error is encountered, all other uploads will stop and any files that were queued will not be attempted to be uploaded. The partial artifact available will only include files up until the failure.
     - If set to `true` and an error is encountered, the failed file will be skipped and ignored and all other queued files will be attempted to be uploaded. There will be an artifact available for download at the end with everything excluding the file that failed to upload
     - Optional, defaults to `true` if not specified
+- `retentionDays`
+    - Durantion after which artifact will expire in days
+    - Minimum value: 1
+    - Maximum value: 90 unless changed by settings
 
 #### Example using Absolute File Paths
 
--- a/packages/artifact/README.md
+++ b/packages/artifact/README.md
@@ -43,7 +43,7 @@ Method Name: `uploadArtifact`
     - Duration after which artifact will expire in days
     - Minimum value: 1
     - Maximum value: 90 unless changed by repository setting
-    - If this is set to a greater value than the retention settings allowed, the retention on artifacts will be reduced to match the max value allowed on server, and the upload process will continue. An input of 0 assumes default retention value.
+    - If this is set to a greater value than the retention settings allowed, the retention on artifacts will be reduced to match the max value allowed on the server, and the upload process will continue. An input of 0 assumes default retention value.
 
 #### Example using Absolute File Paths
 
--- a/packages/artifact/README.md
+++ b/packages/artifact/README.md
@@ -40,9 +40,10 @@ Method Name: `uploadArtifact`
     - If set to `true` and an error is encountered, the failed file will be skipped and ignored and all other queued files will be attempted to be uploaded. There will be an artifact available for download at the end with everything excluding the file that failed to upload
     - Optional, defaults to `true` if not specified
 - `retentionDays`
-    - Durantion after which artifact will expire in days
+    - Duration after which artifact will expire in days
     - Minimum value: 1
-    - Maximum value: 90 unless changed by settings
+    - Maximum value: 90 unless changed by repository setting
+    - If this is set to a greater value than the retention settings allowed, the retention on artifacts will be reduced to match the max value allowed on server, and the upload process will continue. An input of 0 assumes default retention value.
 
 #### Example using Absolute File Paths
 
--- a/packages/artifact/__tests__/upload.test.ts
+++ b/packages/artifact/__tests__/upload.test.ts
@@ -108,7 +108,7 @@ describe('Upload Tests', () => {
   })
 
   it('Create Artifact - Retention Less Than Min Value Error', async () => {
-    const artifactName = 'invalid-artifact-name'
+    const artifactName = 'valid-artifact-name'
     const options: UploadOptions = {
       retentionDays: -1
     }
--- a/packages/artifact/__tests__/upload.test.ts
+++ b/packages/artifact/__tests__/upload.test.ts
@@ -13,6 +13,7 @@ import {
 } from '../src/internal/contracts'
 import {UploadSpecification} from '../src/internal/upload-specification'
 import {getArtifactUrl} from '../src/internal/utils'
+import {UploadOptions} from '../src/internal/upload-options'
 
 const root = path.join(__dirname, '_temp', 'artifact-upload')
 const file1Path = path.join(root, 'file1.txt')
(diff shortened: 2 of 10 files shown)


packages/artifact/src/internal/config-variables.ts
```diff
@@ -61,3 +61,7 @@ export function getWorkSpaceDirectory(): string {
   }
   return workspaceDirectory
 }
+
+export function getRetentionDays(): string | undefined {
+  return process.env['GITHUB_RETENTION_DAYS']
```
yacaovsnc, line 66: https://github.com/actions/runner/pull/714

packages/artifact/src/internal/upload-options.ts
```diff
@@ -15,4 +15,22 @@ export interface UploadOptions {
    *
    */
   continueOnError?: boolean
+
+  /**
+   * Durantion after which artifact will expire in days.
```
konradpabjan, line 20: Nit spelling

packages/artifact/README.md
```diff
@@ -39,6 +39,10 @@ Method Name: `uploadArtifact`
     - If set to `false`, and an error is encountered, all other uploads will stop and any files that were queued will not be attempted to be uploaded. The partial artifact available will only include files up until the failure.
     - If set to `true` and an error is encountered, the failed file will be skipped and ignored and all other queued files will be attempted to be uploaded. There will be an artifact available for download at the end with everything excluding the file that failed to upload
     - Optional, defaults to `true` if not specified
+- `retentionDays`
+    - Durantion after which artifact will expire in days
+    - Minimum value: 1
+    - Maximum value: 90 unless changed by settings
```
konradpabjan, line 45: What will happen if the user puts in a negative value as input? Would it be worth it to have some input validation?

I would expand a little bit on the "changed by settings". Who can change these settings, and where can this be done? You put some more information in `upload-options.ts` and I think it would be good to put as much information in the `README` since users will look at this first.

packages/artifact/src/internal/__mocks__/config-variables.ts
```diff
@@ -45,3 +45,7 @@ export function getRuntimeUrl(): string {
 export function getWorkFlowRunId(): string {
   return '15'
 }
+
+export function getRetentionDays(): string | undefined {
```
konradpabjan, line 49: I would suggest adding a test that uses this mocked value.

The entire POST call to create an artifact is mocked so you can use the existing code to simulate a response from the actions service and add some verification to see if the correct value is going through: https://github.com/actions/toolkit/blob/main/packages/artifact/__tests__/upload.test.ts#L373:L410

Here is an existing test where you would just need to add a new `options` input: https://github.com/actions/toolkit/blob/main/packages/artifact/__tests__/upload.test.ts#L81:L83

packages/artifact/src/internal/upload-http-client.ts
```diff
@@ -55,12 +56,30 @@ export class UploadHttpClient {
...
+
+    // calculate retention period
+    if (options && options.retentionDays) {
+      let retention = options.retentionDays
+      const maxRetentionStr = getRetentionDays()
+      if (maxRetentionStr) {
+        const maxRetention = parseInt(maxRetentionStr)
+        if (!isNaN(maxRetention) && maxRetention < retention) {
+          core.warning(
+            `Retention days is greater than max value allowed by repository setting, reduce retention to ${maxRetention} days`
```
konradpabjan, line 75: Nit: `Retention days is greater than the max value allowed by the repository setting`

konradpabjan, line 75: Nit `Retention days is greater than the max value allowed by the repository setting`

packages/artifact/src/internal/upload-options.ts
```diff
@@ -15,4 +15,22 @@ export interface UploadOptions {
...
+   * Use this option to override the default expiry.
+   *
+   * Min value: 1
+   * Max value: 90 unless changed by repository setting
+   *
+   * If this is set to a greater value than the retention settings allowed, the retention on artifacts
+   * will be reduced to match the max value allowed on server, and the upload process will continue. An
+   * input of 0 assumes default retention setting.
+   */
+
```
konradpabjan, line 34: Nit: blank spacing. If a multi line comment is used right above a variable, it makes the whole comment available via intelisense if someone is using something like vscode

konradpabjan (commented):
Overall looks great 👍 Just a few small/nit issues.

packages/artifact/src/internal/upload-http-client.ts
yacaovsnc: You don't want to tell the user what the new retention value is set to?  This is the only place user can find out the retention period as of now since we haven't implement any UI change to show artifact expiry.

konradpabjan: Yes, keep the: `reduce retention to ${maxRetention} days` text at the end, I was mainly nit-picking the missing `the` at the beginning of the warning

konradpabjan:
One more thing, we will need to release a new version of the package so you need to update
- RELEASES.md
- package.json
- package-lock.json

packages/artifact/src/internal/__mocks__/config-variables.ts
yacaovsnc: Good suggestion!  I refactored the code to make unit tests easier.  Now all the logic to compute the retention value is moved to a util function so we can test it independently.

packages/artifact/__tests__/upload.test.ts
```diff
@@ -106,6 +107,17 @@ describe('Upload Tests', () => {
     )
   })
 
+  it('Create Artifact - Retention Less Than Min Value Error', async () => {
+    const artifactName = 'invalid-artifact-name'
```
konradpabjan, line 111:
```suggestion
    const artifactName = 'valid-artifact-name'
```

The current name can cause another error to be thrown (there is another test for the artifact name), best to rule that out and use a valid name.

packages/artifact/README.md
```diff
@@ -39,6 +39,11 @@ Method Name: `uploadArtifact`
     - If set to `false`, and an error is encountered, all other uploads will stop and any files that were queued will not be attempted to be uploaded. The partial artifact available will only include files up until the failure.
     - If set to `true` and an error is encountered, the failed file will be skipped and ignored and all other queued files will be attempted to be uploaded. There will be an artifact available for download at the end with everything excluding the file that failed to upload
     - Optional, defaults to `true` if not specified
+- `retentionDays`
+    - Duration after which artifact will expire in days
+    - Minimum value: 1
+    - Maximum value: 90 unless changed by repository setting
+    - If this is set to a greater value than the retention settings allowed, the retention on artifacts will be reduced to match the max value allowed on server, and the upload process will continue. An input of 0 assumes default retention value.
```
konradpabjan, line 46:
```suggestion
    - If this is set to a greater value than the retention settings allowed, the retention on artifacts will be reduced to match the max value allowed on the server, and the upload process will continue. An input of 0 assumes default retention value.
```

packages/artifact/src/internal/utils.ts
```diff
@@ -302,3 +302,24 @@ export async function createEmptyFilesForArtifact(
...
+  if (retentionInput < 0) {
+    throw new Error('Invalid retention, minimum value is 1.')
+  }
+
+  let retention = retentionInput
+  if (retentionSetting) {
+    const maxRetention = parseInt(retentionSetting)
+    if (!isNaN(maxRetention) && maxRetention < retention) {
+      warning(
+        `Retention days is greater than max value allowed by the repository setting, reduce retention to ${maxRetention} days`
```
konradpabjan, line 319:
```suggestion
        `Retention days is greater than the max value allowed by the repository setting, reduce retention to ${maxRetention} days`
```

packages/artifact/__tests__/util.test.ts
```diff
@@ -106,6 +106,20 @@ describe('Utils', () => {
     }
   })
 
+  it('Test negative artifact retention throws', () => {
+    expect(() => {
+      utils.getProperRetention(-1, undefined)
+    }).toThrow()
+  })
+
+  it('Test no setting specified take artifact retention inpput', () => {
```
konradpabjan, line 115:
```suggestion
  it('Test no setting specified takes artifact retention input', () => {
```

konradpabjan (approved):
Thanks for adding the tests! Just a few more small suggestions. 

A separate PR for updating the release works! 🚀

Marudhas (approved):

PR #580, opened 2020-09-22
Pass correct working directory from `cache` to `glob`.

reitermarkus:
For testing Homebrew, we are re-using the Homebrew installation already present on the worker machines. For this we replace `GITHUB_WORKSPACE` with a symlink to the Homebrew installation.

This leads to the case where `GITHUB_WORKSPACE` is different from the current working directory, which breaks caching because the cache is created relative to the `GITHUB_WORKSPACE`, however the globbed paths are always relative to the current working directory.

This pull request ensures that globbed paths are also relative to `GITHUB_WORKSPACE` in the `cache` action.

### DIFF
--- a/packages/cache/src/internal/cacheUtils.ts
+++ b/packages/cache/src/internal/cacheUtils.ts
@@ -39,12 +39,20 @@ export function getArchiveFileSizeIsBytes(filePath: string): number {
   return fs.statSync(filePath).size
 }
 
+export function getWorkingDirectory(): string {
+  return process.env['GITHUB_WORKSPACE'] ?? process.cwd()
+}
+
 export async function resolvePaths(patterns: string[]): Promise<string[]> {
   const paths: string[] = []
-  const workspace = process.env['GITHUB_WORKSPACE'] ?? process.cwd()
-  const globber = await glob.create(patterns.join('\n'), {
-    implicitDescendants: false
-  })
+  const workspace = getWorkingDirectory()
+  const globber = await glob.create(
+    patterns.join('\n'),
+    {
+      implicitDescendants: false
+    },
+    workspace
+  )
 
   for await (const file of globber.globGenerator()) {
     const relativeFile = path.relative(workspace, file)
--- a/packages/cache/src/internal/tar.ts
+++ b/packages/cache/src/internal/tar.ts
@@ -37,16 +37,12 @@ async function execTar(
   }
 }
 
-function getWorkingDirectory(): string {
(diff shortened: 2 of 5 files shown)


reitermarkus:
Can someone review this? I guess this needs to be split up to change `glob` and `cache` separately?

reitermarkus:
Ping.

reitermarkus:
Ping.

thboop:
Hey @reitermarkus, we don't currently intend to encourage or support users changing the GITHUB_WORKSPACE. Even if we were to merge this PR, other things would not work correctly, such as the context value `{{github.workspace}}` would not correctly map to the new directory if you change the workspace. If the ability to change the location of the `GITHUB_WORKSPACE` would be useful to you, please file an enhancement issue describing your use case in this repo.

Depending on your use case and platform, rather than changing the workspace on your self hosted runner, you may be able to set up a sym link from that path to the github workspace, or manually copy those files as a step in your workflow.

reitermarkus:
> other things would not work correctly, such as the context value `{{github.workspace}}` would not correctly map to the new directory if you change the workspace

I don't see how these changes would break this. Can you elaborate what exactly would be broken by this? If you don't change `GITHUB_WORKSPACE`, everything should behave exactly the same as before.

thboop:
So, we originally supported the ability for some first party actions to move the `github_workspace`. `actions/checkout@v1` did, [here's the adr on why we moved away from it](https://github.com/actions/checkout/blob/main/adrs/0153-checkout-v2.md#path). It ultimately resulted in a lot of confusion for users that we want to avoid.

Even if we merged this, changing the GITHUB_WORKSPACE would still not work as expected.
- If you tried to use `echo "GITHUB_WORKSPACE=newpath" >> $GITHUB_ENV` to [set an environment variable](https://github.com/actions/toolkit/blob/main/docs/commands.md#set-an-environment-variable), it would just not set it
- If you tried to modify the variable for a single action or step, for example with a step like
```
    - run: |
        echo $GITHUB_WORKSPACE
        GITHUB_WORKSPACE=newpath
        echo "the context value is: ${{github.workspace}}"
        echo "the env value is $GITHUB_WORKSPACE"
```
You would see the env value changes but not the context value. Actions can have these values passed in as an input.

That be said, we plan on implementing a more generic version of this in: https://github.com/actions/toolkit/issues/750. Not specifically based off the working directory, but allowing you to pass a pattern root into glob and have it use that.

For now, I'm going to close this PR @reitermarkus. Feel free to ping me and we can reopen as needed.

PR #588, opened 2020-09-30
Expose method that checks if a key exists in the cache

bowd:
Expose a helper method that only checks for a key in the cache without downloading.
This will help us avoid using internals in the implementation here: https://github.com/actions/cache/pull/420.
The reasoning behind the need for this is also detailed in that PR.

### DIFF
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -44,6 +44,33 @@ function checkKey(key: string): void {
 }
 
 /**
+ * Check if cache contains key
+ *
+ * @param paths a list of file paths to restore from the cache
+ * @param primaryKey an explicit key for restoring the cache
+ * @returns boolean if the key and paths combo exists in the cache
+ */
+export async function containsKey(
+    paths: string[],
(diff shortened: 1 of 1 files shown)


packages/cache/src/cache.ts
```diff
@@ -43,6 +43,33 @@ function checkKey(key: string): void {
...
 
+/**
+ * Check if cache contains key
+ *
+ * @param paths a list of file paths to restore from the cache
+ * @param primaryKey an explicit key for restoring the cache
+ * @returns boolean if the key and paths combo exists in the cache
+ */
+export async function containsKey(
+    paths: string[],
```
vantuan1994, line 54: Kiểm tra

vantuan1994 (commented):
Ok

parched:
What's the status of this PR? Looks like some tests are failing, do you plan to update them @bowd. If not, I'd be happy to look into it?

One question I have about this change though, does getting the cache entry update the accessed time? If this is called, can we assume it won't be evicted shortly after?

PR #594, opened 2020-10-03
Get linux version from os-release file if available

ukstv:
Not all Linux distributions have `/etc/lsb-release` file available. If you try to use local runner like https://github.com/nektos/act it fails to perform actions relying on the toolkit. Say [setup-python](https://github.com/actions/setup-python) action can not proceed because of `/etc/lsb-release` missing. It can not determine version of python distribution to setup for a void os version.

There is one more, apparently more widely used, file `/etc/os-release` that contains release version information similar to lsb-release. This pull request adds support for getting os version from `/etc/os-release` file.


johnboiles (approved):

ukstv:
So when release?

thboop (approved):
LGTM

PR #609, opened 2020-10-19
Add findInPath method to locate all matching executables in the system path

smorimoto:
David is the original author. This PR is mainly for #610.

PR #610, opened 2020-10-19
Prefer using GNU tar

smorimoto:
Note that we need to merge #609 and release the new version before merge this PR, and there are a few things to do after that.


joshmgross:
Using BSD tar on Windows is intentional and avoids a lot of issues with GNU tar , see https://github.com/actions/cache/pull/126 for more details.

What issues are you trying to solve with GNU tar on Windows? 
If we do want to support GNU tar on Windows, it should be optional and we'll need to add workarounds to make it work with Windows file paths (`--force-local` flag + replacing `\` in paths with `/`).

smorimoto:
Hi, Josh! I'm so sorry for the late reply!

Yeah, I definitely remember that, but now there is a problem like https://github.com/actions/cache/issues/301 with BSD tar and this PR is one of the big steps to get around it and make zstd work on Windows as well.

smorimoto:
I actually didn't think it was going to be a big problem until we disabled it on Windows, but after that, I realized that actions/cache was too slow on Windows.

dhadka:
:wave: Just investigated two issues related to BSD vs GNU tar

1. https://github.com/actions/cache/issues/465 - Changing the `PATH` caused the cache action to switch from `BSD tar` to `GNU tar` between the restore and save steps on Windows.  Due to the ZStd issue with BSD tar on Windows, this caused it to cache miss.

2. https://github.com/actions/cache/issues/460 and https://github.com/actions/cache/issues/403 - BSD tar on MacOS corrupting executables and`.dylib` files.  Not sure why this is happening, but using GNU tar fixes the issue.

smorimoto:
Also, if you are using Cygwin. If you are using Cygwin in your workflow, it will call Cygwin's tar, which will cause an error.

smorimoto:
To address this, we need to review, merge, and release the relevant PR this depends on.

smorimoto:
I procrastinating the my work on this PR a little, but will try to get there. @joshmgross @dhadka If you can update the patch quicker, please do that yourself.

thboop:
I'm going to close this for now, please reopen when its ready for review!

PR #620, opened 2020-11-01
doc: commands: add missing assignment char

glensc:
this seemed like an error:

executing the snipped like:
```sh
GITHUB_ENV=/tmp/GITHUB_ENV

echo 'JSON_RESPONSE<<EOF' >> $GITHUB_ENV
curl https://httpbin.org/json >> $GITHUB_ENV
echo 'EOF' >> $GITHUB_ENV
```

produced output:

```
$ cat /tmp/GITHUB_ENV
JSON_RESPONSE<<EOF
{
  "slideshow": {
    "author": "Yours Truly",
    "date": "date of publication",
    "slides": [
      {
        "title": "Wake up to WonderWidgets!",
        "type": "all"
      },
      {
        "items": [
          "Why <em>WonderWidgets</em> are great",
          "Who <em>buys</em> WonderWidgets"
        ],
        "title": "Overview",
        "type": "all"
      }
    ],
    "title": "Sample Slide Show"
  }
}
EOF
```

while I think it should be `JSON_RESPONSE=<EOF`.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -152,7 +152,7 @@ steps:
   - name: Set the value
     id: step_one
     run: |
-        echo 'JSON_RESPONSE<<EOF' >> $GITHUB_ENV
+        echo 'JSON_RESPONSE=<<EOF' >> $GITHUB_ENV
         curl https://httpbin.org/json >> $GITHUB_ENV
         echo 'EOF' >> $GITHUB_ENV
 ```


glensc:
currently it's not clear to me what is the syntax of $GITHUB_ENV file. is it shell script (so the heredoc would work?

documenting syntax is important, to get answer to such questions:
- what escape rules are present?
- how is (double) quotes passed?
- how to include space in value?

also, unclear, is the $GITHUB_ENV trickery needed when using "steps" as documentation says it's needed to pass data between "jobs", but "steps" are sub-components of "jobs"....

glensc:
perhaps the patch is not correct, produces weird error:

```
Run echo 'JSON_RESPONSE=<<EOF' >> $GITHUB_ENV
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   429  100   429    0     0   9750      0 --:--:-- --:--:-- --:--:--  9750
Error: Unable to process file command 'env' successfully.
Error: Invalid environment variable format '{'
```

- https://github.com/eventum/eventum/runs/1338475606#step:9:17

so, I conclude the file is not processed by shell and the syntax is very weird.

for the folks understanding how shell scripts work, the syntax should be changed to include equal sign. for that the parser that handles $GITHUB_ENV file should be updated first.

thboop:
@glensc,

The docs here are correct, you can try it out in a workflow
```
name: CI
on: push
jobs:
  job-container:
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest]
    name: runner-basics-job-container
    steps:
    - run: |
        echo 'JSON_RESPONSE<<EOF' >> $GITHUB_ENV
        curl https://httpbin.org/json >> $GITHUB_ENV
        echo 'EOF' >> $GITHUB_ENV
    - run: echo "hi"
```
and see that the `JSON_RESPONSE` env variable is set in the second step.

Before this example, we do provide a basic example [as well](https://github.com/actions/toolkit/blob/28a6748dd78c6d2bd3492fda2c8df17f51b3e7a0/docs/commands.md#set-an-environment-variable)

> so, I conclude the file is not processed by shell and the syntax is very weird.
It is not.
You can see the source code for how we parse it [here](https://github.com/actions/runner/blob/main/src/Runner.Worker/FileCommandManager.cs#L156)

> use the value.
It just sets an env variable for use in later steps. Depending on your shell there are different ways to access those, but I can look into expanding the sample and providing more escaping information, or feel free to submit a pr.

> what escape rules are present?
We reserve certain characters (`<<` `=`), otherwise no real escaping should be needed
> how is (double) quotes passed?
The correct output just needs to end up in the file, so if you were using bash, you would use
```
echo "myenvwithdoubles=\"\"BAR" >> $GITHUB_ENV
```
> how to include space in value?
```
echo "FOO=BA R" >> $GITHUB_ENV
```

I'm going to close out this pr @glensc , but feel free to submit a pr with more documentation information, or file an issue for us to update it.

PR #624, opened 2020-11-03
Handle non-successful responses from uploadChunk

dhadka:
`uploadChunk` is not handling non-successful status codes.  As a result, if the upload fails it will still try to seal the cache.  This ultimately fails, but is adding unnecessary calls to the service as well as creating noise in our telemetry.


aiqiaoy (approved):
You need to bump the cache package version

aiqiaoy (approved):
👍

PR #626, opened 2020-11-04
Explain to the ignorant reader where this `STATE_` prefix comes from.

edburns:
Explain to the ignorant reader where this `STATE_` prefix comes from.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -89,7 +89,7 @@ Save a state to an environmental variable that can later be used in the main or 
 echo "::save-state name=FOO::foovalue"
 ```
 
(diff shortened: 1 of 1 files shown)


matthewmccullough:
Nice little documentation enhancement, my conference-days friend. Thanks for the investment in the education of the next reader, @edburns !

matthewmccullough:
@dhadka As you have time, would you take a couple second glance at this PR just to crisp up the docs? I was just reviewing it with @edburns and it seems small and sensible at first look.

dhadka (approved):
LGTM.  For the curious, this is where `STATE_` is prepended - https://github.com/actions/runner/blob/95e2158dc6e1d8bdc10112a66c5353c5acdff687/src/Runner.Worker/ActionRunner.cs#L235

Issue #214, opened 2019-11-12
The recommendation to "move" tags


friederbluemle:
This is regarding the official [Versioning documentation](https://github.com/actions/toolkit/blob/master/docs/action-versioning.md) for authoring custom GitHub Actions.

I really had to read it several times to make sure that an "official" GitHub controlled repository is **recommending to "move" tags between commits**. 😨 

Please, if you haven't already, read the section [**On Re-tagging**](https://git-scm.com/docs/git-tag#_on_re_tagging) in the [official Git man page for the tag command](https://git-scm.com/docs/git-tag). Specifically, "The insane thing" (point 2), and why "Git does not change tags behind users back".

These mechanisms are in place for a reason, and together with many other amazing things about Git, have contributed to the huge success and the way Git/GitHub have revolutionized version control and collaboration over the past decade. Please, don't try to circumvent them and recommend anti-patterns as best practices. This totally sends the wrong signal, especially to new users who are less familiar with Git. **Do not force re-tag and force override already published tags**.

All that said, most what is written in the Versioning documentation document makes sense, and _is_ following best practices 👍 It seems like a solution that uses Git tags _in the intended way_ with minimal change to what is already there could be:

- Continue to use the `master` branch for ongoing development
- Tag actual releases using standard semver tags (`v1.0.0`, `v1.1.0`)
- Do not use "moving" tags
- Use a **branch** `v1` instead of `releases/v1`

A branch is the correct construct to point to different commits over time.

### Example

Say `v1` points to `v1.0.0`, and a new version `v1.1.0` has been tagged. After successfully testing `v1.1.0`, `v1` could simply be fast-forwarded to `v1.1.0`. There is no need to force anything, and from a user's perspective, everything remains the same, e.g. using an action would _still_ look like this:

```
uses: actions/custom-action@v1
```

bryanmacfarlane:
Yes, this has had many discussions and even the actions/javascript-action repo has the walk through creating a v1 branch instead of moving a tag.

When we did that, there were other objections that it led to less opportunity to validate and "release" the action with an explicit action.  Merging into the release branch led to it immediately being used by customers.  As opposed to typically merging (potentially multiple) changes into a release branch, validating the integration and then having an explicit reaction action.  The tag move is that.  If the solution is a v1 branch, and you're integrating multiple changes from feature branches, it's still possible with an integration branch off of the release branch so it's still possible.

[...]

PR #631, opened 2020-11-11
Recommend updating branches for major versions and not tags in the versioning docs

eregon:
This should be an obvious change.
`git` tags should not be moved, I would hope GitHub knows that, since most `git` users do.

For instance see what the `git tag` man page says about moving a tag: https://git-scm.com/docs/git-tag#_on_re_tagging
To keep it short, the man page calls moving a tag `The insane thing.`. Should be clear it's an anti-pattern.
It also explains there are various problems if users already saw the old tag, and it will not be updated for them on `git pull` which might result in confusion (e.g., `git pull; git checkout v1` might checkout an old version of the tag).

A branch is what naturally moves in `git`.
Are there any advantages to use a moving tag for major versions? I don't see any.

Also this naturally prevents unintentionally going back with a major version:
```
# Assume the current version is v1.2

$ git tag -fa v1 -m "Update v1 tag" v1.1
$ git push origin v1 --force
# oops, rewinded history, a mistake

$ git push origin v1.1:v1
push rejected, not a fast forward
```

So, how about recommending best `git` practices in the official versioning documentation of GitHub Actions with this PR?

Related: https://github.com/actions/toolkit/issues/214#issuecomment-583623502

### DIFF
--- a/docs/action-versioning.md
+++ b/docs/action-versioning.md
@@ -38,12 +38,28 @@ Binding to the immutable full sha1 may offer more reliability.  However, note th
 
 <img src="https://user-images.githubusercontent.com/33549821/78670739-36f5ae00-78ac-11ea-9660-57d5687ce520.png" alt="screenshot" height="250"/>
 
-3. **Make the new release available to those binding to the major version tag**: Move the major version tag (v1, v2, etc.) to point to the ref of the current release. This will act as the stable release for that major version. You should keep this tag updated to the most recent stable minor/patch release.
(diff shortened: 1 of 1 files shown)


umarcor:
@eregon, did you ever get any feedback about this PR?

Paebbels:
@ GitHub Team, why is such an important PR not merged yet. Even no reaction to it?

How is it even possible that GitHub developers have such a bad knowledge about Git and it's behavior as well as the  internal implementation of Git. Suggesting to move tags is not just bad practice...

PR #633, opened 2020-11-13
Replace find with which in io package description

smorimoto:
This avoids confusion. ​(At least I thought it was there, so I read the source.)


smorimoto:
We can merge this right away.

thboop (approved):
Nice Catch, thank you for the contribution!

Issue #630, opened 2020-11-11
core.addPath is deprecated? 


actboy168:
**Describe the bug**
I updated @actions/core in my action to 1.2.6. Then core.addPath does not seem to work correctly, how should I rewrite my code?

**To Reproduce**
https://github.com/actboy168/setup-luamake

peaceiris:
Duplicate of #622

actboy168:
@peaceiris  I think this is different from #622. I have updated `@actions/core` to the latest, then `core.addPath` failed.

peaceiris:
https://github.com/actboy168/setup-luamake/blob/2ee6b2f519c884b8c6f71703a91d1ab89b78d234/package.json#L22

Please follow the npm basic usage first.

```diff
+    "dependencies": {
+        "@actions/core": "^1.2.6",
+        "@actions/exec": "^1.0.4"
    },
    "devDependencies": {
-        "@actions/core": "^1.2.6",
-        "@actions/exec": "^1.0.4",
        "@zeit/ncc": "^0.20.5"
    }
```

actboy168:
@peaceiris I have changed as you said, but the result is no different.

surfraz:
Hello, I have also tried this with "@actions/core": "^1.2.6", this version is still using the old set-env deprecated method, however the latest commit seems to have the fix to use filecommands. The deprecation warning specifies using 1.2.6: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

Please could we have a new version of actions/core created with the fix included

peaceiris:
That's odd. I think the release 1.2.6 already has both the new approach and the old one.

https://github.com/actions/toolkit/blob/e7eb2c741847a05686a1b8df93dea4b244640b17/packages/core/src/core.ts#L67-L75

Actually, [peaceiris/actions-hugo](https://github.com/peaceiris/actions-hugo)@v2.4.13 works well with no errors.

- `peaceiris/actions-hugo@v2.4.12` using `@actions/core 1.2.5` returns the warning about `add-path`
- `peaceiris/actions-hugo@v2.4.13` using `@actions/core 1.2.6` returns no warning

<img width="689" alt="Screen Shot 2020-11-12 at 23 18 11" src="https://user-images.githubusercontent.com/30958501/98951447-9a7c4400-253d-11eb-9626-524c511243d7.png">
[...]

PR #635, opened 2020-11-13
Update to actions/core@1.2.6 in artifact, glob, and tool-cache packages

jwtracy:
Avoid set-env or add-path blockage on… Nov 16... https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/ https://github.com/actions/toolkit/issues/630

Note: cache depends on glob and should be fixed after glob is updated


jwtracy:
Duplicate of https://github.com/actions/toolkit/pull/636

PR #636, opened 2020-11-13
Update @actions/core in remaining toolkit packages

thboop:
Currently, none of the toolkit packages use the [commands that are being deprecated on November 16th](https://github.blog/changelog/2020-11-09-github-actions-removing-set-env-and-add-path-commands-on-november-16/).

However, they do include other versions of `@actions/core`, that contain that vulnerability even if they aren't invoking it. Let's update these packages to use the latest version.


joshmgross (approved):

jwtracy:
Thanks, Thomas!

PR #650, opened 2020-11-25
Make caching more verbose

rosik:
- Print cache size when saving cache (similar to restoring)
- Print restore success (similar to saving)
- Print cached file list (if debug logging is enabled)

This makes logs of saving and restoring cache more similar to each other.

image

image

Part of https://github.com/actions/cache/issues/471

### DIFF
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -2,7 +2,7 @@ import * as core from '@actions/core'
 import * as path from 'path'
 import * as utils from './internal/cacheUtils'
 import * as cacheHttpClient from './internal/cacheHttpClient'
-import {createTar, extractTar} from './internal/tar'
+import {createTar, extractTar, listTar} from './internal/tar'
 import {DownloadOptions, UploadOptions} from './options'
 
 export class ValidationError extends Error {
@@ -100,6 +100,10 @@ export async function restoreCache(
       options
     )
 
+    if (core.isDebug()) {
+      await listTar(archivePath, compressionMethod)
+    }
+
     const archiveFileSize = utils.getArchiveFileSizeIsBytes(archivePath)
     core.info(
       `Cache Size: ~${Math.round(
@@ -108,6 +112,7 @@ export async function restoreCache(
     )
 
     await extractTar(archivePath, compressionMethod)
+    core.info('Cache restored successfully')
   } finally {
     // Try to delete the archive to save space
     try {
@@ -162,6 +167,9 @@ export async function saveCache(
   core.debug(`Archive Path: ${archivePath}`)
 
   await createTar(archiveFolder, cachePaths, compressionMethod)
+  if (core.isDebug()) {
+    await listTar(archivePath, compressionMethod)
+  }
 
   const fileSizeLimit = 5 * 1024 * 1024 * 1024 // 5GB per repo limit
   const archiveFileSize = utils.getArchiveFileSizeIsBytes(archivePath)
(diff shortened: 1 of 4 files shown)


rosik:
Hi, guys. Could someone review my patch, please?

rosik:
@dhadka @joshmgross @konradpabjan ?

packages/cache/src/cache.ts
```diff
@@ -100,6 +100,7 @@ export async function restoreCache(
       options
     )
 
+    await listTar(archivePath, compressionMethod)
```
joshmgross, line 103: Listing out the contents of the tar file might not be ideal in all scenarios. If users are concerned more about cache speed than log verbosity, we should consider making this optional.

rosik: Thanks for your review. I've added it under `if (core.isDebug())` conditinal.

rosik: I'd like to note that `@actions/glob` also has its own debug output, but it's not the same as `listTar`:

image

rosik:
Ping

rosik:
@joshmgross I've updated the PR a week ago. Do you expect from me any other actions or do I miss something?

joshmgross:
@aiqiaoy or @yacaovsnc could you take a look at this PR?

yacaovsnc:
Thanks @rosik for the PR.  Could you please add a test for the `listTar` method you created?  The tests are here: https://github.com/actions/toolkit/blob/main/packages/cache/__tests__/tar.test.ts

rosik:
I've added a unit test, but I'm worried it's not covered by an end-to-end test. The `listTar` function is only called when `ACTIONS_STEP_DEBUG` is set, so it's not run in PR checks: https://github.com/actions/toolkit/pull/650/checks?check_run_id=1585714569. What do you think? If it's desirable, could you give me a hint on how to implement it?

yacaovsnc:
Sorry for the silence, I was out for the last couple weeks.  Personally I think it's fine that this isn't covered in the end-to-end test - to test that we need to set `ACTIONS_STEP_DEBUG` as a secret value.  

FWIW, I forked the repo and set `ACTIONS_STEP_DEBUG` in my repo setting, and saw the additional debug messages.  If really desired, an admin of this repo can enable debug?  I don't think it blocks this PR anyway.

yacaovsnc (approved):
Thanks for the PR.  Looks good to me.

rosik:
If you ask for my opinion, I vote for enabling ACTIONS_STEP_DEBUG in the main repo. It sounds helpful for the coverage.

So, what's next? I've rebased my patch on top of the fresh main branch. Do you expect any other actions from me now?

rosik:
Ping

rosik:
@dhadka @joshmgross @konradpabjan @aiqiaoy @yacaovsnc ping

yacaovsnc:
@rosik the cache action can include this change after we release a new npm package.  Are you waiting for this change? 

@konradpabjan let's include this change next time we release.  Do you handle npm releases for cache toolkit too?

rosik:
>  Are you waiting for this change?

So so. We can always invent a workaround for debugging, but it's usually a little bit nasty.

TSLC369 (commented):
README.md

TSLC369 (commented):
README.md

TSLC369 (commented):
README.md

TSLC369 (commented):
[]()

da-art85 (commented):
😗🅰️➕⭕

PR #651, opened 2020-11-26
Remove spyon, use readable stream to test pipeResponseToFile

hashtagchris:
Removing `jest.spyOn(DownloadHttpClient.prototype, 'pipeResponseToFile')` from the download artifact tests, and using a readable stream object to test `pipeResponseToFile`. Also verifying the destination file size on disk.

This is set up for a follow-up PR with new test cases for truncated responses, https://github.com/actions/toolkit/pull/652.


yacaovsnc (approved):
LGTM.  Thanks for adding the check for file length.

sohoskool (commented):
`#651`

PR #652, opened 2020-11-26
Retry artifact download when response stream is truncated

hashtagchris:
Retry the artifact download when gunzip fails or the response stream is truncated.

Closes https://github.com/actions/download-artifact/issues/53 and https://github.com/actions/download-artifact/issues/55

- [x] Add tests
- [ ] Fix runtime code
- [x] Merge https://github.com/actions/toolkit/pull/651 first, since this PR builds on top of it.

### DIFF
--- a/packages/artifact/src/internal/download-http-client.ts
+++ b/packages/artifact/src/internal/download-http-client.ts
@@ -264,6 +264,12 @@ export class DownloadHttpClient {
         const gunzip = zlib.createGunzip()
         response.message
           .pipe(gunzip)
+          .on('error', error => {
+            core.error(
+              `An error has been encountered while attempting to decompress a file`
+            )
+            reject(error)
+          })
           .pipe(destinationStream)
           .on('close', () => {
             resolve()
--- a/packages/artifact/src/internal/download-http-client.ts
+++ b/packages/artifact/src/internal/download-http-client.ts
@@ -263,12 +263,20 @@ export class DownloadHttpClient {
       if (isGzip) {
         const gunzip = zlib.createGunzip()
         response.message
+          .on('error', error => {
+            core.error(
+              `An error occurred while attempting to read the response stream`
+            )
+            reject(error)
+            gunzip.close()
+          })
           .pipe(gunzip)
           .on('error', error => {
             core.error(
-              `An error has been encountered while attempting to decompress a file`
+              `An error occurred while attempting to decompress the response stream`
             )
             reject(error)
+            destinationStream.close()
           })
           .pipe(destinationStream)
           .on('close', () => {
@@ -276,19 +284,26 @@ export class DownloadHttpClient {
           })
           .on('error', error => {
             core.error(
-              `An error has been encountered while decompressing and writing a downloaded file to ${destinationStream.path}`
+              `An error occurred while writing a downloaded file to ${destinationStream.path}`
             )
             reject(error)
           })
       } else {
         response.message
+          .on('error', error => {
+            core.error(
+              `An error occurred while attempting to read the response stream`
+            )
+            reject(error)
+            destinationStream.close()
+          })
           .pipe(destinationStream)
           .on('close', () => {
             resolve()
           })
           .on('error', error => {
             core.error(
-              `An error has been encountered while writing a downloaded file to ${destinationStream.path}`
+              `An error occurred while writing a downloaded file to ${destinationStream.path}`
             )
             reject(error)
           })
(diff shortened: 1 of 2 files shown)


packages/artifact/src/internal/download-http-client.ts
```diff
@@ -264,6 +264,12 @@ export class DownloadHttpClient {
         const gunzip = zlib.createGunzip()
         response.message
           .pipe(gunzip)
+          .on('error', error => {
```
hashtagchris, line 267: The `on('error', ...)` callback on line line 277 isn't called when gunzip fails. I don't think `pipe` propagates errors.

Without this block the error is unhandled, and the new test times out.

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -263,26 +263,44 @@ export class DownloadHttpClient {
...
+            )
+            reject(error)
+          })
           .pipe(gunzip)
+          .on('error', error => {
+            core.error(
+              `An error occurred while attempting to decompress the response stream`
+            )
+            reject(error)
+          })
```
hashtagchris, line 278: Reviewers: The existing `on('error', ...)` callback below isn't called when gunzip fails. `pipe` doesn't propagate errors across streams, so you have to monitor each.

Without this block the error is unhandled, and the new test times out.

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -263,26 +263,44 @@ export class DownloadHttpClient {
       if (isGzip) {
         const gunzip = zlib.createGunzip()
         response.message
+          .on('error', error => {
```
hashtagchris, line 266: We could monitor for `aborted` as well. https://nodejs.org/dist/latest-v14.x/docs/api/http.html#http_class_http_incomingmessage

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -263,26 +263,47 @@ export class DownloadHttpClient {
...
+            core.error(
+              `An error occurred while attempting to read the response stream`
+            )
+            reject(error)
+            gunzip.close()
+          })
           .pipe(gunzip)
+          .on('error', error => {
```
hashtagchris, line 274: Reviewers: The existing `on('error', ...)` callback below isn't called when gunzip fails. `pipe` doesn't propagate errors across streams, so you have to monitor each.

Without this block the error is unhandled, and the new test times out.

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -263,26 +263,47 @@ export class DownloadHttpClient {
       if (isGzip) {
         const gunzip = zlib.createGunzip()
         response.message
+          .on('error', error => {
```
hashtagchris, line 266: The existing `on('error', ...)` callback below isn't called when gunzip fails, or response.message produces an error. `pipe` doesn't propagate errors across streams, so you have to monitor each.

Without the new blocks the error is unhandled, and the new test times out.

I'm closing the downstream stream (hah!) on error. From https://nodejs.org/dist/latest-v14.x/docs/api/stream.html#stream_readable_pipe_destination_options:

> One important caveat is that if the `Readable` stream emits an error during processing, the `Writable` destination is not closed automatically. If an error occurs, it will be necessary to manually close each stream in order to prevent memory leaks.

yacaovsnc: I am a little bit unclear about when to retry here.  If the download failed because the socket is opened too long (for downloading large files), would retry help?

hashtagchris: Good question. In short, I'm hoping retries will help. 😄 

In the case of smaller files, there are likely transient requests that stall out or timeout that will be quicker on retry.

For larger files (https://github.com/actions/download-artifact/issues/55), is there a technical reason we'd be able to upload large files we can't download? Do we only chunk on upload? Should we add a max size check to upload-artifact if download-artifact is unlikely to succeed?

One option to add retries to hopefully improve the success rate for smaller files, and then revisit this code if customers still report issues with larger files.

yacaovsnc: We do chunked upload but I don't believe we do chunked download.  

> and then revisit this code if customers still report issues with larger files.

This is what I am afraid of, since in https://github.com/actions/download-artifact/issues/55 the description says there is a 100% failure rate, I am afraid the retry will be in vain, and we will incur more minutes per build from those retries.

hashtagchris:
https://github.com/actions/toolkit/pull/661 supersedes this PR. It includes the commits in this PR.

PR #659, opened 2020-12-02
Update cache: add getCacheEntry to check if a cache entry exists without having to download

MartijnHols:
This adds a `getCacheEntry` export to the `cache` package that can be used to determine if a cache entry exists. When merged, this method can be used to add a flag to the the cache-action to only check if a cache entry exists for a key. Its `cache-hit` output can then by used to speed up workflows that use big caches and to skip work if the result of the work is already in the cache.

For context: I am using the cache action to share my build across workflow steps. If the build already exists, there is no point downloading it or building it again and the later steps can just reuse the old cached result.


sohoskool (approved):

sohoskool:
cmmkt changes

lastthyme (approved):

PR #661, opened 2020-12-03
Retry artifact download when response stream is truncated (continued)

yacaovsnc:
Continuation of https://github.com/actions/toolkit/pull/652

Closes actions/download-artifact#53 and actions/download-artifact#55

We retry when we receive some non success http status code such as 409, 503, etc.  However even if we received a 200 success status code but the response stream were interrupted during download, we should also retry.  

When we stream down the file, we could stream with gzip compression, and if we do, we rely on gzip processing to detect the stream is corrupted.  If the stream is not gzip compressed, we will verify the downloaded file is of the same size specified in `Content-Length` header.

### DIFF
--- a/packages/artifact/__tests__/download.test.ts
+++ b/packages/artifact/__tests__/download.test.ts
@@ -196,7 +196,6 @@ describe('Download Tests', () => {
     const targetPath = path.join(root, 'FileD.txt')
 
     setupDownloadItemResponse(true, 200, true, fileContents, true)
-    setupThrowWhenReadingStream()
     const downloadHttpClient = new DownloadHttpClient()
 
     const items: DownloadItem[] = []
@@ -290,18 +289,6 @@ describe('Download Tests', () => {
         })
       })
     })
-  }
-
-  function setupThrowWhenReadingStream(): void {
-    jest
-      .spyOn(DownloadHttpClient.prototype, 'pipeResponseToFile')
-      .mockImplementationOnce(async () => {
-        return new Promise<void>(() => {
-          throw new Error(
-            'simulate GZip reading truncated buffer and throw Z_BUF_ERROR'
-          )
-        })
-      })
   }
 
   /**
--- a/packages/artifact/__tests__/download.test.ts
+++ b/packages/artifact/__tests__/download.test.ts
@@ -293,10 +293,10 @@ describe('Download Tests', () => {
   }
 
   function setupThrowWhenReadingStream(): void {
-    const spyInstance = jest
+    jest
       .spyOn(DownloadHttpClient.prototype, 'pipeResponseToFile')
       .mockImplementationOnce(async () => {
-        return new Promise<void>(resolve => {
+        return new Promise<void>(() => {
           throw new Error(
             'simulate GZip reading truncated buffer and throw Z_BUF_ERROR'
           )
--- a/packages/artifact/__tests__/download.test.ts
+++ b/packages/artifact/__tests__/download.test.ts
@@ -124,7 +124,7 @@ describe('Download Tests', () => {
     )
     const targetPath = path.join(root, 'FileA.txt')
 
-    setupDownloadItemResponse(true, 200, false, fileContents)
+    setupDownloadItemResponse(true, 200, false, fileContents, false)
     const downloadHttpClient = new DownloadHttpClient()
 
     const items: DownloadItem[] = []
@@ -147,7 +147,7 @@ describe('Download Tests', () => {
     )
     const targetPath = path.join(root, 'FileB.txt')
 
-    setupDownloadItemResponse(false, 200, false, fileContents)
+    setupDownloadItemResponse(false, 200, false, fileContents, false)
     const downloadHttpClient = new DownloadHttpClient()
 
     const items: DownloadItem[] = []
@@ -171,7 +171,7 @@ describe('Download Tests', () => {
       const fileContents = Buffer.from('try, try again\n', defaultEncoding)
       const targetPath = path.join(root, `FileC-${statusCode}.txt`)
 
-      setupDownloadItemResponse(false, statusCode, false, fileContents)
+      setupDownloadItemResponse(false, statusCode, false, fileContents, true)
       const downloadHttpClient = new DownloadHttpClient()
 
       const items: DownloadItem[] = []
@@ -195,7 +195,8 @@ describe('Download Tests', () => {
     )
     const targetPath = path.join(root, 'FileD.txt')
 
-    setupDownloadItemResponse(true, 200, true, fileContents)
+    setupDownloadItemResponse(true, 200, true, fileContents, true)
+    setupThrowWhenReadingStream()
     const downloadHttpClient = new DownloadHttpClient()
 
     const items: DownloadItem[] = []
@@ -218,7 +219,7 @@ describe('Download Tests', () => {
     )
     const targetPath = path.join(root, 'FileE.txt')
 
-    setupDownloadItemResponse(false, 200, true, fileContents)
+    setupDownloadItemResponse(false, 200, true, fileContents, true)
     const downloadHttpClient = new DownloadHttpClient()
 
     const items: DownloadItem[] = []
@@ -291,6 +292,18 @@ describe('Download Tests', () => {
     })
   }
 
+  function setupThrowWhenReadingStream(): void {
+    const spyInstance = jest
+      .spyOn(DownloadHttpClient.prototype, 'pipeResponseToFile')
+      .mockImplementationOnce(async () => {
+        return new Promise<void>(resolve => {
+          throw new Error(
+            'simulate GZip reading truncated buffer and throw Z_BUF_ERROR'
+          )
+        })
+      })
+  }
+
   /**
    * Setups up HTTP GET response for downloading items
    * @param isGzip is the downloaded item gzip encoded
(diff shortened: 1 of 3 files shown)


packages/artifact/src/internal/download-http-client.ts
```diff
@@ -201,6 +203,29 @@ export class DownloadHttpClient {
...
+
+    const resetDestinationStream = async (
+      fileDownloadPath: string
+    ): Promise<void> => {
+      await rmFile(fileDownloadPath)
+      destinationStream = fs.createWriteStream(fileDownloadPath)
+    }
+
     // keep trying to download a file until a retry limit has been reached
     while (retryCount <= retryLimit) {
```
konradpabjan, line 230: Not related to your changes, but I think the retryCount is never getting incremented anywhere here 🤔  Was checking for possible infinite loops if `forceRetry` is ever set to true

yacaovsnc: It's incremented in the `backoff` function.

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -201,6 +203,29 @@ export class DownloadHttpClient {
...
 
+    const isAllBytesReceived = (
+      expected?: string,
+      received?: number
+    ): boolean => {
+      // be lenient, if any input is missing, assume success, i.e. not truncated
+      if (
+        !expected ||
+        !received ||
+        process.env['ACTIONS_ARTIFACT_SKIP_DOWNLOAD_VALIDATION']
```
konradpabjan, line 214: What's the plan with this env variable? Just a fallback if there are issues that users or us can set?

yacaovsnc: Correct, a feature flag sort of thing for now.

konradpabjan: I'm blind 🤦

packages/artifact/src/internal/download-http-client.ts
konradpabjan: I'm blind 🤦

yacaovsnc:
I tested all the positive cases, not sure how to test the negative case since my download was working fine.  I tested with 1GB files.

Tested:
1. Download incompressible file, so no gzip.
1. Download highly compressible file with gzip stream
1. Skipping file size check by setting the env var. 

Headers:
```
Total number of files that will be downloaded: 2
##### Begin Diagnostic HTTP information #####
Status Code: 200
Status Message: OK
Header Information: {
  "cache-control": "no-store,no-cache",
  "pragma": "no-cache",
  "content-length": "1043656",
  "content-type": "application/octet-stream",
  "content-encoding": "gzip",
  "strict-transport-security": "max-age=2592000",
  "x-tfs-processid": "8e7af2a5-0c3f-45e5-9baa-1caea8857637",
  "activityid": "55af6bc4-e899-40ce-87cd-a8c28b7be3bc",
  "x-tfs-session": "55af6bc4-e899-40ce-87cd-a8c28b7be3bc",
  "x-vss-e2eid": "55af6bc4-e899-40ce-87cd-a8c28b7be3bc",
  "x-vss-senderdeploymentid": "193695a0-0dcd-ade4-f810-b10ad24a9829",
  "x-frame-options": "SAMEORIGIN",
  "content-disposition": "attachment; filename=zero.bin; filename*=UTF-8''zero.bin",
  "x-msedge-ref": "Ref A: 4276B404BFFE4DE7A3F445E368977664 Ref B: BL2EDGE0912 Ref C: 2020-12-03T17:37:15Z",
  "date": "Thu, 03 Dec 2020 17:37:14 GMT"
}
###### End Diagnostic HTTP information ######
##### Begin Diagnostic HTTP information #####
Status Code: 200
Status Message: OK
Header Information: {
  "cache-control": "no-store,no-cache",
  "pragma": "no-cache",
  "content-length": "1073741824",
  "content-type": "application/octet-stream",
  "strict-transport-security": "max-age=2592000",
  "x-tfs-processid": "27718d9d-1c29-459a-b6f7-1de5dad76990",
  "activityid": "4708872f-155c-4da7-af4b-d6e153e56a0c",
  "x-tfs-session": "4708872f-155c-4da7-af4b-d6e153e56a0c",
  "x-vss-e2eid": "4708872f-155c-4da7-af4b-d6e153e56a0c",
  "x-vss-senderdeploymentid": "193695a0-0dcd-ade4-f810-b10ad24a9829",
  "x-frame-options": "SAMEORIGIN",
  "content-disposition": "attachment; filename=random.bin; filename*=UTF-8''random.bin",
  "x-msedge-ref": "Ref A: 0363D65B20FD441E8978A56FD94F8DA6 Ref B: ASHEDGE1209 Ref C: 2020-12-03T17:37:15Z",
  "date": "Thu, 03 Dec 2020 17:37:14 GMT"
}
```

konradpabjan (commented):
Changes look really good and the new tests are also 👍 

One concern is regarding some failing tests, there might be some issues on Windows: 
image

yacaovsnc:
> One concern is regarding some failing tests, there might be some issues on Windows:

Yeah, I added couple debugging statement and they started to fail.  Going to take a look at why.

yacaovsnc:
> One concern is regarding some failing tests, there might be some issues on Windows:

Seems like the `stream` is not closed on Windows so we couldn't unlink the file.  🤦

yacaovsnc:
@konradpabjan @hashtagchris do you mind to take another look?  I fixed the unit test.  Thanks!

konradpabjan (approved):
Looks good! 🚀

packages/artifact/__tests__/download.test.ts
```diff
@@ -188,6 +188,53 @@ describe('Download Tests', () => {
...
 
+  it('Test retry on truncated response with gzip', async () => {
+    const fileContents = Buffer.from(
+      'Sometimes gunzip fails on the first try\n',
+      defaultEncoding
+    )
+    const targetPath = path.join(root, 'FileD.txt')
+
+    setupDownloadItemResponse(true, 200, true, fileContents, true)
+    setupThrowWhenReadingStream()
```
hashtagchris, line 199: I don't think this `spyon` call is necessary. When you gzip the `fileContents` buffer and truncate it, the gunzip stream in `pipeResponseToFile` should raise an error.

yacaovsnc: Hmm, from my local testing, I needed this.  We do not raise any error automatically from `pipeResponseToFile`.  If I comment out this `setupThrowWhenReadingStream()` the test will fail because we successfully downloaded the truncated file and didn't retry.

hashtagchris: Odd, I try commenting it out and it passed for me. https://github.com/hashtagchris/toolkit/commit/79dfd4df56ec318d09c32231df11601f8707dd04

yacaovsnc: Hmm you are right!  I think something was cached in my terminal so I wasn't running with the latest code?  Weird but in a new terminal I was able to remove this line and everything worked.

packages/artifact/__tests__/download.test.ts
```diff
@@ -253,17 +312,25 @@ describe('Download Tests', () => {
   function setupDownloadItemResponse(
     isGzip: boolean,
     firstHttpResponseCode: number,
-    fileContents: Buffer
+    truncateFirstResponse: boolean,
+    fileContents: Buffer,
```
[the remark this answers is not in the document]
yacaovsnc: Done.

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -263,26 +311,47 @@ export class DownloadHttpClient {
...
+            reject(error)
+            gunzip.close()
+          })
           .pipe(gunzip)
+          .on('error', error => {
+            core.error(
+              `An error occurred while attempting to decompress the response stream`
+            )
+            reject(error)
+            destinationStream.close()
```
hashtagchris, line 327:
```suggestion
            destinationStream.close()
            reject(error)
```

We should close the stream first before erroring and retrying.

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -263,26 +311,47 @@ export class DownloadHttpClient {
...
             reject(error)
           })
       } else {
         response.message
+          .on('error', error => {
+            core.error(
+              `An error occurred while attempting to read the response stream`
+            )
+            reject(error)
+            destinationStream.close()
```
hashtagchris, line 346:
```suggestion
            destinationStream.close()
            reject(error)
```

packages/artifact/src/internal/download-http-client.ts
```diff
@@ -263,26 +311,48 @@ export class DownloadHttpClient {
       if (isGzip) {
         const gunzip = zlib.createGunzip()
         response.message
+          .on('error', error => {
+            core.error(
+              `An error occurred while attempting to read the response stream`
+            )
```
hashtagchris, line 317: FYI, I was a little worried that calling `gunzip.close()` would trigger `on error` below and cause two core.errors to get logged. But I tried adding the following to the mock, and saw only this error block get hit:

```
          case 1:
            this.emit("error", new Error("injecting read error"))
            break
```

hashtagchris (approved):
Looks good - thanks for fixing this!

PR #675, opened 2020-12-16
Retry all http calls for artifact upload and download

konradpabjan:
## Overview

Currently _most_ http calls during artifact upload and download are retried, however disconnects and timeouts are possible so everything should be retried. Some documentation here regarding how we retry existing calls: https://github.com/actions/toolkit/blob/main/packages/artifact/docs/implementation-details.md

Artifact upload consists of 3 steps
- Single POST call - Create artifact container (not retried)
- Multiple PUT calls - Optionally gzip, then upload concurrently in chunks (retried)
- Single PATCH call - Update artifact size to indicate we are done (not retried)

Artifact download consists of 3 steps
- Single GET call - List available artifacts (not retried) 
- Single GET call - Get container items for a specific artifact (not retried)
- Multiple GET calls - Concurrently download the contents of the artifacts and decompress if necessary (retried)

These changes will ensure that every step along the way can be retried.

These changes should prevent artifact download & upload from failing in a few scenarios that customers are reporting:
https://github.com/actions/download-artifact/issues/72
https://github.com/actions/upload-artifact/issues/116
https://github.com/actions/upload-artifact/issues/123
https://github.com/actions/upload-artifact/issues/135

## Inspiration

Original inspiration behind these changes is some similar retry-ability that was added to actions cache as part of this PR: https://github.com/actions/cache/pull/306

The original retry-ability was re-worked a little bit and everything was moved over to the toolkit repo where you can find the code:
Code: https://github.com/actions/toolkit/blob/main/packages/cache/src/internal/requestUtils.ts
Tests: https://github.com/actions/toolkit/blob/main/packages/cache/__tests__/requestUtils.test.ts


packages/artifact/src/internal/utils.ts
```diff
@@ -65,7 +65,9 @@ export function isForbiddenStatusCode(statusCode?: number): boolean {
   return statusCode === HttpCodes.Forbidden
 }
 
-export function isRetryableStatusCode(statusCode?: number): boolean {
+export function isRetryableStatusCode(
+  statusCode?: number | undefined
```
brcrista, line 69: isn't `?` just sugar for `T | undefined`?

packages/artifact/src/internal/requestUtils.ts
```diff
@@ -0,0 +1,74 @@
+import {IHttpClientResponse} from '@actions/http-client/interfaces'
+import {isRetryableStatusCode, isSuccessStatusCode, sleep} from './utils'
+import * as core from '@actions/core'
+
+export async function retry<T>(
+  name: string,
+  method: () => Promise<T>,
+  getStatusCode: (response: T) => number | undefined,
+  errorMessages: Map<number, string>,
```
brcrista, line 9: I think it would be more flexible just to pass in:

```suggestion
  getErrorMessage: (response: T) => string,
```

Some API responses might have an error message in the body.

packages/artifact/src/internal/requestUtils.ts
```diff
@@ -0,0 +1,74 @@
+import {IHttpClientResponse} from '@actions/http-client/interfaces'
+import {isRetryableStatusCode, isSuccessStatusCode, sleep} from './utils'
+import * as core from '@actions/core'
+
+export async function retry<T>(
+  name: string,
+  method: () => Promise<T>,
```
brcrista, line 7: Technically, this isn't necessarily a method:

```suggestion
  operation: () => Promise<T>,
```

packages/artifact/src/internal/requestUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+  getStatusCode: (response: T) => number | undefined,
```
brcrista, line 8: When would a response not have a status code?

packages/artifact/src/internal/requestUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (!isRetryable) {
+      core.info(`${name} - Error is not retryable`)
+      break
+    }
+
+    core.info(
+      `${name} - Attempt ${attempt} of ${maxAttempts} failed with error: ${errorMessage}`
+    )
+
+    await sleep(delay)
```
brcrista, line 50: Should we increase the delay on each attempt?  I think we should also check `delay > 0`

packages/artifact/src/internal/requestUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+import {isRetryableStatusCode, isSuccessStatusCode, sleep} from './utils'
+import * as core from '@actions/core'
+
+export async function retry<T>(
+  name: string,
+  method: () => Promise<T>,
+  getStatusCode: (response: T) => number | undefined,
+  errorMessages: Map<number, string>,
+  maxAttempts: number,
+  delay: number
```
brcrista, line 11:
```suggestion
  delayMilliseconds: number
```

packages/artifact/src/internal/requestUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+    attempt++
+  }
+
+  if (extraErrorInformation) {
+    throw Error(`${name} failed: ${errorMessage} : ${extraErrorInformation}`)
+  }
+  throw Error(`${name} failed: ${errorMessage}`)
+}
+
+export async function retryHttpClientResponse<T>(
```
brcrista, line 60:
```suggestion
export async function retryHttpClientRequest<T>(
```

or just `httpRetry`?

packages/artifact/__tests__/upload.test.ts
```diff
@@ -102,7 +102,7 @@ describe('Upload Tests', () => {
       uploadHttpClient.createArtifactInFileContainer(artifactName)
     ).rejects.toEqual(
       new Error(
-        `Unable to create a container for the artifact invalid-artifact-name at ${getArtifactUrl()}`
+        `Create Artifact Container failed: Artifact service responded with 400 : The artifact name invalid-artifact-name is not valid. Request URL ${getArtifactUrl()}`
```
yacaovsnc, line 105: Personally, I am not a fan of showing users the error code - that feels like implementation detail and doesn't help user to fix the issue.  Also we could change the server side error code later, and if we show it here, creates an extra place we need to update.

packages/artifact/__tests__/upload.test.ts
```diff
@@ -362,7 +362,9 @@ describe('Upload Tests', () => {
     const uploadHttpClient = new UploadHttpClient()
     expect(
       uploadHttpClient.patchArtifactSize(-2, 'my-artifact')
-    ).rejects.toThrow('Unable to finish uploading artifact my-artifact')
+    ).rejects.toThrow(
+      'Patch Artifact Size failed: Artifact service responded with 400'
```
yacaovsnc, line 366: Same comment around the verb `Patch`, which is a REST implementation detail, we probably should shy away from details, but describe the symptom?

packages/artifact/src/internal/requestUtils.ts
yacaovsnc: Maybe the call was never made?

packages/artifact/src/internal/requestUtils.ts
```diff
@@ -0,0 +1,74 @@
...
+      if (isSuccessStatusCode(statusCode)) {
+        return response
+      }
+
+      // Extra error information that we want to display if a particular response code is hit
+      if (statusCode) {
+        extraErrorInformation = errorMessages.get(statusCode)
+      }
+
+      isRetryable = isRetryableStatusCode(statusCode)
```
brcrista, line 34: I think you can take this a step further to separate this into an HTTP layer and an HTTP-agnostic `retry` function:
- Pull out `getStatusCode`, `isSuccessStatusCode`, `isRetryableStatusCode` from the `retry` function and replace it with a single function you pass in called `shouldRetry`
- `retryHttpClientResponse` then can take the HTTP-related stuff and reduce it to a `shouldRetry` callback

packages/artifact/src/internal/requestUtils.ts
konradpabjan: It is possible that the response message can be undefined, and thus there is no status code. Changing this to just `number` produces the following error

image

packages/artifact/src/internal/requestUtils.ts
```diff
@@ -0,0 +1,74 @@
+import {IHttpClientResponse} from '@actions/http-client/interfaces'
```
brcrista, line 1: I saw https://github.com/actions/toolkit/blob/main/packages/cache/src/internal/requestUtils.ts which you linked.  Would we be able to share the same basic retry logic?

packages/artifact/src/internal/upload-http-client.ts
```diff
@@ -417,13 +423,13 @@ export class UploadHttpClient {
         core.info(
           `Backoff due to too many requests, retry #${retryCount}. Waiting for ${retryAfterValue} milliseconds before continuing the upload`
         )
-        await new Promise(resolve => setTimeout(resolve, retryAfterValue))
+        await sleep(retryAfterValue)
```
brcrista, line 426: If we implement exponential backoff inside `retry`, can we use `retryHttpClientResponse` here as well?

packages/artifact/src/internal/requestUtils.ts
brcrista: cc @dhadka and @joshmgross to see if we can share code with the cache action

joshmgross: I think we'd need to move the shared code to it's own package, or add this all to https://github.com/actions/http-client

yacaovsnc:
My last comment is on this call:

Single PATCH call - Update artifact size to indicate we are done (not retried)

Personally I think we should retry this call for any failures, not only on those status codes.  This is the last call to create an artifact, and it feels bad if this call fails for some infrastructure issue and all prior efforts are wasted.

packages/artifact/src/internal/requestUtils.ts
konradpabjan: A separate (small) NPM package or http-client would make sense. This feels like a long term investment though

brcrista: Ok, if it needs a separate package then I agree we can hold off.  We've tried sharing files among packages before in Azure Pipelines and it doesn't work well.

I think it could fit in in `@actions/io`, but that would mean making it a part of the public interface there.

packages/artifact/src/internal/requestUtils.ts
konradpabjan: Switched to `retryHttpClientRequest`

packages/artifact/src/internal/utils.ts
konradpabjan: Got rid of the `?`,

packages/artifact/src/internal/requestUtils.ts
konradpabjan: So, the existing methods that are retryable (the PUT calls during upload and GET calls during download, see PR description) have their own retryablity which includes exponential back-off after first checking for a `retry-after` header that we might send back if too many requests are being made. You can see it here:
https://github.com/actions/toolkit/blob/73d5917a6b5ea646ac3173cfceb727ee914ff6ed/packages/artifact/src/internal/upload-http-client.ts#L414-L432

What I did in this PR, is I moved the sleep method to a shared `util` file. The existing methods that are retried are a little bit more complicated because there are multiple calls being done concurrently so they're not switching over to the new `retryHttpClientRequest` method in this PR.  What I am planning on doing in a follow-up PR is:

- Move checking for the `retry-after` header into `util` alongside computing the exponential back-off time there
- Add exponential back-off to `retryHttpClientRequest` that is being added as part of this PR
- Switch over all the HTTP calls to use the same `retryHttpClientRequest` method. 

The last part would require a lot more refactoring and I don't want to make this PR too big so I want to do things in phases. In the past, large updates to the artifact actions would take a while to get through and they were harder to test so I want to split things up into manageable chunks

packages/artifact/src/internal/upload-http-client.ts
konradpabjan: See my earlier comment, right above in this code there already is exponential back-off, just a load of refactoring that I would like to hold-off for a future PR: https://github.com/actions/toolkit/pull/675#discussion_r545308637

packages/artifact/__tests__/upload.test.ts
konradpabjan: Removed `Patch Artifact Size` and changed it to `Finalize artifact upload`

packages/artifact/__tests__/upload.test.ts
konradpabjan: In this case I feel like it's harmless to say it's a 400. Most action authors are tech-say so this will help them know that this was caused by them

packages/artifact/src/internal/requestUtils.ts
konradpabjan: My original intention with this `errorMessages` map was so that certain http calls can produce slightly different exception messages if certain response codes are encountered. For example, during artifact creation we might get a 400 which is indicative of a invalid artifact name. However during other calls a 400 might be something totally different so this map helps the method caller customize that users will ultimately see in the logs.

packages/artifact/src/internal/requestUtils.ts
konradpabjan: I was mostly going off of the existing `retry` function that `cache` uses. I feel like pulling out `isSuccessStatusCode` and `getStatusCode` could cause some repetition throughout the rest of the code. Overall I think the current pattern is sufficient for the calls that we need to make.

packages/artifact/src/internal/requestUtils.ts
brcrista: Regardless of what the other code is doing, I think we should still implement exponential backoff in the new code.  The reason we use exponential backoff is that any fixed delay time we choose might be too short and put too much pressure on the service.

packages/artifact/src/internal/requestUtils.ts
brcrista: It shouldn't cause any repetition because `retryHttpClientRequest` will shim it.  It might work but it's not very easy to understand right now.  Separation of concerns will fix with that.

packages/artifact/src/internal/requestUtils.ts
konradpabjan: Added! Most of the code is already there, so it was pretty simple

https://github.com/actions/toolkit/blob/73d5917a6b5ea646ac3173cfceb727ee914ff6ed/packages/artifact/src/internal/utils.ts#L15-L34

packages/artifact/src/internal/requestUtils.ts
konradpabjan: Spent a bit of time trying to pull out each of the methods, but I could really get it in a nice enough format without breaking too much and I don't think it works all that well. The current `getStatusCode` method is actually structured the way it is so that it's easier to test each of the responses: 

https://github.com/actions/toolkit/blob/1bcfb42ea312c2640cde7d04a81b6c03c19025b5/packages/artifact/__tests__/retry.test.ts#L55

My original intention was to use mostly what `cache` had without changing it too much so that if there is a fix in one package the two are relatively the same and it's easy to follow.

brcrista (approved):

packages/artifact/src/internal/requestUtils.ts
```diff
@@ -0,0 +1,78 @@
...
+  let response: T | undefined = undefined
+  let statusCode: number | undefined = undefined
+  let isRetryable = false
+  let errorMessage = ''
+  let extraErrorInformation: string | undefined = undefined
+  let attempt = 1
+
+  while (attempt <= maxAttempts) {
+    try {
+      response = await operation()
```
yacaovsnc, line 27: You have removed all the `displayHttpDiagnostics` calls.  Sometimes they have valuable debug information.  How about you add it here so every http call can be traced by debug log?

konradpabjan: Added back the `displayHttpDiagnostics` if retries are exhausted or we hit something that can't be retried. Took a suprising amount of reworking but it's back

packages/artifact/__tests__/upload.test.ts
konradpabjan: Switched it up to not mention the 400

packages/artifact/src/internal/requestUtils.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    core.info(
+      `${name} - Attempt ${attempt} of ${maxAttempts} failed with error: ${errorMessage}`
+    )
+
+    await sleep(getExponentialRetryTimeInMilliseconds(attempt))
+    attempt++
+  }
+
+  if (response) {
```
yacaovsnc, line 62: I think we want to move this closer to the actual call, so we can log diagnostic info of every operation.  Otherwise LGTM.

yacaovsnc (approved):

packages/artifact/src/internal/requestUtils.ts
yacaovsnc: Line 27 in this same file.  We should just log it there.

konradpabjan: I'm going to leave it as is. This one method gets called only if we exhaust all retries so we only display the full diagnostics info in two scenarios:
- we have exhausted all retries
- a non-retryable status code was hit

Only if the call actually fails we display the full diagnostics. In all other cases I don't think it's sufficient as the response code will be displayed

Issue #681, opened 2020-12-24
tool-cache's downloadTool can't download private repo assets


KeisukeYamashita:
**Describe the bug**
Even I configure auth argument as below:

```typescript
const assetPath = await tc.downloadTool(
      'https://github.com/KeisukeYamashita/setup-release-test-repo-private/releases/download/v0.2.0/setup-release-test-repo-private_0.2.0_linux_amd64.tar.gz', <-- Private repo, okay to share
      `/tmp/tmp,
      `token ${token}`
    )
```

It will fail with the error below

```
##[debug]Failed to download from "https://github.com/KeisukeYamashita/setup-release-test-repo-private/releases/download/v0.2.0/setup-release-test-repo-private_0.2.0_linux_amd64.tar.gz". Code(404) Message(Not Found)
Error: Unexpected HTTP response: 404
##[debug]Error: Unexpected HTTP response: 404
##[debug]    at /home/runner/work/setup-release/setup-release/webpack:/typescript-action/node_modules/@actions/tool-cache/lib/tool-cache.js:100:1
##[debug]    at Generator.next (<anonymous>)
##[debug]    at fulfilled (/home/runner/work/setup-release/setup-release/webpack:/typescript-action/node_modules/@actions/tool-cache/lib/tool-cache.js:5:1)
##[debug]    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

**To Reproduce**
1. Create a PAT with `repo` scope
2. Add `tc.downloadTool` script and pass the `auth` (third) argument
3. Run the script

**Expected behavior**
Be able to download assets from private repo.

**Additional context**

I checked the log:

```
##[debug]Downloading https://github.com/KeisukeYamashita/setup-release-test-repo-private/releases/download/v0.2.0/setup-release-test-repo-private_0.2.0_linux_amd64.tar.gz
##[debug]Destination /tmp/c1cbb25c-b925-4c07-b1e5-98a9e6fd7313
##[debug]set auth
```

It contains `set auth` that means the header is set. And also, I tested with the admin PAT (has all scope) but still run into this error.

I checked that this URL `https://github.com/KeisukeYamashita/setup-release-test-repo-private/releases/download/v0.2.0/setup-release-test-repo-private_0.2.0_linux_amd64.tar.gz` is valid. I can download the asset from the browser.

![image](https://user-images.githubusercontent.com/23056537/103103685-19e55300-4666-11eb-8327-492f1048b0be.png)

[...]

PR #682, opened 2020-12-24
Fix authorization header for tool-cache

KeisukeYamashita:
## What
Bug fix.
I couldn't download private repo's assets and I realized that the auth header is `authorization` not `Authorization`.

Sorry if I misunderstanding, I think it should  `Authorization`.

that HTTP Header field names are case-sen

Should solve https://github.com/actions/toolkit/issues/681.


KeisukeYamashita:
I realized that HTTP headers are case-insensitive. The problem must be other cause. Closing this.

PR #701, opened 2021-01-29
Use GNU tar on macOS if available

Cyberbeni:
As I understant this table: https://github.com/actions/virtual-environments#available-environments
And this comment: https://github.com/actions/virtual-environments/issues/1534#issuecomment-745804159
The 20210123 images are the first ones to contain GNU tar and they are at 90%+ rollout progress.

This aims to fix multiple people's issues with caching inside macOS actions:
https://github.com/actions/cache/issues/403
https://github.community/t/bizarre-issue-with-tar-archive-for-macos-build/145377
https://github.com/Cyberbeni/install-swift-tool/issues/69

### DIFF
--- a/packages/cache/__tests__/tar.test.ts
+++ b/packages/cache/__tests__/tar.test.ts
@@ -11,6 +11,7 @@ jest.mock('@actions/exec')
 jest.mock('@actions/io')
 
 const IS_WINDOWS = process.platform === 'win32'
+const IS_MACOS = process.platform === 'darwin'
 
 function getTempDir(): string {
   return path.join(__dirname, '_temp', 'tar')
@@ -38,7 +39,7 @@ test('zstd extract tar', async () => {
     ? `${process.env['windir']}\\fakepath\\cache.tar`
     : 'cache.tar'
   const workspace = process.env['GITHUB_WORKSPACE']
-  const tarPath = 'tar'
+  const tarPath = IS_MACOS ? 'gtar' : 'tar'
 
   await tar.extractTar(archivePath, CompressionMethod.Zstd)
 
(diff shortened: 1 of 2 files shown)


smorimoto:
PRs for this is already open and it's focused on support more platforms.

Cyberbeni:
> PRs for this is already open and it's focused on support more platforms.

Yeah, that PR looks pretty messy and incorrect and abandoned.

You neglected the fact that `gnu-tar` was not even installed on macOS runners at that time (but thankfully that request went through pretty smoothly), it is also included in the path as `gtar` by default (so checking the version of all `tar`s in the path won't work) and checking for the output of version makes it way harder to intentionally override in case you have a problem with GNU tar and want to actually keep using BSD tar for caching for some reason and don't want to fork everything that uses caching in your workflow just to do that.

This pull request aims to fix exactly 1 problem (`tar` often resulting in malformed output on macOS) without making possibly breaking changes in other platforms/cases.

smorimoto:
You should know that not only Windows, but also all platforms are having problems.
I agree with the delay in the review process, but it doesn't make the review process faster just because the scope of the effect is small. (The Actions team's review process is basically slow.)

konradpabjan (approved):
Looks good to me! Thanks for putting out this PR.

I'm going to do a little bit more testing before merging this in and then I'll publish a new version of `@actions/cache` that can then be used in the our first party action.

Cyberbeni:
I've tested the most obvious edge case a while ago (restore cache saved with bsd tar) and it worked fine: https://github.com/Cyberbeni/install-swift-tool/runs/1456400103#step:4:30

(The workflow later runs the restored binary)

dhadka (approved):

konradpabjan:
Published a new release with the changes: https://www.npmjs.com/package/@actions/cache/v/1.0.6

PR #717, opened 2021-02-15
Overwrite files when extracting with GNU tar

stmcginnis:
When using GNU tar, if a file already exists, from a previous run or
another cached action, it results in a slew of warning messages that
can cause issues for the run.

Non-GNU versions appear to automatically overwrite existing files.
From the BSD tar manpage:

```
-x    Extract to disk from the archive.  If a file with the same name
      appears more than once in the archive, each copy will be
      extracted, with later copies overwriting (replacing) earlier
      copies.
```

This change adds the use of `--overwrite` when using GNU tar so that
any existing files will just be overwritten as well.

This is to help address issues encountered with golangci/golangci-lint-action
documented here:

https://github.com/golangci/golangci-lint-action/issues/135

### DIFF
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -269,6 +269,9 @@ export async function extractTar(
   if (isGnuTar) {
     // Suppress warnings when using GNU tar to extract archives created by BSD tar
     args.push('--warning=no-unknown-keyword')
+
+    // Make sure any existing files are overwritten
+    args.push('--overwrite')
   }
 
   args.push('-C', destArg, '-f', fileArg)


thboop:
Closing in favor of https://github.com/actions/toolkit/pull/807, will release a new version this week

StevenACoffman:
Hi. This doesn't appear to be fully resolved.

> As I can see the `extractTar` function is imported from `internal` package: https://github.com/actions/toolkit/blob/main/packages/cache/src/cache.ts#L5 and the internal implementation does not contain `--overwrite` flag: https://github.com/actions/toolkit/blob/fcb8c4ca798bd0d81a9b07e2f7dffee8397c4707/packages/cache/src/internal/tar.ts#L57-L86
> 
> so there is no way to fix the issue within this action, while GitHub Action uses `tar` without `--overwrite` flag.
>

hermanbanken:
Can confirm this isn't fixed.
The golangci-lint action which is using toolkit cache, can't be fixed if this isn't addressed. See: https://github.com/golangci/golangci-lint-action/issues/244

StevenACoffman:
Fairly sure no one is looking at comments on old closed issues, so a new issue and an accompanying PR is probably the best chance to get this resolved.

PR #725, opened 2021-02-24
feat(core): add getBooleanInput function

### LINKED ISSUE #723: [feature] Add new getBooleanInput function for the core
Thank you 🙇‍♀ for wanting to create an issue in this repository. Before you do, please ensure you are filing the issue in the right place. Issues should only be opened on if the issue **relates to code in this repository**.  

* If you have found a security issue [please submit it here](https://hackerone.com/github)
* If you have questions about writing workflows or action files, then please [visit the GitHub Community Forum's Actions Board](https://github.community/t5/GitHub-Actions/bd-p/actions)
* If you are having an issue or question about GitHub Actions then please [contact customer support](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#contacting-support)

If your issue is relevant to this repository, please include the information below:

**Describe the enhancement**

get the correct boolean input of workflows.

**Code Snippet**

```typescript
export function getBooleanInput(booleanInputName: string, options?: core.InputOptions) {
  const trueValue = ['true', 'True', 'TRUE', 'yes', 'Yes', 'YES', 'y', 'Y', 'on', 'On', 'ON'];
  const falseValue = ['false', 'False', 'FALSE', 'no', 'No', 'NO', 'n', 'N', 'off', 'Off', 'OFF'];
  var stringInput = core.getInput(booleanInputName, options);
  if (trueValue.indexOf(stringInput) > -1) return true;
  if (falseValue.indexOf(stringInput) > -1) return false;
  throw TypeError(`Wrong boolean input value of ${booleanInputName}`);
}
const booleanInput = getBooleanInput('booleanInput')
```

**Additional information**

https://yaml.org/type/bool.html

yi-Xu-0100:
Gets the input value of the boolean type in the YAML specification.
The return value is also in boolean type.

ref: https://yaml.org/type/bool.html

close #723

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -16,11 +16,13 @@ import * as core from '@actions/core';
 
 #### Inputs/Outputs
 
-Action inputs can be read with `getInput`.  Outputs can be set with `setOutput` which makes them available to be mapped into inputs of other actions to ensure they are decoupled.
+Action inputs can be read with `getInput` which returns a `string` or `getBooleanInput` which parses a boolean based on the [yaml 1.2 specification](https://yaml.org/spec/1.2/spec.html#id2804923). If `required` set to be false, the input should have a default value in `action.yml`.
+
+Outputs can be set with `setOutput` which makes them available to be mapped into inputs of other actions to ensure they are decoupled.
 
 ```js
 const myInput = core.getInput('inputName', { required: true });
-
+const myBooleanInput = core.getBooleanInput('booleanInputName', { required: true });
 core.setOutput('outputKey', 'outputVal');
 ```
 
@@ -65,7 +67,6 @@ catch (err) {
 ```
 
 Note that `setNeutral` is not yet implemented in actions V2 but equivalent functionality is being planned.
-
 
 #### Logging
 
@@ -118,6 +119,7 @@ const result = await core.group('Do something async', async () => {
 Colored output is supported in the Action logs via standard [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code). 3/4 bit, 8 bit and 24 bit colors are all supported.
 
 Foreground colors:
+
 ```js
 // 3/4 bit
 core.info('\u001b[35mThis foreground will be magenta')
@@ -130,6 +132,7 @@ core.info('\u001b[38;2;255;0;0mThis foreground will be bright red')
 ```
 
 Background colors:
+
 ```js
 // 3/4 bit
 core.info('\u001b[43mThis background will be yellow');
@@ -156,6 +159,7 @@ core.info('\u001b[31;46mRed foreground with a cyan background and \u001b[1mbold 
 ```
 
 > Note: Escape codes reset at the start of each line
+
 ```js
 core.info('\u001b[35mThis foreground will be magenta')
 core.info('This foreground will reset to the default')
@@ -170,9 +174,10 @@ core.info(style.color.ansi16m.hex('#abcdef') + 'Hello world!')
 
 #### Action state
 
-You can use this library to save state and get state for sharing information between a given wrapper action: 
+You can use this library to save state and get state for sharing information between a given wrapper action:
 
-**action.yml**
+**action.yml**:
+
 ```yaml
 name: 'Wrapper action sample'
 inputs:
@@ -193,6 +198,7 @@ core.saveState("pidToKill", 12345);
 ```
 
 In action's `cleanup.js`:
+
 ```js
 const core = require('@actions/core');
 
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -92,6 +92,28 @@ export function getInput(name: string, options?: InputOptions): string {
 }
 
 /**
+ * Gets the input value of the boolean type in the YAML 1.2 "core schema" specification.
+ * Support boolean input list: `true | True | TRUE | false | False | FALSE` .
+ * The return value is also in boolean type.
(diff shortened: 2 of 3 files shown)


yi-Xu-0100:
@ericsciple After I read the discussion in #362, I find that the YAML 1.2 core schema is implemented (https://github.com/actions/toolkit/pull/362#issuecomment-590555770). So could we can add the `getBooleanInput` function into the core package? 😀

WilliamMichelMSFT (approved):

packages/core/README.md
```diff
@@ -16,11 +16,11 @@ import * as core from '@actions/core';
 
 #### Inputs/Outputs
 
-Action inputs can be read with `getInput`.  Outputs can be set with `setOutput` which makes them available to be mapped into inputs of other actions to ensure they are decoupled.
+Action inputs can be read with `getInput` meanwhile the return type is `string`, and the inputs of the boolean type in the [YAML 1.2 "Core Schema" specification](https://yaml.org/spec/1.2/spec.html#id2804923) can be read with `getBooleanInput` meanwhile the return type is `boolean`. Outputs can be set with `setOutput` which makes them available to be mapped into inputs of other actions to ensure they are decoupled.
```
thboop, line 19:
```suggestion
Action inputs can be read with `getInput` which returns a `string` or `getBooleanInput` which parses a boolean based on the [yaml 1.2 specification](https://yaml.org/spec/1.2/spec.html#id2804923). Outputs can be set with `setOutput` which makes them available to be mapped into inputs of other actions to ensure they are decoupled.
```

packages/core/src/core.ts
```diff
@@ -91,6 +91,27 @@ export function getInput(name: string, options?: InputOptions): string {
...
 
+/**
+ * Gets the input value of the boolean type in the YAML 1.2 "core schema" specification.
+ * The return value is also in boolean type.
+ * ref: https://yaml.org/spec/1.2/spec.html#id2804923
+ *
+ * @param     name     name of the input to get
+ * @returns   boolean
+ */
+export function getBooleanInput(name: string): boolean {
```
thboop, line 102:
```suggestion
export function getBooleanInput(name: string, options?: InputOptions: boolean) {
```

Lets keep consistent with the `getInput` function. We really shouldn't throw if the value does not exist or is invalid, just return false. We can debug if the parsing does not match either case, to help with troubleshooting if users encounter this issue

We can add another option to `InputOptions` to set a default value, and return that instead of false if needed as well. I'm happy to tackle that in another pr if you want.

thboop (changes requested):
Thanks for the contribution @yi-Xu-0100! I left some minor feedback before merging, happy to take the pr over if you don't have the time to continue working on this!

yi-Xu-0100:
@thboop Sorry for the delayed reply, I tried to fix the conflict with the master branch, but it didn't seem to get it right. 😅 So I pulled it again, attached my changes, and pushed it forcibly. 

The change of what I did:

1. I fixed the `README` descriptions, add a suggestion about the `required` option, and try to format the content.
2. I add `options` for `getBooleanInput` function and use the `getInput` to deal with it.
3. I add the support boolean input list into the type error.
3. I add more tests for the `getBooleanInput` function.

For the error, I think the function is used to deal with boolean input, It should have a default value in `action.yml` if it is not necessary. And when it should be decided by users, it also should return a type error when users input the wrong value, which will help users to solve it without asking. What do you think? 😀

thboop:
> And when it should be decided by users, it also should return a type error when users input the wrong value, which will help users to solve it without asking.

Agreed, thanks for the contribution!

thboop (approved):
LGTM thanks for the contribution @yi-Xu-0100!

packages/core/src/core.ts
```diff
@@ -91,6 +91,28 @@ export function getInput(name: string, options?: InputOptions): string {
...
+ * @param     options  optional. See InputOptions.
+ * @returns   boolean
+ */
+export function getBooleanInput(name: string, options?: InputOptions): boolean {
+  const trueValue = ['true', 'True', 'TRUE']
+  const falseValue = ['false', 'False', 'FALSE']
+  const val = getInput(name, options)
+  if (trueValue.includes(val)) return true
+  if (falseValue.includes(val)) return false
+  throw new TypeError(
```
chamini2, line 110: what if it was `required: false`?

yi-Xu-0100: @chamini2 The `required: false` will deal with the `getInput`(Line 107) and get the default value. 👀

chamini2: But if there is **no default value**, that will make `val` an empty string, which we don't consider in line 108 or 109. Resulting in `throw new TypeError` because `trueValue.includes('')` and `falseValue.includes('')` are both false.

yi-Xu-0100: > Action inputs can be read with `getInput` which returns a `string` or `getBooleanInput` which parses a boolean based on the [yaml 1.2 specification](https://yaml.org/spec/1.2/spec.html#id2804923). If required set to be false, the input should have a default value in `action.yml`.

https://github.com/yi-Xu-0100/toolkit/blob/main/packages/core/README.md

@chamini2 Why a boolean type input have no default value? 👀

chamini2: Oh, I did not know this was expected. If `required: false` and no default I thought it had just to return `null` or `''`. Thanks, sorry for the inconvenience.

Redmishiaomi (approved):

PR #728, opened 2021-03-02
Add directory filtering to globber

MatisseHack:
This adds a new option to only return _file_ matches for a given glob pattern and skip any _directories_. I often find myself filtering directories out of the final results with subsequent `fs` calls and this change would streamline that process.

By default, the current behavior is maintained, but passing the new `{ matchDirectories: false }` option prevents the globber from returning directory matches.


thboop:
@MatisseHack thanks for the contribution! 🚀 🚀 🚀

thboop (approved):
LGTM

PR #730, opened 2021-03-06
fix: typo

sammychinedu2ky:
Awesome explanation 😋 . Just made a slight change to the verb
Changed `run` to `runs` because a singular subject is used


sammychinedu2ky:
It is actually correct. Sorry about that

sammychinedu2ky:
cheese

PR #733, opened 2021-03-10
WIP: Make the path be handled politely during path resolution

smorimoto:


smorimoto:
The tests on Windows failed, but will fix them later.

thboop:
Hey @smorimoto can you provide more details in this ticket on what you are trying to solve, for example a bit of code showing the issue?

smorimoto:
Oh, I'm still working on this and don't know if the idea will fix exactly as I expected, so just ignore this for now.

frenchvandal:
@smorimoto if you want to replace all `\` by `/` on Windows, you should use the `posix.normalize` method rather than the regular `normalize`. For example, `path.posix.normalize(archivePath)` instead of `path.normalize(archivePath)`

`normalize` will just use the platform-specific separator, i.e. a backward slash on a Windows GitHub runner. With the `posix` variant, you enforce the use of a forward slash even on a Windows context.

This is just a wild guess, maybe I am getting all wrong on what you are trying to achieve.

Source: https://nodejs.org/api/path.html#path_path_posix

thboop:
Hey @smorimoto , I'm going to close this out until its ready to be reviewed! Feel free to ping me when its ready!

PR #735, opened 2021-03-12
Note multiline match limitation in Problem Matchers documentation

electronjoe:
Add detail on one dimension of multi-line match limitations - as I only discovered this after prototyping a Problem Matcher at some expense.


thboop (approved):
lgtm

PR #760, opened 2021-04-05
@actions/artifact 0.5.1 release

konradpabjan:
Bump `@actions/http-client` to the latest version and create a new release fix a proxy related issue that is effecting some larger enterprise customers: https://github.com/actions/http-client/pull/42

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,7 +6,7 @@
     "declaration": true,
     "target": "es6",
     "sourceMap": true,
-    "lib": ["es6"]
+    "lib": ["es6","dom"]
   },
   "exclude": [
     "node_modules",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,8 +5,7 @@
     "strict": true,
     "declaration": true,
     "target": "es6",
-    "sourceMap": true,
-    "lib": ["es6","dom"]
+    "sourceMap": true
   },
   "exclude": [
     "node_modules",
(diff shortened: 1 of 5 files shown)


tsconfig.json
```diff
@@ -6,7 +6,7 @@
     "declaration": true,
     "target": "es6",
     "sourceMap": true,
-    "lib": ["es6"]
+    "lib": ["es6","dom"]
```
konradpabjan, line 9: I though this would be a normal update, but doing `npm install @actions/http-client@1.0.11` starts failing while doing `tsc`.

image

I found this solution, and I think is the best course of action: https://stackoverflow.com/questions/42603783/missing-basic-dom-types-in-typescript-project

More information about the "dom" lib here: https://www.typescriptlang.org/tsconfig#lib

brcrista: I think [DOM](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) `URL` is different from Node [URL](https://nodejs.org/api/url.html).  We have v12 of `@types/node` [installed](https://github.com/actions/toolkit/blob/fc005283374f1cea5e03ea1caf959cd9ff12fdc2/package.json#L17); I wonder why it's not picking that up?

konradpabjan: Fixed this issue by just removing `"lib": ["es6"]` completely from `tsconfig`. I had to make a small tweak in `@actions/exec` but other than that `@actions/http-client` does the same thing: https://github.com/actions/http-client/blob/main/tsconfig.json

ericsciple (approved):

ananasik365 (changes requested):
+

PR #762, opened 2021-04-06
Update releases.yml

thboop:
Remove the --dry-run flag so we can release the packages on the new ci for toolkit npm packages


konradpabjan (approved):
🚀

PR #763, opened 2021-04-06
Set npmrc to use the NPM Token

thboop:
Fixes an issue where npm fails to authenticate when publishing [example](https://github.com/actions/toolkit/actions/runs/723483158)


konradpabjan (approved):

konradpabjan (approved):
🚀

PR #767, opened 2021-04-09
Fix permissions issues using gtar on mac

dhadka:
Fixes https://github.com/actions/cache/issues/527

### DIFF
--- a/packages/cache/RELEASES.md
+++ b/packages/cache/RELEASES.md
@@ -37,3 +37,6 @@
 ### 1.0.6
 - Make caching more verbose [#650](https://github.com/actions/toolkit/pull/650)
 - Use GNU tar on macOS if available [#701](https://github.com/actions/toolkit/pull/701)
+
+### 1.0.7
+- Fixes permissions issue extracting archives with GNU tar on macOS ([issue](https://github.com/actions/cache/issues/527))
--- a/packages/cache/src/internal/tar.ts
+++ b/packages/cache/src/internal/tar.ts
@@ -26,6 +26,8 @@ async function getTarPath(
     case 'darwin': {
       const gnuTar = await io.which('gtar', false)
       if (gnuTar) {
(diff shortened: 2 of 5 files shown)


packages/cache/src/internal/tar.ts
```diff
@@ -26,6 +26,8 @@ async function getTarPath(
     case 'darwin': {
       const gnuTar = await io.which('gtar', false)
       if (gnuTar) {
+        // fix permission denied errors when extracting BSD tar archive with GNU tar - https://github.com/actions/cache/issues/527
+        args.push('--delay-directory-restore')
```
thboop, line 30: Changes look good, however we have a test that verifies what args these are called with you will need to modify.

konradpabjan:
Nit: Could you also update `releases.md` with a sentence or two

packages/cache/src/internal/tar.ts
dhadka: This does unfortunately append this arg to all invocations of `tar`, including list and create, but appears to be silently ignored in those cases.

dhadka: Tests are fixed

konradpabjan (approved):

packages/cache/RELEASES.md
```diff
@@ -36,4 +36,7 @@
 
 ### 1.0.6
 - Make caching more verbose [#650](https://github.com/actions/toolkit/pull/650)
-- Use GNU tar on macOS if available [#701](https://github.com/actions/toolkit/pull/701)
\ No newline at end of file
+- Use GNU tar on macOS if available [#701](https://github.com/actions/toolkit/pull/701)
+
+### 1.0.7
+- Fixes permissions issue extracting archives with GNU tar on macOS ([issue](https://github.com/actions/cache/issues/527))
```
thboop, line 42: NIT: Fixes permissions issue **when** extracting archives with GNU tar on macOS

thboop (approved):
LGTM

PR #770, opened 2021-04-11
Add `getCommandOutput`

### LINKED ISSUE #769: Easier to get command output
**Describe the enhancement**

There are a few Github Actions that require the command output, like [docker/build-push-action](https://github.com/docker/build-push-action/blob/master/src/exec.ts), [bjerkio/kopier](https://github.com/bjerkio/kopier/blob/9ec42c990dfd6ee394a4b16d0be2b054259552d3/src/git-commit.ts#L25-L47) and [pulumi/action](https://github.com/pulumi/actions/blob/master/src/libs/exec.ts).

I've personally added this functionality twice so far, so I thought it might be something we should consider adding to `@actions/exec` directly?

**Code Snippet**

```typescript
import { getCommandOutput } from '@actions/exec';

const output = await getCommandOutput('echo', ['hello', 'world']);

// output = "hello world"
```

cobraz:
Hello 👋

This pull request fixes #769 which is about making it easier to get the output from a command.

Few things we probably want to handle before landing:

- I didn't add a way to retrieve the exit code, do we want that?
- If we want exit code, how? As an object (e.g. `{ exitCode: number, stdout: string }`)?

### DIFF
--- a/packages/exec/RELEASES.md
+++ b/packages/exec/RELEASES.md
@@ -1,5 +1,9 @@
 # @actions/exec Releases
 
+### 1.1.0
+
+- [Add `getCommandOutput`](https://github.com/actions/toolkit/pull/770)
+
 ### 1.0.3
 
 - [Add \"types\" to package.json](https://github.com/actions/toolkit/pull/221)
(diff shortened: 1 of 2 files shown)


luketomlinson:
hi @cobraz, this is now implemented in #814!

PR #772, opened 2021-04-13
Prepend newline for set-output

ericsciple:


packages/core/src/core.ts
```diff
@@ -99,6 +99,7 @@ export function getInput(name: string, options?: InputOptions): string {
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function setOutput(name: string, value: any): void {
+  process.stdout.write(os.EOL)
```
ericsciple, line 102: i dont think it's necessary to add to all... adding to setOutput since it's more critical from a security perspective

thboop (approved):
LGTM

ericcornelissen:
I (and other, ref https://github.com/actions/toolkit/issues/777#issuecomment-825872715) would like some details on this change. Is it security-related, or not? Should Action's using [`actions/core`](https://www.npmjs.com/package/@actions/core) release an update with the new version (i.e. when upgrading from [v1.2.6](https://github.com/actions/toolkit/blob/main/packages/core/RELEASES.md#126) to [v1.2.7](https://github.com/actions/toolkit/blob/main/packages/core/RELEASES.md#127))?

devlynnx:
> _ No description provided. _

[2_Pull ghcr.ioactionsjekyll-build-pagesv1.0.8.txt](https://github.com/actions/toolkit/files/12267026/2_Pull.ghcr.ioactionsjekyll-build-pagesv1.0.8.txt)

Issue #749, opened 2021-03-26
in exec, stdline aparantly drops some lines with large outputs


tejasd1990:
Thank you 🙇‍♀ for wanting to create an issue in this repository. Before you do, please ensure you are filing the issue in the right place. Issues should only be opened on if the issue **relates to code in this repository**.  

* If you have found a security issue [please submit it here](https://hackerone.com/github)
* If you have questions about writing workflows or action files, then please [visit the GitHub Community Forum's Actions Board](https://github.community/t5/GitHub-Actions/bd-p/actions)
* If you are having an issue or question about GitHub Actions then please [contact customer support](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#contacting-support)

If your issue is relevant to this repository, please include the information below:

**Describe the bug**
We are using `exec`  in our [arm-deploy](https://github.com/azure/arm-deploy) action. Currently, we are using `stdout`  and `stdline`  listeners to collect the error and output. In `stdline` , it appears that some lines are not being received. We expect a JSON in the output, and we are getting JSON parsing errors. We confirmed this by printing whatever is received in the listener. This issue occurs persistently for large outputs, but not for small outputs. We have customers reporting this issue to us. We changed the `stdline`  to `stdout`  and so far while testing with it, we did not hit this issue. 

We used `stdline`  in the first place because since we expect the output to be in JSON format, our assumption is that the libraries could be emitting debug/command messages in between, which will break JSON parsing if we use stdout. So our strategy was to remove any debug/command messages from the output, and collect the actual JSON output.
Questions:
- Is the assumption correct? If no, then we can use `stdout`  since the issue does not seem to be happening with it.
- If the assumption is correct, will the strategy we are using(i.e. to remove the `[debug]`  and `[command]` lines) work?
- Why is the issue happening when we use `stdline` listener with large outputs, when it appears to be working with `stdout`  even with large outputs?
- If using stdline  listener, some lines are not being received because of any error in sending/listening, then can we use retry anywhere? How to track sequence of the lines?

**To Reproduce**
- Use the `arm-deploy` action in a workflow. Deploy a template which will create a large number of resoruces(The example i used to repro our customer issue was a template which 200 resources in a loop with a batchsize of 10 resources, so the output is pretty huge in this case)
- The action fails due to JSON parsing error

**Expected behavior**
- The action should succeed. It would succeed with small outputs

**Screenshots**

**Desktop (please complete the following information):**
- Any

[...]

PR #773, opened 2021-04-13
Fix broken line buffers

qRoC:
Line buffer does not work on big strings.

### DIFF
--- a/packages/exec/src/toolrunner.ts
+++ b/packages/exec/src/toolrunner.ts
@@ -453,11 +453,15 @@ export class ToolRunner extends events.EventEmitter {
             optionsNonNull.outStream.write(data)
           }
 
-          stdbuffer = this._processLineBuffer(data, stdbuffer, (line: string) => {
-            if (this.options.listeners && this.options.listeners.stdline) {
-              this.options.listeners.stdline(line)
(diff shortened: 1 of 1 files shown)


luketomlinson:
hey @qRoC, thanks for the PR! Do you have some sample output to demonstrate the issue that your changes would fix? I think it could be helpful to have a test to demonstrate the behavior. I was having trouble reproducing this behavior over in #749, so I'd be curious if you have a consistent scenario that breaks.

qRoC:
The chunk in `data` event contains either one line of output or a line with the size of the buffer.

So to reproduce this error, just print a string larger than the buffer size.

Case: `cargo build` with json output.

I fixed this in https://github.com/fibiol/cargo-actions/commit/fff2c76cd9c066e3d25b17e99052ca96439ee35c

luketomlinson:
hi @qRoC, thank you for your response. I did some more digging, and I think this is a good fix for #749. If we can add a test for this case I will feel confident approving this. If you don't want to take that on, I am happy to take a stab at my own implementation.

luketomlinson (approved):
Thanks for this PR @qRoC! I'm going to go ahead and merge and add a test later.

PR #774, opened 2021-04-13
Fixes update major version command

qRoC:


docs/action-versioning.md
```diff
@@ -42,7 +42,7 @@ Binding to the immutable full sha1 may offer more reliability.  However, note th
 
 ```
 git tag -fa v1 -m "Update v1 tag"
-git push origin v1 --force
+git push origin v1 --tags --force
```
thboop, line 45: This pushes all tags, which may be a weird side effect that people aren't expecting. Is there a reason you want to include it here? The `v1` refspec should resolve to the tag and push it

thboop:
Hey @qRoC thanks for the contribution, but I'm still trying to understand what issue this pr is trying to solve. I've let some feedback on this pr. I'm going to close this out, but please reopen and provide more context on the issue if this is something you would still like to be addressed!

PR #778, opened 2021-04-24
Remove parser option from .prettierrc.json

smorimoto:
In general, you should not set the parser option outside of the override section; it's really annoying for formatting via the editor to report errors in all files except TypeScript files which have Prettier support. If you are concerned that a format commit might contain the vulnerability, remove the commit and make the same changes by the Actions team before merging. (Of course, I didn't try that!)


smorimoto:
@sosukesuzuki Perhaps you can give your opinion on this.

sosukesuzuki (commented):
Disclaimer: I'm outsider of this project.

Looks good to me. If you want to set `parser` option for TypeScript files(For example, for using `babel-ts`), it is better to use [`overrides`](https://prettier.io/docs/en/configuration.html#configuration-overrides).

.prettierrc.json
```diff
@@ -2,10 +2,9 @@
...
   "useTabs": false,
-  "semi": false,
   "singleQuote": true,
   "trailingComma": "none",
   "bracketSpacing": false,
   "arrowParens": "avoid",
-  "parser": "typescript"
-}
\ No newline at end of file
+  "endOfLine": "auto"
```
smorimoto, line 9: See: https://prettier.io/docs/en/options.html#end-of-line

.prettierrc.json
```diff
@@ -2,10 +2,9 @@
   "printWidth": 80,
   "tabWidth": 2,
   "useTabs": false,
-  "semi": false,
```
smorimoto, line 5: Forcing the use of semicolons is fine, but forcing not to use semicolons is rude. To minimize unexpected bugs, you should absolutely avoid using these **FANCY** options.

smorimoto:
@sosukesuzuki Thanks. Good to make sure that there is no problem with the current configuration.

PR #779, opened 2021-04-24
[WIP] Update Node.js version to 16.x

smorimoto:
The version of Node.js used for testing is not strictly required to be 12, and you can gain many benefits by using 16 in the testing environment.

https://github.com/actions/toolkit/blob/e76decaf8a52bc981d17ddd3efcca548f6bc1a0e/tsconfig.json#L7

For example:
- You can get rid of Lerna by using npm@7 or higher
- You can upgrade package-lock.json to 2 by using npm@7 or higher

This is the first step to get rid of Lerna.

### DIFF
--- a/.github/workflows/artifact-tests.yml
+++ b/.github/workflows/artifact-tests.yml
@@ -24,10 +24,10 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
-    - name: Set Node.js 12.x
+    - name: Set Node.js 16.x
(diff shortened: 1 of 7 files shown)


thboop:
Hey @smorimoto , our team will be updating the node version of the toolkit and the [actions runner](https://github.com/actions/runner) sometime later this year or early next year. When we do that, we typically move to the latest stable node version.

With that in mind, I'm going to close out this pr until we do that upgrade!

Issue #629, opened 2020-11-10
getInput not working with repo-token


estahn:
**Describe the bug**

I'm trying to test my action and according to the code the input variables seem to be injected through input variables. See
https://github.com/actions/toolkit/blob/e7eb2c741847a05686a1b8df93dea4b244640b17/packages/core/src/core.ts#L86

Given the GITHUB_TOKEN is provided via `with`:

```
with:
  repo-token: ...
```

This would result in `INPUT_REPO-TOKEN` as an environment variable.  Which is invalid.

The following is therefore not possible:
```
INPUT_REPO-TOKEN=$GITHUB_TOKEN ts-node src/main.ts
```

luketomlinson:
Hi @estahn thanks for opening an issue! Do you have a small example where this is failing for you? Is this OS specific?

thboop:
Going to close this, we can reopen once we understand the issue better!

dvargas92495:
`INPUT_REPO-TOKEN` is an invalid environment variable name and therefore you cannot test your actions locally. The proposed change I think would be to replace the `-` character to a character that's valid for environment variable names. For example:

```
name.replace(/ /g, '_').replace(/-/g, '_').toUpperCase()
```

thboop:
I've tagged this as an enhancement to make testing easier.

Some shells don't support it, noticeably `bash`. But node does. If we made this change,  we would also have to update it in the runner, which is where these values are set [here](https://github.com/actions/runner/blob/be9632302ceef50bfb36ea998cea9c94c75e5d4d/src/Runner.Worker/Handlers/Handler.cs#L126). Unless we exported both values, this would be breaking for a lot of users, so we would need to have a grace period with this update.

For now, you can workaround this by setting the environment variable using `env` in bash, like `env "INPUT_REPO-TOKEN=bar" bash -c "printenv | grep INPUT_REPO-TOKEN"` or in node `process.env["INPUT_REPO-TOKEN"]=bar`

PR #780, opened 2021-04-27
core.getInput: Replace dash with underscore

el7cosmos:
#629 Dashed environment variable is invalid in most system


lastthyme (approved):

thboop:
I'm going to close this out until we have a better understanding of what you are trying to solve. We would need to make changes to actions runner as well, otherwise loading the inputs would just fail here.

PR #783, opened 2021-05-03
Release @actions/github v.5.0.0

thboop:
### Description

- This pr updates the base octokit libraries to expose the newly added rest endpoints.
- However, it also changes the definition of the octokit object, the endpoints are now accessed via `octokit.rest.{endpoint}` rather then `octokit.{endpoint}`, so we need to major version this version and ship `@actions/github v5.0.0`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19336,9 +19336,9 @@
       }
     },
     "typescript": {
-      "version": "3.7.4",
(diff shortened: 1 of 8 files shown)


luketomlinson (approved):
LGTM 🚢

PR #788, opened 2021-05-04
Add option to cp to only copy contents of directory instead of the directory itself

### LINKED ISSUE #741: Don't create basename directory if copying directory into existing one
**Current implementation/problem**

I want to copy the contents of a directory into another existing directory. Right now, because of these lines:

https://github.com/actions/toolkit/blob/4dd900dde00dcc90e5c649b688c643b1fc7684a2/packages/io/src/io.ts#L47-L51

The files/folders in the source directory get copied into another folder (the basename of the source path) inside the target directory.

For example, I want to copy the directory `.github/ISSUE_TEMPLATE/` to `issues/`. With the current version of [`@actions/io`](https://github.com/actions/toolkit/tree/main/packages/io) the files will end up in `issues/ISSUE_TEMPLATE` instead of `issues/` if `issues/` already exists.

**Describe the enhancement**

Add an option to disable this behavior and copy all files from one directory into the other without creating the basename folder.

**Code Snippet**

```js
const newDest: string =
    destStat && destStat.isDirectory() && copyInside // Add this new option
      ? path.join(dest, path.basename(source))
      : dest
```

**Additional information**

If there's another way of achieving this, please let me know. If not, I would be willing to create a PR for this.

luketomlinson:
Resolves #741 

Adds a `copySourceDirectory` field to `CopyOptions`. This allows the recursive copying of all sub folders and files without adding a nested level if the source folder name inside the target directory. 

Example:
When: 
```typescript 
copySourceDirectory: true
````

This
```typescript
io.cp('source/stuff/', 'destination', {recursive: true: copySourceDirectory: true})
```
Will produce: 
`destination/stuff/fileInsideStuff.txt`

When 
```typescript 
copySourceDirectory: false
````

This
```typescript
io.cp('source/stuff/', 'destination', {recursive: true: copySourceDirectory: false})
```

Will produce: 
`destination/fileInsideStuff.txt`


packages/io/src/io.ts
```diff
@@ -278,7 +280,11 @@ export async function findInPath(tool: string): Promise<string[]> {
 function readCopyOptions(options: CopyOptions): Required<CopyOptions> {
   const force = options.force == null ? true : options.force
   const recursive = Boolean(options.recursive)
-  return {force, recursive}
+  const copySourceDirectory =
```
thboop, line 283: why not do what force does? `const force = options.force == null ? true : options.force`
It also defaults to true.

thboop (commented):
Minor thoughts!

packages/io/src/io.ts
luketomlinson: I think `undefined` is actually the value we should check against since `== null` is doing type coercion and they just happen to be equal based on the implementation of `==` I'll update both to use `===`

thboop (approved):
LGTM Clean 🚀

PR #792, opened 2021-05-05
Cache: Delete temporary archive after upload

### LINKED ISSUE #791: Clean up temporary archive file at the end of saveCache
**Describe the enhancement**
When saving multiple caches, the temporary tar files created for each upload are left on the filesystem. With large caches the wasted space can be the difference between a successful run and an `IOException: No space left on device` error.

Consider a workflow that needs to save four separate caches containing 1GB each. The first call to saveCache creates a temporary archive, which could be up to 1GB in size (for uncompressible data). That temporary file is never deleted, so the subsequent calls have 1GB less SSD space available. By the end of the upload process there could be 4GB of wasted space that's unavailable for the rest of the workflow.

`restoreCache` already has a "finally" block to delete each temporary archive after its contents have been extracted. If this were implemented for `saveCache` the scenario above would use at most 1GB of space temporarily during upload. Once uploads were complete there would be no temporary files left on the SSD.

rcowsill:
This is to avoid filling the SSD while saving multiple large caches

Fixes #791

### DIFF
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -166,24 +166,33 @@ export async function saveCache(
 
   core.debug(`Archive Path: ${archivePath}`)
 
-  await createTar(archiveFolder, cachePaths, compressionMethod)
-  if (core.isDebug()) {
-    await listTar(archivePath, compressionMethod)
-  }
+  try {
+    await createTar(archiveFolder, cachePaths, compressionMethod)
+    if (core.isDebug()) {
+      await listTar(archivePath, compressionMethod)
+    }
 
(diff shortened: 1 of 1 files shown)


konradpabjan (approved):
Looks good! Thanks for the investigation and the PR!

thboop:
> Do I need to do anything to address the toolkit-audit failure? I haven't changed the dependencies so I expect it to pass after the merge.

No, thanks for bringing it up but we will take care of it! Feel free to ignore it on this PR.

nodoubtz (commented):
ok

Issue #737, opened 2021-03-15
exec.exec() and core.setFailed() do not fail build


fniephaus:
Our [`setup-smalltalkCI`](https://github.com/hpi-swa/setup-smalltalkCI) action installs some dependencies via `apt-get` using the following expression:

```
await exec.exec('sudo apt-get update')
```

Even though the outer `run` function contains a `try-catch` block that calls [`core.setFailed()`](https://github.com/hpi-swa/setup-smalltalkCI/blob/v1/index.js#L75) in case of errors, the build does not fail there (see [here](https://github.com/tom95/sandblocks/runs/2113619481?check_suite_focus=true#step:3:59)).

How can we make sure that builds fail early, so that our users are not confused by later errors that come, for example, from a failing `apt-get`?

Related issue: #390

/cc @tom95

luketomlinson:
Thank you for submitting an issue @fniephaus! I did some digging, but I was unable to reproduce this issue using a small example. 

```typescript
import * as core from '@actions/core'
import * as exec from '@actions/exec'

async function run(): Promise<void> {
  try {
    
    await exec.exec('./fail-now.sh')
    
  } catch (error) {
    core.setFailed(error.message)
  }
}

run()

```

fail-now.sh
```bash
echo 'Running fail-now.sh'
exit 100
```

When I run this workflow I get a failure output as expected. 
<img width="994" alt="Screen Shot 2021-05-03 at 2 41 25 PM" src="https://user-images.githubusercontent.com/19689611/116918317-becbfa00-ac1d-11eb-9a1e-a6f92bfc8c8f.png">

Is this issue still occurring for you?

fniephaus:
Thanks for checking, @luketomlinson! It turned out that an [`aware` was missing](https://github.com/hpi-swa/setup-smalltalkCI/commit/a89a05efc1acb01977b6acf27bd1fa506cacaacb) and now builds fail early when `apt-get` fails (see [here](https://github.com/hpi-swa/setup-smalltalkCI/runs/2574937793?check_suite_focus=true#step:3:10)).

Sorry for the noise!

luketomlinson:
No worries @fniephaus. Glad you got it sorted out 😄

PR #793, opened 2021-05-06
Exec: throw error when cwd option does not exist

### LINKED ISSUE #573: Passing an invalid path into `@actions.exec.exec` options argument throws confusing errror
**Describe the bug**

Calling `exec.exec(some_command, some_args, {cwd: <bad-path>}` ends up throwing an error that says `#[error]There was an error when attempting to execute the process '/bin/<some_command>'. This may indicate the process failed to start. Error: spawn /bin/<some-command> ENOENT`. I assume the ENOENT is because of the missing directory, but it seems like it's the binary command that's actually missing.

**To Reproduce**
1. Add `await exec.exec("ls", "-Fla", { cwd: "/bad/path" });` to an action that's using `@actions/exec`
2. Run the action, and see ` ::error::There was an error when attempting to execute the process '/bin/ls'. This may indicate the process failed to start. Error: spawn /bin/ls ENOENT` in the log output

**Expected behavior**

An error message along the lines of `Setting working directory to "/bad/path" which does not exist` would be more user friendly

luketomlinson:
Resolves #573 

This will throw an error when the provided `options.cwd` does not exist. I believe this will also handle #737 as we are adding a try/catch at the top level which should handle any previously uncaught errors. 

There is some linting, so I recommend turn whitespace diffs off for review.


thboop (approved):
LGTM

Issue #576, opened 2020-09-16
Request: Adding Github Host to context to enable writing actions for github enterprise


adam-bratin:
**Describe the enhancement**
Currently most actions are written to only work with repos hosted on Github.com. As Github enterprise 2.22.0 has been released with github actions beta I was hoping to enable adding to the github context a host variable. This will enable actions using context from '@actions/github' to set up a gh client consistently without the action developer needing to know where the action is being used.  
**Code Snippet**
I am assuming adding a line after here https://github.com/actions/toolkit/blob/71b19c1d65de167c9cead4be8f5c18423fd1d3fd/packages/github/src/context.ts#L45:
```typescript
this.host = process.env.GITHUB_HOST ?? "https://api.github.com" as string 
```

**Additional information**

PR #794, opened 2021-05-06
Include urls  in `@actions/github` context

thboop:
Resolves https://github.com/actions/toolkit/issues/576

We have various env variables that are set to indicate the github endpoints for rest/graphql/web. These are typically the same for dotcom users, but for ghes, being able to programmatically get that information in an action is helpful. This pr adds those values to the github context, you can see the documentation on these variables [here](https://docs.github.com/en/actions/reference/environment-variables)

### DIFF
--- a/packages/github/src/context.ts
+++ b/packages/github/src/context.ts
@@ -18,6 +18,9 @@ export class Context {
   job: string
   runNumber: number
   runId: number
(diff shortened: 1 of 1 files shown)


luketomlinson (approved):
🚢

PR #795, opened 2021-05-06
Update dependencies of tool-cache to fix `npm audit`

thboop:
This PR fixes the npm audit check by update dependencies in response to a security advisory.


thboop:
This pr is still failing for `actions/github`. A new advisory came out recently https://github.com/advisories/GHSA-43f8-2h32-f4cj and we will need to wait for jest to update. However, its only a dev dependency so we should be safe to ignore for now based on the contents of the advisory

thboop:
Jest bug: https://github.com/facebook/jest/issues/11380

luketomlinson (approved):

PR #796, opened 2021-05-06
Export isExplicitVersion and evaluateVersions

### LINKED ISSUE #244: Export evaluateVersions and isExplicitVersion functions from @actions/tool-cache
**Describe the enhancement**
Export `evaluateVersions` and `isExplicitVersion` functions from `@actions/tool-cache` by:
- renaming `_evaluateVersions` to `evaluateVersions`
- renaming `_isExplicitVersion` to `isExplicitVersion`
- exporting them

**Code Snippet**
https://github.com/actions/setup-node/blob/9bb7038a073441f1f0c3c33070fc1601a2638b1e/src/installer.ts#L118

**Additional information**
These functions are helpful to use when creating action for setting up environment with specific version of tools. Instead of directly lifting them from `@actions/tool-cache`, they could just be exported.

Thanks for considering.

luketomlinson:
Closes #244 

Export version utility functions. Could be used [here](https://github.com/actions/setup-node/blob/9bb7038a073441f1f0c3c33070fc1601a2638b1e/src/installer.ts#L118) as an example.

### DIFF
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -652,6 +652,11 @@ function _completeToolPath(tool: string, version: string, arch?: string): void {
   core.debug('finished caching tool')
 }
 
+/**
+ * Check if version string is explicit
+ *
+ * @param versionSpec      version string to check
+ */
 export function isExplicitVersion(versionSpec: string): boolean {
   const c = semver.clean(versionSpec) || ''
   core.debug(`isExplicit: ${c}`)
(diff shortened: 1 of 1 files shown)


packages/tool-cache/src/tool-cache.ts
```diff
@@ -662,7 +662,7 @@ function _isExplicitVersion(versionSpec: string): boolean {
   return valid
 }
```
thboop, line 664: We should doc these functions as well, like we do our other public functions, ie:

```
/**
 * Finds the path to a tool version in the local installed tool cache
 *
 * @param toolName      name of the tool
 * @param versionSpec   version of the tool
 * @param arch          optional arch.  defaults to arch of computer
 */
```

thboop (approved):
LGTM

PR #805, opened 2021-05-11
Prevent accidental file matching when glob contains trailing slash

luketomlinson:
Resolves #268 

Handles an odd case with `Minimatch` that we worked around before that inadvertently would match files when a trailing slash was added.

### DIFF
--- a/packages/glob/src/internal-globber.ts
+++ b/packages/glob/src/internal-globber.ts
@@ -66,23 +66,31 @@ export class DefaultGlobber implements Globber {
   async *globGenerator(): AsyncGenerator<string, void> {
     // Fill in defaults options
     const options = globOptionsHelper.getOptions(this.options)
-
+    // this.patterns.map((p) => p.segments.join('/')).forEach((p) => {
+    //   console.log(`this.Pattern: ${p}`)
+    // })
     // Implicit descendants?
     const patterns: Pattern[] = []
     for (const pattern of this.patterns) {
+      // core.debug(`=== Pattern search path: ${pattern.searchPath}`)
+      // core.debug(`=== Pattern.trailingSeparator: ${pattern.trailingSeparator}`)
       patterns.push(pattern)
       if (
         options.implicitDescendants &&
         (pattern.trailingSeparator ||
           pattern.segments[pattern.segments.length - 1] !== '**')
       ) {
+        // core.debug(`=== Pushing negate pattern`)
         patterns.push(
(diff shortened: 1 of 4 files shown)


packages/glob/src/internal-pattern.ts
```diff
@@ -132,7 +137,7 @@ export class Pattern {
       // Append a trailing slash. Otherwise Minimatch will not match the directory immediately
       // preceding the globstar. For example, given the pattern `/foo/**`, Minimatch returns
       // false for `/foo` but returns true for `/foo/`. Append a trailing slash to handle that quirk.
-      if (!itemPath.endsWith(path.sep)) {
+      if (!itemPath.endsWith(path.sep) && !this.implicitTrailing) {
```
ericsciple, line 140: implicit globstar

packages/glob/src/internal-globber.ts
```diff
@@ -102,6 +110,8 @@ export class DefaultGlobber implements Globber {
       stack.unshift(new SearchState(searchPath, 1))
     }
 
+
```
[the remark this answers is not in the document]
ericsciple: 😆

packages/glob/src/internal-pattern.ts
```diff
@@ -47,13 +47,16 @@ export class Pattern {
   // Disable no-dupe-class-members due to false positive for method overload
   // https://github.com/typescript-eslint/typescript-eslint/issues/291
 
+  private readonly implicitTrailing: boolean
+
   constructor(pattern: string)
   constructor(pattern: string, segments: undefined, homedir: string)
-  constructor(negate: boolean, segments: string[])
+  constructor(negate: boolean, segments: string[], homedir?: string, implicitTrailing?: boolean)
```
ericsciple, line 54: isImplicitPattern

ericsciple (approved):

packages/glob/src/internal-pattern.ts
```diff
@@ -144,7 +162,10 @@ export class Pattern {
 
     // Match
     if (this.minimatch.match(itemPath)) {
-      return this.trailingSeparator ? MatchKind.Directory : MatchKind.All
+      const matchValue = this.trailingSeparator
```
TingluoHuang, line 165: why do we need to create a const?

luketomlinson: good call. debugging left in

thboop (approved):
LGTM

PR #807, opened 2021-05-12
Bugfix: Fix issue with interactive unzip on Linux

### LINKED ISSUE #456: Unziping tools from cache on linux freezes the entire workflow in interactive mode waiting for user input
**Describe the bug**
When unzipping the files from the tool cache `unzip` util is run in interactive mode and when there already exists a file in the target dir it freezes the entire workflow waiting for some user input which is ridiculous.

**To Reproduce**
1. Create a workflow with `actions-rs/install` which uses the tool cache and instal `sccache` with it
and also cache the `~/.cargo/bin` dir where `sscache` utility is installed
```yml
    -  uses: actions/cache@v1
        with:
          path: ~/.cargo/bin
          key: cargo-bin
    -   uses: actions-rs/install@v0.1.2 
        with:
          crate: sccache
          use-tool-cache: true
```
2. Run this workflow a couple of times.

**Expected behavior**
I expect that the workflow is run without freezing on unzipping step. It should either replace the existing file or leave it like it is without unzipping, but never freeze the execution waiting for the user input.

**Screenshots**
image

**Additional context**
Related PRs: 
 - https://github.com/actions/toolkit/pull/418

Related issues:
- https://github.com/actions-rs/install/issues/3, 
- https://github.com/actions-rs/tool-cache/issues/5

luketomlinson:
Resolves #456

Fixes issue where unzip would freeze when asking for interactive input. 

This unifies the behavior across platforms to overwrite files instead of failing (windows, previously) or asking for user input (macOS and linux previously).


packages/tool-cache/src/tool-cache.ts
```diff
@@ -344,21 +344,49 @@ async function extractZipWin(file: string, dest: string): Promise<void> {
...
-    '-Sta',
-    '-NoProfile',
-    '-NonInteractive',
-    '-ExecutionPolicy',
-    'Unrestricted',
-    '-Command',
-    command
-  ]
-  await exec(`"${powershellPath}"`, args)
+  const pwshPath = await io.which('pwsh', false)
```
ericsciple, line 347: notes from discussion:
- test behavior of tar.gz (linux/macOS/win) and xar (macOS), and 7z (win) to determine whether they match. Good to know, not necessarily a show-stopper.
- i think overlay (and overwrite conflicts) is what we want. Might be good to get clarification from customer.
- pwsh on windows-latest
  - i think windows-latest has pwsh in the PATH
  - fallback case, you might need to hack the code (specify an invalid overload)
  - error case (i.e. where we put `$_` back into the pipeline), you can specify an invalid path
  - might need to add debug tracing so you know which path was taken
- powershell on windows-latest
  - rename pwsh out of the way (or delete it)

packages/tool-cache/src/tool-cache.ts
```diff
@@ -344,21 +344,49 @@ async function extractZipWin(file: string, dest: string): Promise<void> {
...
+      pwshCommand
+    ]
+    await exec(`"${pwshPath}"`, args)
+
+  } else {
+
+    const powershellCommand = [
+      `$ErrorActionPreference = 'Stop' ;`,
+      `try { Add-Type -AssemblyName System.IO.Compression.FileSystem } catch { } ;`,
+      `[System.IO.Compression.ZipFile]::ExtractToDirectory('${escapedFile}', '${escapedDest}')`
```
ericsciple, line 373: in the powershell.exe case, instead do `Expand-Archive -Force`

packages/tool-cache/src/tool-cache.ts
```diff
@@ -344,21 +344,55 @@ async function extractZipWin(file: string, dest: string): Promise<void> {
...
-  await exec(`"${powershellPath}"`, args)
+  const pwshPath = await io.which('pwsh', false)
+
+  //To match the file overwrite behavior on nix systems, we use the overwrite = true flag for ExtractToDirectory
+  //and the -Force flag for Expand-Archive as a fallback
+  if (pwshPath) {
+    //attempt to use pwsh with ExtractToDirectory, if this fails attempt Expand-Archive
+    const pwshCommand = [
+      `$ErrorActionPreference = 'Stop' ;`,
+      `try { Add-Type -AssemblyName System.IO.Compression.ZipFile } catch { } ;`,
```
luketomlinson, line 355: @ericsciple Should we be doing `Expand-Archive` in this empty catch block?

ericsciple: might need to catch `System.Management.Automation.RuntimeException` (method name not even found)

whereas MethodException seems to indicate overload not found

packages/tool-cache/src/tool-cache.ts
```diff
@@ -344,21 +344,55 @@ async function extractZipWin(file: string, dest: string): Promise<void> {
...
+      `try { [System.IO.Compression.ZipFile]::ExtractToDirectory('${escapedFile}', '${escapedDest}', $true) }`,
+      `catch { if ($_.Exception.GetType().FullName -eq 'System.Management.Automation.MethodException'){ Expand-Archive -LiteralPath '${escapedFile}' -DestinationPath '${escapedDest}' -Force } else { throw $_ } } ;`
```
ericsciple, line 357: might need to catch `System.Management.Automation.RuntimeException` (method name not even found)

whereas MethodException seems to indicate overload not found

packages/tool-cache/src/tool-cache.ts
```diff
@@ -344,21 +344,55 @@ async function extractZipWin(file: string, dest: string): Promise<void> {
...
+      '-Command',
+      pwshCommand
+    ]
+
+    core.debug(`Using pwsh at path: ${pwshPath}`)
+    await exec(`"${pwshPath}"`, args)
+  } else {
+    const powershellCommand = [
+      `$ErrorActionPreference = 'Stop' ;`,
+      `try { Add-Type -AssemblyName System.IO.Compression.FileSystem } catch { } ;`,
```
ericsciple, line 375: System.IO.Compression.FileSystem

packages/tool-cache/src/tool-cache.ts
```diff
@@ -344,21 +344,55 @@ async function extractZipWin(file: string, dest: string): Promise<void> {
...
+      `if ((Get-Command -Name Expand-Archive -Module Microsoft.PowerShell.Archive -ErrorAction Ignore)) { Expand-Archive -LiteralPath '${escapedFile}' -DestinationPath '${escapedDest}' -Force }`,
+      `else { try { Add-Type -AssemblyName System.IO.Compression.ZipFile } catch { } ; [System.IO.Compression.ZipFile]::ExtractToDirectory('${escapedFile}', '${escapedDest}', $true) }`
```
ericsciple, line 377: remove `try { Add-Type -AssemblyName System.IO.Compression.ZipFile } catch { } ;`

ericsciple: (that's for dotnet core)

packages/tool-cache/src/tool-cache.ts
```diff
@@ -367,6 +401,7 @@ async function extractZipNix(file: string, dest: string): Promise<void> {
   if (!core.isDebug()) {
     args.unshift('-q')
   }
+  args.unshift('-o') //overwrite with -o, otherwise a prompt is shown which freezes the run
```
ericsciple, line 404: do we need to worry about gnu unzip versus bsd? not sure whether that is a think w/ unzip or not.

ericsciple: not sure whether unzip is in the gnu toolset or just some other tool

ericsciple: we looked and this is fine

packages/tool-cache/src/tool-cache.ts
ericsciple: discussed

ericsciple (approved):

StevenACoffman:
Hi. This doesn't appear to be full resolved.

> We cannot do anything within `golangci-lint-action`, since the `@action/cache` lib is maintained by GitHub. As I can see the `extractTar` function is imported from `internal` package: https://github.com/actions/toolkit/blob/main/packages/cache/src/cache.ts#L5 and the internal implementation does not contain `--overwrite` flag: https://github.com/actions/toolkit/blob/fcb8c4ca798bd0d81a9b07e2f7dffee8397c4707/packages/cache/src/internal/tar.ts#L57-L86
> 
> so there is no way to fix the issue within this action, while GitHub Action uses `tar` without `--overwrite` flag.
>

lawyoum (commented):
fix it

PR #814, opened 2021-05-20
Add getExecOutput function

### LINKED ISSUE #769: Easier to get command output

luketomlinson:
Adds a convenience function go more easily get the output of an exec command. Rather than having to explicitly define listeners in `ExecOptions` this method will return the collected output of `stdOut` and `stdErr` in addition to the exit code. This should make usage a little less awkward in situations where users need the output of an `exec` command. 

Resolves #769

### DIFF
--- a/packages/exec/__tests__/scripts/stdoutoutputlarge.js
+++ b/packages/exec/__tests__/scripts/stdoutoutputlarge.js
@@ -1,3 +1,3 @@
 //Default highWaterMark for readable stream buffers us 64K (2^16)
 //so we go over that to get more than a buffer's worth
-process.stdout.write('a'.repeat(2**24));
+process.stdout.write('a\n'.repeat(2**24));
--- a/packages/exec/__tests__/scripts/stdoutoutputlarge.js
+++ b/packages/exec/__tests__/scripts/stdoutoutputlarge.js
@@ -1,3 +1,3 @@
 //Default highWaterMark for readable stream buffers us 64K (2^16)
 //so we go over that to get more than a buffer's worth
-process.stdout.write('a'.repeat(2**32));
+process.stdout.write('a'.repeat(2**24));
--- a/packages/exec/__tests__/scripts/stdoutoutputlarge.js
+++ b/packages/exec/__tests__/scripts/stdoutoutputlarge.js
@@ -1,0 +1,3 @@
+//Default highWaterMark for readable stream buffers us 64K (2^16)
+//so we go over that to get more than a buffer's worth
+process.stdout.write('a'.repeat(2**32));
--- a/packages/exec/src/exec.ts
+++ b/packages/exec/src/exec.ts
@@ -29,40 +29,43 @@ export async function exec(
   return runner.exec()
 }
 
-export async function getExecOutput(commandLine: string, args?: string[], options?: ExecOptions): Promise<ExecOutput> {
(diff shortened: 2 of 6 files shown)


packages/exec/__tests__/scripts/stdoutoutputlarge.js
```diff
@@ -0,0 +1,3 @@
+//Default highWaterMark for readable stream buffers us 64K (2^16)
+//so we go over that to get more than a buffer's worth
+process.stdout.write('a'.repeat(2**24));
```
thboop, line 3: Nit: can we add tests that include newlines and stderr in the output.

packages/exec/src/exec.ts
```diff
@@ -28,3 +28,44 @@ export async function exec(
...
+
+  const listeners: ExecListeners = {
+    ...options?.listeners,
+    stdout: stdOutListener,
+    stderr: stdErrListener
+  }
+
+  const exitCode = await exec(commandLine, args, {...options, listeners})
+
+  //return undefined for stdout/stderr if they are empty
```
thboop, line 65: Why undefined over the empty string? If there's no output, returning "" should be fine

packages/exec/__tests__/scripts/stdoutoutputlarge.js
```diff
@@ -0,0 +1,3 @@
+//Default highWaterMark for readable stream buffers us 64K (2^16)
```
thboop, line 1: Nit: Might be worth adding scripts to test output with newlines as well

packages/io/src/io.ts
```diff
@@ -17,7 +17,6 @@ export interface CopyOptions {
   /** Optional. Whether to copy the source directory along with all the files. Only takes effect when recursive=true and copying a directory. Default is true*/
   copySourceDirectory?: boolean
 }
-
```
thboop, line 20: Revert this change please!

thboop (dismissed):
LGTM minor thoughts

thboop (commented):
Tests appear to be failing

packages/io/src/io.ts
luketomlinson: will do. I believe it was just linting

thboop (approved):
LGTM

t1m0thyj:
The functionality introduced in this PR is very useful. For now I've copied the `getExecOutput` function into my GH action. Any idea when an update to `@actions/exec` will be published that includes this change? Thanks!

luketomlinson:
@t1m0thyj This should go out next week. See [this PR](https://github.com/actions/toolkit/pull/834)

Issue #539, opened 2020-08-03
Audit Tools Step fails due to lerna


thboop:
- [ ] Update the lerna version
- [ ] Reenable the audit-tools step

I've disabled the audit tools step as we need to update lerna or it will continue to fail. We want to continue to audit the actual packages until we have an update ready. We are waiting on lerna to publish an update.

Related: https://github.com/lerna/lerna/issues/2682

PR #815, opened 2021-05-20
Re-enable the audit tools step and update dependencies

thboop:
This pr re-enables the audit tools step and updates those dependencies. We had to make a number of changes when updating eslint, so it may be easier to go through commit by commit. All changes should just be linter changes, no actual functionality changes.

Resolves https://github.com/actions/toolkit/issues/539


luketomlinson (approved):
🥇

konradpabjan (approved):
🚀

PR #817, opened 2021-05-21
Fix flakey test with buffers on macOS

luketomlinson:


pje:
Nice! I was just about to ask you about this. :)

luketomlinson:
yep @pje classic "works on my machine" 😄

PR #826, opened 2021-06-01
Only run codeql on main branch pushes

thboop:
We already run on PR so we should catch any changes pushed by devs in this repo. Dependabot doesn't have permission to run on push, so we end up failing this job, and its kind of pointless because a pr will just be opened, so lets just skip it.


luketomlinson (approved):
💯

PR #827, opened 2021-06-02
Add test for large stdline output

luketomlinson:
A followup to #773. This test that a very large line can be sent over the `stdline` listener.


thboop (approved):
LGTM!

PR #828, opened 2021-06-04
Update escaping rules in io's rmRF

thboop:


packages/io/__tests__/io.test.ts
```diff
@@ -556,6 +556,43 @@ describe('rmRF', () => {
       await assertNotExists(symlinkFile)
       await assertNotExists(outerDirectory)
     })
+  } else {
+    it('correctly escapes % on windows', async () => {
+      const root: string = path.join(getTestTemp(), 'rmRF_escape_test_win')
+      const realDirectory: string = path.join(root, '%test%')
+      await io.mkdirP(root)
+      await io.mkdirP(realDirectory)
+      process.env['%test%'] = 'thisshouldnotresolve'
```
ericsciple, line 565: should you unset this env var afterward?

ericsciple: should this instead be `process.env['test']`?

packages/io/__tests__/io.test.ts
```diff
@@ -556,6 +556,43 @@ describe('rmRF', () => {
       await assertNotExists(symlinkFile)
       await assertNotExists(outerDirectory)
     })
+  } else {
+    it('correctly escapes % on windows', async () => {
+      const root: string = path.join(getTestTemp(), 'rmRF_escape_test_win')
+      const realDirectory: string = path.join(root, '%test%')
+      await io.mkdirP(root)
+      await io.mkdirP(realDirectory)
```
ericsciple, line 564: curious naming `realDirectory` :)

Is this copied from a symlink test?

packages/io/src/io-util.ts
```diff
@@ -166,3 +166,8 @@ function isUnixExecutable(stats: fs.Stats): boolean {
     ((stats.mode & 64) > 0 && stats.uid === process.getuid())
   )
 }
+
+// Get the path of cmd.exe in windows
+export function getCmdPath(): string {
+  return process.env['COMSPEC'] ?? `C:\\WINDOWS\\system32\\cmd.exe`
```
ericsciple, line 172: we might want to fallback to `'cmd.exe'` instead of `C:\Windows\System32\cmd.exe`

ericsciple: Since we're checking explicitly for quotes already, we should be safe to fallback to simply `cmd.exe`

packages/io/src/io.ts
```diff
@@ -117,11 +118,24 @@ export async function rmRF(inputPath: string): Promise<void> {
   if (ioUtil.IS_WINDOWS) {
     // Node doesn't provide a delete operation, only an unlink function. This means that if the file is being used by another
     // program (e.g. antivirus), it won't be deleted. To address this, we shell out the work to rd/del.
+
+    // Check for invalid characters
+    // https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file
+    if (/[*"<>|]+/.test(inputPath)) {
```
ericsciple, line 124: nit: the `+` is unnecessary and `/[*"<>|]/` is sufficient

ericsciple (approved):

packages/io/__tests__/io.test.ts
```diff
@@ -556,6 +556,45 @@ describe('rmRF', () => {
...
+      await io.mkdirP(directory)
+      const oldEnv = process.env['test']
+      process.env['test'] = 'thisshouldnotresolve'
+
+      await io.rmRF(directory)
+      await assertNotExists(directory)
+      process.env['test'] = oldEnv
+    })
+
+    it('Should throw for invalid characters', async () => {
```
luketomlinson, line 573: not at all necessary, but you can parameterize tests with 
```typescript
it.each([`*`, `"`, `<`, `>` `|`, async (invalidChar) => { })
```
 to avoid duplication

luketomlinson (approved):

PR #829, opened 2021-06-04
Sarpik/get input list support

thboop:
This is a port of [this PR](https://github.com/actions/toolkit/pull/336) with the linter ran and a few comment changes.

This pr adds a new function to the toolkit's `@actions/core` package. `getMultilineInput`

We may end up supporting yaml style lists in the future, but for now we are using multilines in place of those. In general this function will be useful for actual multiline strings and the workaround we are using for lists, so we should just support it.

Thanks @kiprasmel for coding this PR


packages/core/src/core.ts
```diff
@@ -100,6 +100,25 @@ export function getInput(name: string, options?: InputOptions): string {
...
+ * @param     options  optional. See InputOptions.
+ * @returns   string[]
+ *
+ */
+export function getMultilineInput(
+  name: string,
+  options?: InputOptions
+): string[] {
+  const inputs: string[] = getInput(name, options)
+    .split('\n')
```
luketomlinson, line 116: should we always split on `\n` or should we support `os.EOL`? I'm just curious if the current implementation might confused windows users?

thboop: It needs to be \n because that's how these get mapped into env variables in the runner! \r\n won't work!

luketomlinson (approved):

Redmishiaomi (approved):

Issue #472, opened 2020-05-21
Evaluate expressions from within an action


thisismydesign:
**Describe the enhancement**

I'm writing an action. Would like to call `${{ hashFiles(...) }}` and `${{ runner.OS }}` and get the result from within the action. Is this possible? I understand I could have these as inputs, however the purpose of my action is to simplify that process.

**Code Snippet**

Something along these lines:

`index.js`
```js
const core = require("@actions/core");

async function run() {
  const hash = core.runExpression('hashFiles(...)')
}

run();
```

thisismydesign:
Related to https://github.com/actions/toolkit/issues/430

ericsciple:
You can use the env var `RUNNER_OS`

The hashFiles source code [is here](https://github.com/actions/runner/blob/master/src/Misc/expressionFunc/hashFiles/src/hashFiles.ts)

@joshmgross @TingluoHuang should we add hashFiles to the toolkit?

@thisismydesign today the runner evaluates all expressions. Currently not consumable from javascript action. You can also add a default input with an expression, for example actions/checkout uses an expression to set the [default repository input](https://github.com/actions/checkout/blob/2ff2fbdea48a8f5da77a31e7dd5ecb46c017ffc3/action.yml#L6).

thisismydesign:
Thanks! Would be great for my use case to have access to `hashFiles`.

joshmgross:
>should we add hashFiles to the toolkit?

Absolutely. With `@actions/cache` now available as an npm package, action authors should have a way to consume `hashFiles` without requiring user input.

thisismydesign:
Awesome, that's exactly my use case :)

ericsciple:
Updated the title.

@thisismydesign as a workaround you can copy the relevant source code from the link i pasted above. It's just that one file, the only dependency is `@actions/glob` and all other dependencies are from provided by nodejs.

eine:
Since I'm writing a JS Action, I'm using [folder-hash](https://www.npmjs.com/package/folder-hash) instead of copying [hashFiles.ts](https://github.com/actions/runner/blob/master/src/Misc/expressionFunc/hashFiles/src/hashFiles.ts). Nonetheless, I'd prefer to have the function available in this toolkit.

Ref eine/setup-msys2#23

PR #830, opened 2021-06-04
Add HashFiles to the toolkit

thboop:
Resolves https://github.com/actions/toolkit/issues/472

### DIFF
--- a/packages/glob/__tests__/hash-files.test.ts
+++ b/packages/glob/__tests__/hash-files.test.ts
@@ -1,0 +1,79 @@
+import * as io from '../../io/src/io'
+import * as path from 'path'
+import {hashFiles} from '../src/glob'
+import {promises as fs} from 'fs'
+
+const IS_WINDOWS = process.platform === 'win32'
+
+/**
+ * These test focus on the ability of globber to find files
+ * and not on the pattern matching aspect
+ */
+describe('globber', () => {
+  beforeAll(async () => {
+    await io.rmRF(getTestTemp())
+  })
+
+  it('basic hashfiles test', async () => {
+    const root = path.join(getTestTemp(), 'basic-hashfiles')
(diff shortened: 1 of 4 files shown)


packages/glob/src/internal-hash-files.ts
```diff
@@ -0,0 +1,42 @@
...
+      hasMatch = true
+    }
+  }
+  result.end()
+
+  if (hasMatch) {
+    core.debug(`Found ${count} files to hash.`)
+    return result.digest('hex')
+  } else {
+    core.error(`No matches found for glob`)
```
thboop, line 39: what should we do when no files are found?

packages/glob/__tests__/hash-files.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    await createSymlinkDir(
+      path.join(root, 'realdir'),
+      path.join(root, 'symDir')
+    )
+    const testPath = path.join(root, 'symdir')
+    const hash = await hashFiles(testPath, {followSymbolicLinks: false})
+    expect(hash).toEqual('')
+  })
+})
+
```
thboop, line 84: todo add a multiline test

packages/glob/__tests__/hash-files.test.ts
```diff
@@ -0,0 +1,99 @@
...
+function getTestTemp(): string {
```
thboop, line 85: todo create a helper for these so we can share them with blob

packages/glob/src/internal-hash-files.ts
luketomlinson: IMO this feels more like a `warning` than an `error`

luketomlinson (approved):
Looks good! One comment, not a show stopper

Be-ing:
Hi, I am trying to use this in an action. Could you please publish a new version of this module to NPM?

PR #833, opened 2021-06-04
Retry artifact uploads on HTTP 500

brcrista:
We have gotten bug reports of https://github.com/actions/upload-artifact/ failing when the service returns HTTP 500, which we do not retry on currently.  An HTTP 500 may be an intermittent or instance-specific issue, so retrying here should improve reliability.

We already perform exponential backoff with jitter on retries.  We discussed whether we should lengthen the retry interval.  Right now the intervals are:

```
#1: 4.5-6.75 seconds
#2: 9-13.5 seconds
#3: 13.5-20.25 seconds
#4: 18-27 seconds
#5: 22.5-33.75 seconds
```

I think we can consider that out-of-scope for this change.  If we find that retries are not effective or putting too much pressure on the service, we can look at data and figure out what a better backoff multiplier would be.

### DIFF
--- a/packages/artifact/__tests__/download.test.ts
+++ b/packages/artifact/__tests__/download.test.ts
@@ -71,7 +71,7 @@ describe('Download Tests', () => {
     setupFailedResponse()
     const downloadHttpClient = new DownloadHttpClient()
     expect(downloadHttpClient.listArtifacts()).rejects.toThrow(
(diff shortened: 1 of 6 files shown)


yacaovsnc (approved):

brcrista:
Backed out the change to the package version; we'll do that along with the release notes when we release the update

yaananth (approved):

PR #845, opened 2021-06-14
Bump artifact package version to v0.5.2

brcrista:
Last version: https://github.com/actions/toolkit/pull/761

Changes since last version:
- https://github.com/actions/toolkit/pull/815 (no change in functionality to `artifact`)
- https://github.com/actions/toolkit/pull/833

PR #846, opened 2021-06-14
Resolve vulnerabilities found by `npm audit`

brcrista:
In https://github.com/actions/toolkit/pull/845 the workflow https://github.com/actions/toolkit/actions/workflows/audit.yml failed because of some known vulnerabilities.

Most of these I fixed with `npm audit fix`.  However, the `lerna` package, which is a dev dependency, needs to be updated ([discussion](https://github.com/actions/toolkit/pull/845#issuecomment-860785826)).  Therefore, I'm changing the workflow to run with `--production`, which excludes dev dependencies.

Sample output:

image

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -32,7 +32,8 @@ jobs:
       run: npm run bootstrap
 
     - name: audit tools
-      run: npm audit --audit-level=moderate --json | scripts/audit-allow-list
+      # `|| npm audit` to pretty-print the output if vulnerabilies are found after filtering.
+      run: npm audit --audit-level=moderate --json | scripts/audit-allow-list || npm audit --audit-level=moderate
 
     - name: audit packages
       run: npm run audit-all
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -32,7 +32,7 @@ jobs:
       run: npm run bootstrap
 
     - name: audit tools
-      run: npm audit --production --audit-level=moderate
+      run: npm audit --audit-level=moderate --json | scripts/audit-allow-list
 
     - name: audit packages
       run: npm run audit-all
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -31,7 +31,7 @@ jobs:
     - name: Bootstrap
       run: npm run bootstrap
 
-    - name: audit tools #disabled while we wait for https://github.com/actions/toolkit/issues/539
+    - name: audit tools
       run: npm audit --production --audit-level=moderate
 
     - name: audit packages
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -32,7 +32,7 @@ jobs:
       run: npm run bootstrap
 
     - name: audit tools #disabled while we wait for https://github.com/actions/toolkit/issues/539
-      run: npm audit --audit-level=moderate
+      run: npm audit --production --audit-level=moderate
 
     - name: audit packages
       run: npm run audit-all
--- a/scripts/audit-allow-list
+++ b/scripts/audit-allow-list
@@ -6,18 +6,33 @@ and writes a filtered version to stdout.
 The filtered version will have the entries listed in `AUDIT_ALLOW_LIST` removed.
 Specifically, each property of `vulnerabilities` in the input is matched by name in the allow list.
 
-Sample output of `npm audit --json`:
+Sample output of `npm audit --json` (NPM v6):
+
 {
-  // Other properties ...
-  "vulnerabilities": {
-    "meow": {
-      // Other properties ...
-    },
-    "trim-newlines": {
-      // Other properties ...
+  "actions": [
+    {
+      "action": "review",
+      "module": "trim-newlines",
+      "resolves": [
+        {
+          "id": 1753,
+          "path": "lerna>@lerna/publish>@lerna/version>@lerna/conventional-commits>conventional-changelog-core>get-pkg-repo>meow>trim-newlines",
+          "dev": true,
+          "optional": false,
+          "bundled": false
+        }
+      ]
     }
-  }
+  ],
+  // Other properties ...
 }
+
+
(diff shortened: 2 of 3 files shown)


scripts/audit-allow-list
```diff
@@ -0,0 +1,72 @@
...
+// This function assumes `audits` has the right structure.
+// Just let the error terminate the process if the input doesn't match the schema.
+const filteredAudits = filterAudits(JSON.parse(input), AUDIT_ALLOW_LIST.map(x => x.name))
+
+// This is to mimic the behavior of `npm audit`:
+const numVulnerabilities = Object.keys(filteredAudits.vulnerabilities).length
+if (numVulnerabilities > 0) {
+  console.log(filteredAudits)
+  console.log("There are some unrecognized vulnerabilities from `npm audit`.")
+  console.log("Run `npm audit` to get pretty-printed output.")
```
brcrista, line 70: One thing we are losing here is the pretty-printed table of output in the run if it fails.  We could just invoke `npm audit` here if we prefer.  However, some downsides to that is it will be the full (unfiltered) output and it won't have `--audit-level=moderate` unless we want to hardcode that.

scripts/audit-allow-list
```diff
@@ -0,0 +1,72 @@
...
+  console.error(USAGE)
+  process.exit(1)
+}
+
+// This function assumes `audits` has the right structure.
+// Just let the error terminate the process if the input doesn't match the schema.
+const filteredAudits = filterAudits(JSON.parse(input), AUDIT_ALLOW_LIST.map(x => x.name))
+
+// This is to mimic the behavior of `npm audit`:
+const numVulnerabilities = Object.keys(filteredAudits.vulnerabilities).length
```
brcrista, line 66: [Broken in CI](https://github.com/actions/toolkit/pull/846/checks?check_run_id=2822328072#step:6:8); says that `Object` is undefined / null.

We are running with Node 12, but Node 12 should support this 🤔

scripts/audit-allow-list
brcrista: Oh, I think it's `filteredAudits.vulnerabilities` that's null.  I tested on `npm` v7 locally and CI runs with `npm` v6.  The output of `npm audit` changed for v7.

scripts/audit-allow-list
brcrista: Oh, I think it's `filteredAudits.vulnerabilities` that's null.  I tested on `npm` v7 locally and CI runs with `npm` v6.  The output of `npm audit` changed for v7.

scripts/audit-allow-list
brcrista: Removing this script so no longer applies

konradpabjan:
Why is the `lockfileVersion` being bumped to version 2 in all the `package-lock.json` files? 🤔 

It's introducing quite a bit changes that I _think_ we can avoid

brcrista:
@konradpabjan that's probably because I ran `npm audit fix` with npm 7.  I'll undo and redo with npm 6.

konradpabjan (approved):
LGTM! 🚀 

@thboop could you confirm if these changes are good

.github/workflows/audit.yml
```diff
@@ -31,8 +31,8 @@ jobs:
     - name: Bootstrap
       run: npm run bootstrap
 
-    - name: audit tools #disabled while we wait for https://github.com/actions/toolkit/issues/539
-      run: npm audit --audit-level=moderate
+    - name: audit tools
+      run: npm audit --production --audit-level=moderate
```
thboop, line 35: Honestly, this isn't any different from disabling this step considering [every package is a dev dependency](https://github.com/actions/toolkit/blob/main/package.json) and that is unlikely to change. We should file an issue to reenable this once lerna is fixed

brcrista: Is `npm audit` not recursive?  The subpackages have non-dev dependencies, e.g. https://github.com/actions/toolkit/blob/11601c0d2dbb2d2cf287c77d03cc3e53866a4f56/packages/glob/package.json#L39

brcrista: And if it's not the recursive, what's the value of running this check on our dev dependencies?

thboop: `npm audit` is recursive, but the --production flag only scans things that would be used in production (only `dependencies` all the way down the tree`), while normally it scans all packages all the way down the tree. You can see how many packages it scans by doing `npm audit --production`

```                               
found 0 vulnerabilities
in 0 scanned packages
```

Just because something isn't executed in production, doesn't mean we shouldn't be wary of it. Dev dependencies may still execute when building or testing on a local machine, so if there was something like a remote code execution vulnerability that was triggered by a third party server: a developers laptop could be compromised, and their credentials stolen.

brcrista: Would you prefer the approach taken in https://github.com/actions/toolkit/commit/0607046dedda2011e068f3d961aa8a70ee2d0dbb adapted for NPM 6?  I had been working on that when I found the `--production` flag.

brcrista: switched back to that approach

konradpabjan (approved):

thboop (approved):
LGTM

scripts/audit-allow-list
```diff
@@ -0,0 +1,89 @@
...
+'use strict'
+const fs = require('fs')
+
+const USAGE = "Usage: npm audit --json | scripts/audit-allow-list"
+
+// To add entires to the allow list:
+//   - Run `npm audit --json`
+//   - Copy `path` from each `actions[k].resolves` you want to allow
+//   - Fill in the `advisoryUrl` and `justification` (these are just for documentation)
+const AUDIT_ALLOW_LIST = [
```
thboop, line 47: Nit: consider a second file in the scripts folder for this list `allowlist.json` or `allowlist_config.json`

brcrista: If the list grows very long -- well, I would consider that a smell -- it may be good to split it out.  For now I think it's fine.

PR #849, opened 2021-06-21
feat: Check if action run `hasInput`

shrink:
I am trying to determine if an input has been provided in an action where an empty string is a valid value, unfortunately an empty string is treated the same as no input by `getInput`. I would _ideally_ like to change `getInput` to distinguish between "no value" and "empty value" but that would be a breaking change: the introduction of `hasInput` is a safe middle-ground that enables this determination _without_ any breaking changes.

--- 

An alternative I considered was introducing a `default` to `InputOptions` which would help bring `InputOptions` and `action.yml` in line but I think that introduces more questions (should `trimWhitespace` be an option in `action.yml`?) so I went for the simpler option.

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -27,6 +27,12 @@ const myMultilineInput = core.getMultilineInput('multilineInputName', { required
 core.setOutput('outputKey', 'outputVal');
 ```
 
+##### `hasInput`
+
+The `getInput` method will always return a string, even in cases where a value 
+has not been provided in the Workflow. To deterine if a value has been provided
+the `hasInput` method should be used.
+
 #### Exporting variables
 
 Since each step runs in a separate process, you can use `exportVariable` to add it to this step and future steps environment blocks.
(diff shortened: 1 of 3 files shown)


TimmyTie2 (approved):

shrink:
unfortunately, it turns out this doesn't work, because of the runner's [`AddInputsToEnvironment()`](https://github.com/actions/runner/blob/be9632302ceef50bfb36ea998cea9c94c75e5d4d/src/Runner.Worker/Handlers/Handler.cs#L117-L130):

```cs
// Add the inputs to the environment variable dictionary.
foreach (KeyValuePair<string, string> pair in Inputs)
{
    AddEnvironmentVariable(
        key: $"INPUT_{pair.Key?.Replace(' ', '_').ToUpperInvariant()}",
        value: pair.Value);
}
```

I will try investigate further if it's possible to achieve this in a different way.

shrink:
Acknowledged as a bug in https://github.com/actions/runner/issues/924. I'll PR a fix for that and then bring this back.

PR #851, opened 2021-06-24
docs: better describe semantics of exec

rethab:
* again and again I see people run `exec` and check for the return value. However, this never makes sense unless you pass some options. Hopefully, adding this to the description makes this more clear.


rethab:
@thboop  what do you think? is this something that would be helpful for people?

PR #855, opened 2021-06-29
Add notice annotation and support more annotation fields

luketomlinson:
This PR adds support for:
* The `notice` annotation level. 
* The `title`, `line`, `endLine`, `col`, and `endColumn` fields. 

Depends on these [runner changes](https://github.com/actions/runner/pull/1175).


packages/core/src/core.ts
```diff
@@ -199,17 +231,28 @@ export function debug(message: string): void {
...
+}
+
+/**
+ * Adds a notice issue
+ * @param message notice issue message. Errors will be converted to string via toString()
+ * @param properties optional properties to add to the annotation. 
  */
-export function warning(message: string | Error): void {
-  issue('warning', message instanceof Error ? message.toString() : message)
+export function notice(message: string | Error, properties: AnnotationProperties = {}): void {
```
luketomlinson, line 254: I'm trying to decide what to do with the `info` message below. Should `info` proxy to `notice`? I'm thinking maybe not because we wouldn't necessarily want existing actions to submit large numbers of annotations if they make heavy use of `info`

packages/core/src/utils.ts
```diff
@@ -13,3 +16,25 @@ export function toCommandValue(input: any): string {
   }
   return JSON.stringify(input)
 }
+
+/**
+ *
+ * @param annotationProperties
+ * @returns The command properties to send with the actual annotaiton command
```
thboop, line 23:
```suggestion
 * @returns The command properties to send with the actual annotation command
```

packages/core/src/utils.ts
```diff
@@ -13,3 +16,25 @@ export function toCommandValue(input: any): string {
...
+ * See: https://github.com/actions/runner/blob/ee34f4842e747b452e13235836c92b2bb1606816/src/Runner.Worker/ActionCommandManager.cs#L566
```
thboop, line 24: link to the main branch here so this doesn't go out of date

packages/core/src/core.ts
```diff
@@ -199,17 +231,49 @@ export function debug(message: string): void {
...
+    message instanceof Error ? message.toString() : message
+  )
+}
+
+/**
+ * Adds a notice issue
+ * @param message notice issue message. Errors will be converted to string via toString()
+ * @param properties optional properties to add to the annotation.
+ */
+export function notice(
```
thboop, line 268: can you update the core readme with this function and add some more information about the overloads for column/line and such

packages/core/src/core.ts
```diff
@@ -199,17 +231,49 @@ export function debug(message: string): void {
...
+  message: string | Error,
+  properties: AnnotationProperties = {}
+): void {
+  issueCommand(
```
thboop, line 272: this won't work on GHES/GHAE until the runners and such are updated. Should we export RUNNER_VERSION as a env variable in the runner and use that to check for back-compat?

luketomlinson: The runner just ignores any unregistered commands.

packages/core/src/utils.ts
luketomlinson: yep I'll update that once the runner PR is merged

thboop (approved):
LGTM

packages/core/README.md
```diff
@@ -115,6 +117,53 @@ const result = await core.group('Do something async', async () => {
...
+
+These annotations can also be attached to particular lines and columns of your source files to show exactly where a problem is occuring. 
+
+These options are: 
+```typescript
+export interface AnnotationProperties {
+  /**
+   * A title for the annotation.
+   */
+  title?: string
```
ericsciple, line 142: does an annotation have a title and a message?

thboop: It does

packages/core/src/utils.ts
```diff
@@ -13,3 +16,25 @@ export function toCommandValue(input: any): string {
   }
   return JSON.stringify(input)
 }
+
+/**
+ *
+ * @param annotationProperties
+ * @returns The command properties to send with the actual annotation command
+ * See: https://github.com/actions/runner/blob/ee34f4842e747b452e13235836c92b2bb1606816/src/Runner.Worker/ActionCommandManager.cs#L566
```
thboop, line 24: Still need to update this!

packages/core/README.md
```diff
@@ -115,6 +117,53 @@ const result = await core.group('Do something async', async () => {
 })
 ```
 
+#### Annotations
+This library has 3 methods that will produce [annotations](https://docs.github.com/en/rest/reference/checks#create-a-check-run).
```
ericsciple, line 121: nit: line break after header

packages/core/src/command.ts
```diff
@@ -6,7 +6,7 @@ import {toCommandValue} from './utils'
 // We use any as a valid input type
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
-interface CommandProperties {
+export interface CommandProperties {
```
ericsciple, line 9: do we have to export this? do we intend for customers to use this?

if not, we might want to move into an "internal" module

packages/core/src/utils.ts
luketomlinson: yep. Hadn't pushed before you saw it 🙈

ericsciple (approved):

PR #860, opened 2021-07-05
artifact: @types/tmp should be devDependencies

shogo82148:
`@types/tmp` is type definitions for `tmp`.
It is not needed in runtime, so it should be devDependencies.


konradpabjan (approved):
Looks good! 👍 

Thank you for the PR!

PR #867, opened 2021-07-27
Toolkit package created for OIDC client

souravchanduka:
In order to get the temporary credentials for accessing cloud resources, we need to get the ID Token from the action service. This ID Token will be used by login actions of different cloud providers. Since the code to fetch the ID Token will be same for all the login actions, we have decided to create a toolkit package for it. All the login actions would import this package and get the ID token for further use.


packages/oidc-client/src/main.ts
```diff
@@ -0,0 +1,83 @@
...
+  try {
+    //Check if id token is stored in environment variable
+
+    let id_token: string = getIDTokenFromEnv()
+    if (id_token !== undefined && id_token !== '') {
+      const secondsSinceEpoch = Math.round(Date.now() / 1000)
+      const id_token_json: any = jwt_decode(id_token)
+      if ('exp' in id_token_json) {
+        if (id_token_json['exp'] - secondsSinceEpoch > 300) {
+          // Expiry time is more than 5 mins
```
navin22, line 21: by default the ID_TOKEN lifetime will be 5 min, so this will never be true.


And we have corner case here, since aud is input. we should not check only time. 
if aud is different it's a new id token the client needs.

navin22 (changes requested):

packages/oidc-client/src/main.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export async function getIDToken(audience: string): Promise<string> {
+  try {
+    //Check if id token is stored in environment variable
+
+    let id_token: string = getIDTokenFromEnv()
+    if (id_token !== undefined && id_token !== '') {
+      const secondsSinceEpoch = Math.round(Date.now() / 1000)
+      const id_token_json: any = jwt_decode(id_token)
+      if ('exp' in id_token_json) {
```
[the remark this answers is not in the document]
cpratap994: +1

packages/oidc-client/__tests__/main.test.ts
```diff
@@ -0,0 +1,32 @@
...
+describe('oidc-client-tests', () => {
+  it('Get Http Client', async () => {
+    const http = new httpclient.HttpClient('actions/oidc-client')
+    expect(http).toBeDefined()
+  })
+
+  it('HTTP get request to get token endpoint', async () => {
+    const http = new httpclient.HttpClient('actions/oidc-client')
+    const res = await http.get(
+      'https://ghactionsoidc.azurewebsites.net/.well-known/openid-configuration'
```
suyash691, line 14: @souravchanduka Do we intend to take dependancy on the POC instance and not the GH actions instance for this (at least temporarily)?
cc: @navin22

packages/oidc-client/src/main.ts
```diff
@@ -0,0 +1,83 @@
...
+        if (id_token_json['exp'] - secondsSinceEpoch > 300) {
+          // Expiry time is more than 5 mins
+          return id_token
+        }
+      } else {
+        throw new Error('Expiry time not defined in ID Token')
+      }
```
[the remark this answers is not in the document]
suyash691: We should print this in the debug log

packages/oidc-client/__tests__/main.test.ts
navin22: no, we should remove this. L0 should be fine for this. And if there are e2e tests we should follow the existing infra to call the apis.

packages/oidc-client/src/main.ts
```diff
@@ -0,0 +1,83 @@
...
+      additionalHeaders
+    )
+
+    if (!isSuccessStatusCode(response.message.statusCode)) {
+      throw new Error(
+        `Failed to get ID Token. Error message  :${response.message.statusMessage} `
+      )
+    }
+
+    const body: string = await response.readBody()
```
navin22, line 63: is it a string or some type that we need to type cast

packages/oidc-client/src/main.ts
navin22: Is there any telemetry we can capture here? does actions framework have any standard capture of telemetry?

packages/oidc-client/src/main.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    let id_token_url: string = getIDTokenUrl()
+
+    if (id_token_url === undefined) {
+      throw new Error(`ID Token URL not found`)
+    }
+    id_token_url = id_token_url + '?api-version=' + getApiVersion()
+    core.debug(`ID token url is ${id_token_url}`)
+
+    const httpclient = createHttpClient()
```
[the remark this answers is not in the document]
dwivedine: yes, it does

packages/oidc-client/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+<h3>Example:</h3>
+
+```
+const core = require('@actions/core');
+const id = require('@actions/oidc-client')
+
+
+async function getID(){
+   const id_token = await id.getIDToken('client-id', 'client-secret')
```
dwivedine, line 23: Remove client-id and client-secret. They are not passed. Add audience instead.

packages/oidc-client/README.md
```diff
@@ -0,0 +1,30 @@
...
+You can use this package to interact with the github oidc provider.
+
+<h3>Get the ID token</h3>
+
+Method Name: getIDToken
+
+<h3>Inputs</h3>
+
+audience : optional
+
```
dwivedine, line 14: Mention about the output too.

packages/oidc-client/src/main.ts
```diff
@@ -0,0 +1,83 @@
...
+export async function getIDToken(audience: string): Promise<string> {
+  try {
+    //Check if id token is stored in environment variable
+
+    let id_token: string = getIDTokenFromEnv()
+    if (id_token !== undefined && id_token !== '') {
+      const secondsSinceEpoch = Math.round(Date.now() / 1000)
+      const id_token_json: any = jwt_decode(id_token)
+      if ('exp' in id_token_json) {
+        if (id_token_json['exp'] - secondsSinceEpoch > 300) {
```
dwivedine, line 20: This value should be less than 5 mins. We initially thought to keep it as 2 mins. But it should be based on some telemetric data as mentioned by @navin22 .

packages/oidc-client/src/main.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    const httpclient = createHttpClient()
+    if (httpclient === undefined) {
+      throw new Error(`Failed to get Httpclient `)
+    }
+    core.debug(`Httpclient created ${httpclient} `) // debug is only output if you set the secret `ACTIONS_RUNNER_DEBUG` to true
+
+    const additionalHeaders = {
+      [actions_http_client.Headers.ContentType]:
+        actions_http_client.MediaTypes.ApplicationJson
```
dwivedine, line 47: should Accept: application/json also be added?

packages/oidc-client/src/main.ts
```diff
@@ -0,0 +1,83 @@
...
+    return id_token
+  } catch (error) {
+    core.setFailed(error.message)
+    return error.message
+  }
+}
+
+//module.exports.getIDToken = getIDToken
+
+getIDToken('ghactions')
```
dwivedine, line 83: Remove this line and uncomment the above line: module.exports.getIDToken = getIDToken

packages/oidc-client/src/internal/utils.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export function createHttpClient(): HttpClient {
+  return new HttpClient('actions/oidc-client', [
+    new BearerCredentialHandler(getRuntimeToken())
+  ])
+}
+
+export function getApiVersion(): string {
+  return '2.0'
+}
```
dwivedine, line 20: None of these functions should have export keyword according to me.

packages/oidc-client/README.md
```diff
@@ -0,0 +1,30 @@
+<h2>@sourav_chanduka/oidc-client</h2>
```
dwivedine, line 1: Change this to @actions/oidc-client

packages/oidc-client/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+<h3>Usage</h3>
+
+You can use this package to interact with the github oidc provider.
```
dwivedine, line 5: Please make it more descriptive. Mention that it can be used to get ID token.

packages/oidc-client/__tests__/main.test.ts
```diff
@@ -0,0 +1,32 @@
...
+  })
+
+  it('Get token endpoint', async () => {
+    let url = await configvar.getIDTokenUrl()
+    expect(url).toBeDefined()
+  })
+
+  it('Fetch Id token', async () => {
+    var id_token = main.getIDToken('helloworld')
+    expect(id_token).toBeDefined()
```
dwivedine, line 26: Is this test passing? id token should not get generated for any random audience.

PR #887, opened 2021-08-19
Added OIDC client functionality in core package

souravchanduka:
In order to get the temporary credentials for accessing cloud resources, we need to get the ID Token from the action service. This ID Token will be used by login actions of different cloud providers. Since the code to fetch the ID Token will be same for all the login actions, we can use this functionality. All the login actions would import this package and get the ID token for further use.


packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/core",
-  "version": "1.5.0",
+  "version": "1.5.1",
```
thboop, line 3: Lets go to 1.6.0. 

Please also update the https://github.com/actions/toolkit/blob/main/packages/core/RELEASES.md file with this change under the version 1.6.0

thboop:
Excited for this PR to land 🚀 . Left some minor comments on process, @luketomlinson is going to review the contents of the pr.

packages/core/src/oidc-utils.ts
```diff
@@ -0,0 +1,85 @@
...
+
+  private static getIDTokenUrl(){
+    let runtimeUrl = process.env['ACTIONS_ID_TOKEN_REQUEST_URL']
+    if (!runtimeUrl) {
+      throw new Error('Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable')
+    }
+    return runtimeUrl + '?api-version=' + OidcClient.getApiVersion()
+  }
+
+  private static async postCall(httpclient: actions_http_client.HttpClient, id_token_url: string, data: TokenRequest): Promise<string> {
```
luketomlinson, line 48: Any reason we need to pass the http client in here? Can we just create it inside here?

packages/core/src/oidc-utils.ts
```diff
@@ -0,0 +1,85 @@
...
+    const res = await httpclient.postJson<TokenResponse>(id_token_url,data).catch((error) => {
+      throw new Error(
+        `Failed to get ID Token. \n 
+        Error Code : ${error.statusCode}\n 
+        Error Message: ${error.result.message}`
+      )
+    })
+
+    const id_token = res.result?.value
```
luketomlinson, line 57: can the value every be null? If so, should we check for `if(!id_token)`?

packages/core/src/oidc-utils.ts
```diff
@@ -0,0 +1,85 @@
+import * as actions_http_client from '@actions/http-client'
+import {IRequestOptions} from '@actions/http-client/interfaces'
+import {HttpClient} from '@actions/http-client'
+import {BearerCredentialHandler} from '@actions/http-client/auth'
+import {debug, setSecret} from './core'
+
+interface TokenRequest {
+  aud: string | undefined
```
luketomlinson, line 8: I believe we can use the optional syntax here: 
```
aud?: string
```

packages/core/src/oidc-utils.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+interface TokenResponse {
+  value: string | undefined
```
luketomlinson, line 12: same here: `value?: string`

packages/core/src/core.ts
```diff
@@ -348,3 +350,7 @@ export function saveState(name: string, value: any): void {
 export function getState(name: string): string {
   return process.env[`STATE_${name}`] || ''
 }
+
+export async function getIDToken(aud?: string | undefined): Promise<string> {
```
luketomlinson, line 354: I'm not sure you need the `| undefined` when the parameter is optional.

luketomlinson (approved):
A few minor things, but no show-stoppers. Good work @souravchanduka!

luketomlinson:
@souravchanduka, @TingluoHuang is going to coordinate when this should get merged/released.

PR #899, opened 2021-09-04
Fix reading zstd version

stupidloud:
image

semver will not return the right value until fiter the raw output of `zstd --version`

### DIFF
--- a/packages/cache/src/internal/cacheUtils.ts
+++ b/packages/cache/src/internal/cacheUtils.ts
@@ -91,7 +91,7 @@ export async function getCompressionMethod(): Promise<CompressionMethod> {
   }
 
   const versionOutput = await getVersion('zstd')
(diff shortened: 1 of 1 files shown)


packages/cache/src/internal/cacheUtils.ts
```diff
@@ -91,7 +91,7 @@ export async function getCompressionMethod(): Promise<CompressionMethod> {
   }
 
   const versionOutput = await getVersion('zstd')
-  const version = semver.clean(versionOutput)
+  const version = semver.clean(versionOutput.match(/v\d\.\d\.\d/)[0])
```
rcowsill, line 94: This doesn't match if the major or minor version have more than one digit, and it only keeps the most significant digit of the patch version.

```suggestion
  const version = semver.clean(versionOutput.match(/v\d+\.\d+\.\d+/)[0])
```

yaananth:
@actions/actions-cache for review here

PR #905, opened 2021-09-07
core: Update "core.error" documentation

### LINKED ISSUE #902: core: Contradictory documentation for core.error
**Describe the bug**

The documentation for `@actions/core` has contradictory descriptions of the core.error function

**To Reproduce**
1. Go to https://github.com/actions/toolkit/tree/main/packages/core#readme
2. Scroll down to the "[logging](https://github.com/actions/toolkit/tree/main/packages/core#logging)" section
3. In the example code, see the line:
```javascript
core.error(`Error ${err}, action may still succeed though`);
```
4. Scroll down to the "[annotations](https://github.com/actions/toolkit/tree/main/packages/core#annotations)" section
5. In the example code, see the line:
```javascript
core.error('This is a bad error. This will also fail the build.')
```

**Expected behavior**
Both examples should state that core.error does not fail the build (which is the current behaviour of the runner)

 - OS: iOS
 - Browser: safari
 - Version: 22

rcowsill:
Change the `core.error` documentation to say that it won't automatically fail the action. This matches the existing example in the "logging" section.

Fixes #902

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -121,7 +121,7 @@ const result = await core.group('Do something async', async () => {
 
 This library has 3 methods that will produce [annotations](https://docs.github.com/en/rest/reference/checks#create-a-check-run). 
 ```js
-core.error('This is a bad error. This will also fail the build.')
+core.error('This is a bad error, action may still succeed though.')
 
 core.warning('Something went wrong, but it\'s not bad enough to fail the build.')
 


T3rm1:
And while the person is fixing the documentation they can also update that these annotation don't show up on the pull request page. I can see them only on the actions page.

konradpabjan (approved):

PR #858, opened 2021-07-05
tool-cache: bump uuid v8.3.2

shogo82148:
When I install `@actions/tool-cache`, I get the following warning.

```
% npm install @actions/tool-cache
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.

added 8 packages, and audited 9 packages in 5s

found 0 vulnerabilities
```

This pull request fixes it.
ref. https://www.npmjs.com/package/uuid#upgrading-from-uuid3x

PR #910, opened 2021-09-16
fix: update uuid to v10

aminya:
Closes #859 
Closes #858


iamstarkov:
node now ships uuid v4 by itself (Added in: v15.6.0, v14.17.0)
https://nodejs.org/docs/latest/api/crypto.html#cryptorandomuuidoptions

```
const crypto = require('node:crypto');
const uuid = crypto.randomUUID();
```

joshmgross:
Thank you for your contribution!

`uuid` is being removed in #1824 as part of #925, so this is no longer necessary.

Liabaemt (commented):
Run fix

PR #924, opened 2021-10-15
Fix high sev in github package

luketomlinson:
This will now just use the newer jest version from the parent `package.json`


zakariahmusic:
> 
Duplicate of #

Liabaemt (commented):
[]()

PR #935, opened 2021-11-18
Add GITHUB_REF_TYPE support to github context

DenKoren:
Hope it is all changes required to pass ref_type info from env to github context.
It's first time I touch actions toolkit and I'm not sure if I hadn't miss something.

### DIFF
--- a/packages/github/src/context.ts
+++ b/packages/github/src/context.ts
@@ -12,6 +12,7 @@ export class Context {
   eventName: string
   sha: string
   ref: string
+  ref_type: string
   workflow: string
   action: string
   actor: string
@@ -40,6 +41,7 @@ export class Context {
     this.eventName = process.env.GITHUB_EVENT_NAME as string
     this.sha = process.env.GITHUB_SHA as string
     this.ref = process.env.GITHUB_REF as string
+    this.ref_type = process.env.GITHUB_REF_TYPE as string
     this.workflow = process.env.GITHUB_WORKFLOW as string
     this.action = process.env.GITHUB_ACTION as string
     this.actor = process.env.GITHUB_ACTOR as string


peaceiris:
Why don't you add `GITHUB_REF_PROTECTED`? :)

cf. https://docs.github.com/en/actions/learn-github-actions/environment-variables

DenKoren:
Done. :)

packages/github/src/context.ts
```diff
@@ -40,6 +43,9 @@ export class Context {
     this.eventName = process.env.GITHUB_EVENT_NAME as string
     this.sha = process.env.GITHUB_SHA as string
     this.ref = process.env.GITHUB_REF as string
+    this.refName = process.env.GITHUB_REF_NAME as string
+    this.refType = process.env.GITHUB_REF_TYPE as string
+    this.refProtected = (process.env.GITHUB_REF_PROTECTED as string) === 'true'
```
dbritto-dev, line 48:
```suggestion
    this.refProtected = new Boolean(process.env.GITHUB_REF_PROTECTED as string)
```

dbritto-dev (changes requested):

dbritto-dev:
@peaceiris can we merge this ?

packages/github/src/context.ts
peaceiris: The `GITHUB_REF_PROTECTED` is expected to have true, so I think the current implementation is sufficient. We need the opinion of the maintainers about this.

cf. https://docs.github.com/en/actions/learn-github-actions/environment-variables

peaceiris:
LGTM

packages/github/src/context.ts
dbritto-dev: @DenKoren what you think ?

peaceiris: I just noticed that it causes a compilation error. We will get as follows

> Type 'Boolean' is not assignable to type 'boolean'.
>  'boolean' is a primitive, but 'Boolean' is a wrapper object. Prefer using 'boolean' when possible.

So the current implementation is good.

dbritto-dev: @peaceiris  yeah I just noticed that couple minutes ago so I added `as boolean` at the end

peaceiris: I guess that some users maybe want to control `GITHUB_REF_PROTECTED` with `true/false`. Using `Boolean()` cannot consider that use case. So the following approach (the current implementation) looks good for me.

```ts
this.refProtected = (process.env.GITHUB_REF_PROTECTED as string) === 'true'
```

DenKoren:
How to get the approval from actions/actions-runtime? Seems it is the only thing left before merge.

igreg:
FYI @GhadimiR @Link- would you be able to review/approve this PR?
The change seems relatively straightforward

DenKoren:
Polite ping. :)

PR #949, opened 2021-11-29
Updates to logging for artifact uploads

konradpabjan:
This fixes a bunch of small things that make the upload process better
- Clarifications around the upload size
- Improved debug logs around gzip behavior
- truncate millisecond time with error message during exponential back-off is in-progress
- Improve logs to be be a bit more verbose 
- Links to how to enable debug logs + link to limitations around gzip behavior

---

Improved logs (with step debugs enabled): https://github.com/konradpabjan/artifact-test/runs/4359261863?check_suite_focus=true

image

Older logs for comparison: https://github.com/konradpabjan/artifact-test/runs/4266680264?check_suite_focus=true

image

---

Closes:
https://github.com/actions/upload-artifact/issues/208

More clear error message to close out:
https://github.com/actions/upload-artifact/issues/243

### DIFF
--- a/packages/artifact/src/internal/artifact-client.ts
+++ b/packages/artifact/src/internal/artifact-client.ts
@@ -74,8 +74,7 @@ export class DefaultArtifactClient implements ArtifactClient {
   ): Promise<UploadResponse> {
     core.info(
       `Starting artifact upload
- 
-      For more detailed logs during the artifact upload process, enable step-debugging: https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging#enabling-step-debug-logging`
+For more detailed logs during the artifact upload process, enable step-debugging: https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging#enabling-step-debug-logging`
     )
     checkArtifactName(name)
 
@@ -110,7 +109,9 @@ export class DefaultArtifactClient implements ArtifactClient {
       }
 
       core.debug(`Upload Resource URL: ${response.fileContainerResourceUrl}`)
-      core.info(`Container for artifact: ${name} successfully created. Starting upload of file(s)`)
+      core.info(
+        `Container for artifact "${name}" successfully created. Starting upload of file(s)`
+      )
 
       // Upload each of the files that were found concurrently
       const uploadResult = await uploadHttpClient.uploadArtifactToFileContainer(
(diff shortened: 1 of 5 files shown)


packages/artifact/src/internal/artifact-client.ts
```diff
@@ -114,10 +122,27 @@ export class DefaultArtifactClient implements ArtifactClient {
 
       // Update the size of the artifact to indicate we are done uploading
       // The uncompressed size is used for display when downloading a zip of the artifact from the UI
+      core.info(
+        `File upload process has finished. Finalizing the artifact upload`
+      )
       await uploadHttpClient.patchArtifactSize(uploadResult.totalSize, name)
 
+      if (uploadResult.failedItems.length > 0) {
+        core.info(
```
pfleidi, line 131: Could this be a warning or is this scenario expected to happen?

pfleidi (approved):
I had one minor question. Looks great otherwise!

packages/artifact/src/internal/artifact-client.ts
konradpabjan: So there is a `continueOnError` option that we have available: https://github.com/actions/toolkit/tree/main/packages/artifact#available-options

It's not really used and we default to `false` for that so this is mainly for action authors that are doing some weird stuff. If that is `true` then it would error out earlier. The existing library also returns an array of `failedItems`: https://github.com/actions/toolkit/tree/main/packages/artifact#upload-result so it's basically up to authors to do whatever they want with that failed upload result list.

I lean on not creating a warning because it shows up as an annotation and in the past there was pushback about warning annotations we made if there were no files found with the input path

packages/artifact/src/internal/artifact-client.ts
```diff
@@ -72,6 +72,10 @@ export class DefaultArtifactClient implements ArtifactClient {
     rootDirectory: string,
     options?: UploadOptions | undefined
   ): Promise<UploadResponse> {
+    core.info(
+      `Starting artifact upload
+For more detailed logs during the artifact upload process, enable step-debugging: https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging#enabling-step-debug-logging`
```
campersau, line 77: Remove fixed language from url so it could redirect to a localized version in the future.

```suggestion
For more detailed logs during the artifact upload process, enable step-debugging: https://docs.github.com/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging#enabling-step-debug-logging`
```

PR #951, opened 2021-12-01
Check for newlines and carriage return in artifact paths and name

konradpabjan:
Enforce checking for newline characters and carriage returns in the artifact name and file paths that are uploaded

Will close out: https://github.com/actions/upload-artifact/issues/183

- Restructuring some files + tests so that all the code for validating artifact names + paths is in a single file (not in `utils.ts`) 
- Make the error output more readible, the old code would split the message into multiple lines because the actual newline characters were displayed 
- Add tests for `\\r` and `\\n` 

For invalid artifact name: https://github.com/konradpabjan/artifact-test/actions/runs/1527098177
image

For invalid path: https://github.com/konradpabjan/artifact-test/actions/runs/1527052660
image

### DIFF
--- a/packages/artifact/__tests__/path-and-artifact-name-validation.test.ts
+++ b/packages/artifact/__tests__/path-and-artifact-name-validation.test.ts
@@ -1,0 +1,78 @@
+import {
+  checkArtifactName,
+  checkArtifactFilePath
+} from '../src/internal/path-and-artifact-name-validation'
+import * as core from '@actions/core'
+
+describe('Path and artifact name validation', () => {
+  beforeAll(() => {
+    // mock all output so that there is less noise when running tests
+    jest.spyOn(console, 'log').mockImplementation(() => {})
+    jest.spyOn(core, 'debug').mockImplementation(() => {})
+    jest.spyOn(core, 'info').mockImplementation(() => {})
+    jest.spyOn(core, 'warning').mockImplementation(() => {})
+  })
+
+  it('Check Artifact Name for any invalid characters', () => {
+    const invalidNames = [
+      'my\\artifact',
+      'my/artifact',
+      'my"artifact',
+      'my:artifact',
+      'my<artifact',
+      'my>artifact',
+      'my|artifact',
+      'my*artifact',
+      'my?artifact',
+      ''
+    ]
+    for (const invalidName of invalidNames) {
+      expect(() => {
+        checkArtifactName(invalidName)
+      }).toThrow()
+    }
+
+    const validNames = [
+      'my-normal-artifact',
+      'myNormalArtifact',
+      'm¥ñðrmålÄr†ï£å¢†'
+    ]
+    for (const validName of validNames) {
+      expect(() => {
+        checkArtifactName(validName)
+      }).not.toThrow()
+    }
+  })
+
+  it('Check Artifact File Path for any invalid characters', () => {
+    const invalidNames = [
+      'some/invalid"artifact/path',
+      'some/invalid:artifact/path',
+      'some/invalid<artifact/path',
+      'some/invalid>artifact/path',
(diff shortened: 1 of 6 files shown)


packages/artifact/__tests__/path-and-artifact-name-validation.test.ts
```diff
@@ -0,0 +1,78 @@
...
+      'some/invalid"artifact/path',
+      'some/invalid:artifact/path',
+      'some/invalid<artifact/path',
+      'some/invalid>artifact/path',
+      'some/invalid|artifact/path',
+      'some/invalid*artifact/path',
+      'some/invalid?artifact/path',
+      'some/invalid\rartifact/path',
+      'some/invalid\nartifact/path',
+      'some/invalid\r\nartifact/path',
```
konradpabjan, line 58: New tests are here, I just moved everything into a separate file for better organization

packages/artifact/src/internal/path-and-artifact-name-validation.ts
```diff
@@ -0,0 +1,82 @@
...
+const invalidArtifactFilePathCharacters = new Map<string, string>([
+  ['"', 'Double quote "'],
+  [':', ' Colon :'],
+  ['<', ' Less than <'],
+  ['>', ' Greater than >'],
+  ['|', ' Vertical bar |'],
+  ['*', ' Asterisk *'],
+  ['?', ' Question mark ?'],
+  ['\r', ' Carriage return \\r'],
+  ['\n', ' Line feed \\n']
```
robherley, line 20: nit: most of these messages have spaces before them except Double quote?

packages/artifact/src/internal/path-and-artifact-name-validation.ts
```diff
@@ -0,0 +1,82 @@
...
+])
+
+const invalidArtifactNameCharacters = new Map<string, string>([
+  ...invalidArtifactFilePathCharacters,
+  ['\\', ' Backslash \\'],
```
robherley, line 25: Is the backslash escaped for a specific reason? And do we need to escape it for the display message?

robherley (approved):
👍 Just some small comments/nits

Should we worry about file name length? A lot of common fs limit to 255 iirc

packages/artifact/src/internal/path-and-artifact-name-validation.ts
konradpabjan: Yes, this has to be escaped. In general it's a special character so if you want to display it or literally express it then you need `\\`

image

packages/artifact/src/internal/path-and-artifact-name-validation.ts
konradpabjan: That's intentional. It was to make the display nicer since calling `.toString()` adds a comma between entry

This is what it looked like before the extra spaces, I initially played with adding them at the end:
image

The first one doesn't have one so the text in the annotations lines up on the left side

I've added a comment for clarification

konradpabjan: I've removed the extra space. For context I had to add spaces because calling `.toString()` made the output a bit ugly with commas right next to each other. Double quote was the first thing in the output 

image


What it looks like now.

image

packages/artifact/src/internal/path-and-artifact-name-validation.ts
robherley: oops, forgot 🤦

packages/artifact/src/internal/path-and-artifact-name-validation.ts
robherley: we prob could've done `.join(", ")` instead of the `.toString()` on the array

PR #956, opened 2021-12-06
[Artifacts] Exempt certain types of files from gzip compression

konradpabjan:
This is an optimization to speed up artifacts uploads for large files that are already compressed.

See this chart for full implementation details: https://github.com/actions/toolkit/blob/main/packages/artifact/docs/implementation-details.md#uploadcompression-flow

For certain types of files that are already compressed, we basically waste a lot of time creating temporary zips that end up being immediately discarded

Test with 2 800000000 byte (0.75GB) artifacts: 1 min 1second
https://github.com/konradpabjan/artifact-test/runs/4434777033?check_suite_focus=true

Test with existing action, the upload doesn't start until you're 1 min in or so. 2 min 22 seconds total so more than a 50% reduction in time.
https://github.com/konradpabjan/artifact-test/runs/4434790411?check_suite_focus=true

### DIFF
--- a/packages/artifact/__tests__/upload-gzip.test.ts
+++ b/packages/artifact/__tests__/upload-gzip.test.ts
@@ -1,0 +1,79 @@
+import * as core from '@actions/core'
+import * as tmp from 'tmp-promise'
+import * as path from 'path'
+import * as io from '../../io/src/io'
+import {promises as fs} from 'fs'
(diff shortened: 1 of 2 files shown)


yacaovsnc (approved):
LGTM.

pfleidi (approved):
LGTM 👍

PR #958, opened 2021-12-06
[Artifacts] Prep for @actions/artifact 0.6.0 release

konradpabjan:
Bump package version from 0.5.2 to 0.6.0 in the following files
- `package.json`
- `package-lock.json`
Document all of the changes going in
Some small tweaks that I figured could squeeze into this PR
- Changes `chunk` to `bytes` for output when displaying chunked upload process. It makes more sense for `bytes` since that is the range that is being displayed

Looks like this: https://github.com/konradpabjan/artifact-test/runs/4437446512?check_suite_focus=true
image

Remove some unnecessary diagnostic output when downloading all artifacts if debug mode is on. Before this would always show up. We only should call this method if there is an error but you can see in the screenshot we would display this even if a 200 response was received 

image

### DIFF
--- a/packages/artifact/src/internal/artifact-client.ts
+++ b/packages/artifact/src/internal/artifact-client.ts
@@ -240,7 +240,9 @@ Note: The size of downloaded zips can differ significantly from the reported siz
     while (downloadedArtifacts < artifacts.count) {
       const currentArtifactToDownload = artifacts.value[downloadedArtifacts]
       downloadedArtifacts += 1
-      core.info(`starting download of artifact ${currentArtifactToDownload.name} : ${downloadedArtifacts}/${artifacts.count}`)
+      core.info(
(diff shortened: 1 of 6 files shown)


robherley (approved):
🚀

packages/artifact/src/internal/artifact-client.ts
```diff
@@ -240,6 +240,7 @@ Note: The size of downloaded zips can differ significantly from the reported siz
     while (downloadedArtifacts < artifacts.count) {
       const currentArtifactToDownload = artifacts.value[downloadedArtifacts]
       downloadedArtifacts += 1
+      core.info(`starting download of artifact ${currentArtifactToDownload.name} : ${downloadedArtifacts}/${artifacts.count}`)
```
konradpabjan, line 243: Some extra logging that won't hurt. Looks like this: 

image

robherley: was just gonna leave a comment about this, but it's pretty useful 👍

PR #980, opened 2022-01-18
fix(tool-cache): support enterprise instances

### LINKED ISSUE #644: Make tool-cache compatible with GHES without internet access
Hello,
I am trying the preview of GitHub Actions on GitHub Enterprise Sever and found that tool-cahe was hard linked to the public GitHub:
https://github.com/actions/toolkit/blob/main/packages/tool-cache/src/tool-cache.ts#L557

Using the modified context from #576 or `process.env.GITHUB_HOST  ?? "https://api.github.com` directly would solve the issue.

I've not searched yet for other similar issues in the toolkit. This was where the toolkit died first because it didn't have internet access.

stevenh:
Remove hard code api.github.com reference in tool-cache so that it works when used in GitHub Enterprise instances.

Fixes #644

### DIFF
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -593,7 +593,8 @@ export async function getManifestFromRepo(
   branch = 'master'
 ): Promise<IToolRelease[]> {
   let releases: IToolRelease[] = []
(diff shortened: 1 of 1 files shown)


OursDesCavernes:
Hello, could we have an update on this please ?
It's an easy to fix bug that's blocking for Actions on GHES.

stevenh:
@bdehamer its been a while so just checking in to see if we could get this over the line?

PR #989, opened 2022-01-28
Trade compression ratio for speed

JJ:
According to [this benchmark](https://github.com/facebook/zstd#benchmarks), it should be 25% faster at the cost of 17% less compression ratio. This works for me (and it would be nice to experiment anyway)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1158,40 +1158,6 @@
         "node": "^10.13.0 || ^12.13.0 || ^14.15.0 || >=15.0.0"
       }
     },
-    "node_modules/@jest/console/node_modules/@jest/types": {
-      "version": "27.2.5",
-      "resolved": "https://registry.npmjs.org/@jest/types/-/types-27.2.5.tgz",
-      "integrity": "sha512-nmuM4VuDtCZcY+eTpw+0nvstwReMsjPoj7ZR80/BbixulhLaiX+fbv8oeLW8WZlJMcsGQsTmMKT/iTZu1Uy/lQ==",
-      "dev": true,
-      "dependencies": {
-        "@types/istanbul-lib-coverage": "^2.0.0",
-        "@types/istanbul-reports": "^3.0.0",
-        "@types/node": "*",
-        "@types/yargs": "^16.0.0",
-        "chalk": "^4.0.0"
-      },
-      "engines": {
-        "node": "^10.13.0 || ^12.13.0 || ^14.15.0 || >=15.0.0"
-      }
-    },
-    "node_modules/@jest/console/node_modules/@types/istanbul-reports": {
-      "version": "3.0.1",
-      "resolved": "https://registry.npmjs.org/@types/istanbul-reports/-/istanbul-reports-3.0.1.tgz",
-      "integrity": "sha512-c3mAZEuK0lvBp8tmuL74XRKn1+y2dcwOUpH7x4WrF6gk1GIgiluDRgMYQtw2OFcBvAJWlt6ASU3tSqxp0Uu0Aw==",
-      "dev": true,
-      "dependencies": {
-        "@types/istanbul-lib-report": "*"
-      }
-    },
-    "node_modules/@jest/console/node_modules/@types/yargs": {
-      "version": "16.0.4",
-      "resolved": "https://registry.npmjs.org/@types/yargs/-/yargs-16.0.4.tgz",
-      "integrity": "sha512-T8Yc9wt/5LbJyCaLiHPReJa0kApcIgJ7Bn735GjItUfh08Z1pJvu8QZqb9s+mMvKV6WUQRV7K2R46YbjMXTTJw==",
-      "dev": true,
-      "dependencies": {
-        "@types/yargs-parser": "*"
-      }
-    },
     "node_modules/@jest/console/node_modules/ansi-styles": {
       "version": "4.3.0",
       "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-4.3.0.tgz",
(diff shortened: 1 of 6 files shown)


brcrista:
Adding @actions/actions-cache for this area.  I'll go update the CODEOWNERS as well.

bishal-pdMSFT:
+ @vsvipul

bishal-pdMSFT:
@JJ although it may result in faster compression, it will mean bigger package and hence more latency in upload/download and will occupy more space in storage quota of 10 GB for a repo.

JJ:
> @JJ although it may result in faster compression, it will mean bigger package and hence more latency in upload/download and will occupy more space in storage quota of 10 GB for a repo.

Of course, there's a tradeoff. As I say, that tradeoff works for me. I would say most people are caching for speed, and there will not be so many that are 17% or less below the quota. Take into account that 1Gb can take now ~50 seconds to decompress. Cutting that to 38 seconds would really be awesome.

Alternatively, it could be left as an option, with the current as a default.

bishal-pdMSFT:
Agree. I am more worried about network latency. It would be good to get some test data around this with some sample workflows runs with and without fast compression.

bishal-pdMSFT:
The [benchmark](https://github.com/facebook/zstd#benchmarks) mentions compression to improve by ~450 Mbps. The network bandwidth available on a runner is seems to be around [1500 mbps](https://docs.microsoft.com/en-us/azure/virtual-machines/dv2-dsv2-series#dsv2-series). So theoretically it seems network latency should not be a problem.

JJ:
Right, need to change tests. Can do that, no problem.

vsvipul:
@JJ I ran some tests on a private repo and the results were not very supportive of going ahead with this. 

For a single cache file of size 3.4 GB, the results were- 
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-df0e09aa-7fff-d0bf-e225-4a39e5bafe00"><div dir="ltr" style="margin-left:0pt;" align="left">

  | Normal Execution | With flag fast=1
-- | -- | --
Compress | 91s | 87s
Decompress | 57s | 96s
Size of zip created | ~3200 MB (3355520407 B) | ~3200 MB (3355520389 B)

</div></b>

For 130 files of total size 6.5 GB, the results were-
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-89c5d8d5-7fff-795d-5065-cd38f8961940"><div dir="ltr" style="margin-left:0pt;" align="left">

  | Normal Execution | With flag fast=1
-- | -- | --
Compress | 157s | 206s
Decompress | 113s | 134s
Size of zip created | 5016 MB | 6653 MB

</div></b>

As you can see, the performance degrades instead of improving it, both in time and space.
I used this workflow to run this testing -

```
name: Testing Speed and Space Fast 2

on: [push, workflow_dispatch]

jobs:
  CreateV2:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v1.1.0

    - name: Create Cache Folder
      run: |
        mkdir cache-folder
    - uses: actions/cache@72d1e4fdff0ff7b1b6e86b415f2d4f5941e5c006
      id: cache
      with:
        path: cache-folder
        key: ${{ runner.os }}-${{ github.sha }}-key

    - name: Creating Content
      if: steps.cache.outputs.cache-hit != 'true'
      run: |
        for y in {0..130};do openssl rand -out cache-folder/data$y  -base64 $(( 50 * 2**20 * 3/4 )); done
        du -sh cache-folder
  RestoreV2:
    needs: CreateV2
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v1.1.0

    - name: Create Cache Folder
      run: |
        mkdir cache-folder
    - uses: actions/cache@72d1e4fdff0ff7b1b6e86b415f2d4f5941e5c006
      id: cache
      with:
        path: cache-folder
        key: ${{ runner.os }}-${{ github.sha }}-key

    - name: Check Cache Hit
      if: steps.cache.outputs.cache-hit != 'true'
      run: |
        echo "::error ::Failed to Restore Cache"
        exit 1
    - name: Validate Cache Content
      if: steps.cache.outputs.cache-hit == 'true'
      run: |
        cd cache-folder
        echo "cache hit"
```

You can try using the same or anyother custom workflow with the commit id `72d1e4fdff0ff7b1b6e86b415f2d4f5941e5c006` which contains this flag change, and would be curious to see your results as well. 

For now, I don't think we should go ahead with merging this. If there is some benefit of doing this, it is very partial, and not visible in testing for me at least.

cc/ @bishal-pdMSFT

JJ:
(OK; the fact that compressed files take *more* than original files points to a high degree of randomness, which is obvious since you're generating files randomly. That might not *always* be the case in real life, but it's definitely a worst-case scenario, and in that worst-case scenario you're risking making compression *and* decompression slower, not to mention using a lot of storage)

vsvipul:
For 130 files of total size 666 MB
<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-cc0a7a8f-7fff-df88-c38e-6b9cd566ee4d"><div dir="ltr" style="margin-left:0pt;" align="left">

  | Normal Execution | With flag fast=1
-- | -- | --
Compress | 14s | 25s
Decompress | 11s | 7s
Size of zip created | 502 MB | 665 MB

</div><br /></b>

@JJ You can test your real world scenario as well using the commit id `47ad7e4077b36ae4030a640c4613f53ede7e8e6a` and we can see what improvements you see. Then we can decide if it produces any significant improvements. Thank you.

JJ:
This last one is improving decompression by almost 40% while increasing size by 20%. I could live with that. But yes, I'll try and give it a try.

PR #991, opened 2022-02-01
Update ms-rest-js and storage-blog dependencies for cache

vsvipul:
Updating these dependencies suppresses the dependabot alert for node-fetch we are getting in [actions/cache](https://github.com/actions/cache).

Alert - `node-fetch is vulnerable to Exposure of Sensitive Information to an Unauthorized Actor`
Affected versions - `< 2.6.7`
Patched Version - `2.6.7`

We'll also need to bump up the patch version on npm registry - https://www.npmjs.com/package/@actions/cache .

### DIFF
--- a/packages/cache/RELEASES.md
+++ b/packages/cache/RELEASES.md
@@ -43,3 +43,7 @@
 
 ### 1.0.8
 - Increase the allowed artifact cache size from 5GB to 10GB ([issue](https://github.com/actions/cache/discussions/497))
+
+### 1.0.9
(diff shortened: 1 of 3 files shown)


bishal-pdMSFT:
Looks good to me! But lets check if @brcrista or someone else from @actions/actions-service would like to take a look.

bishal-pdMSFT (approved):

thboop (approved):
LGTM

PR #1008, opened 2022-03-01
Update `lockfileVersion` in `package-lock.json` for `@actions/artifact`

jtamsut:
This change addresses [this](https://github.com/github/c2c-actions-checks/issues/270) issue.

Documentation for the semantics of `lockfileVersion` can be found [here](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#lockfileversion).

`lockfileVersion: 2` is corresponds to an npm `v7` (but is also backwards compatible with npm `v5` and `v6`).

### DIFF
--- a/packages/artifact/package-lock.json
+++ b/packages/artifact/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "@actions/artifact",
   "version": "0.6.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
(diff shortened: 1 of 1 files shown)


konradpabjan (approved):
LGTM! 🚀 

cc @thboop we're bumping up the lockFileVersion for `@actions/artifact` just as an FYI. I see a few other packages like `@actions/core` are already on it

PR #1025, opened 2022-03-16
Update `lockfileVersion` in `package-lock.json` in `tool-cache` package

jtamsut:
Closes github/c2c-actions-checks#293

### DIFF
--- a/packages/cache/RELEASES.md
+++ b/packages/cache/RELEASES.md
@@ -47,3 +47,6 @@
 ### 1.0.9
   - Use @azure/ms-rest-js v2.6.0
   - Use @azure/storage-blob v12.8.0
+
+### 1.0.10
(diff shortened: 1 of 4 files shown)


packages/cache/RELEASES.md
```diff
@@ -46,4 +46,4 @@
 
 ### 1.0.9
   - Use @azure/ms-rest-js v2.6.0
-  - Use @azure/storage-blob v12.8.0
+  - Use @azure/storage-blob v12.8.0
```
konradpabjan, line 49: What's with this `cache` change showing up in `tool-cache` 🤔

jtamsut: Hmm maybe that's like a newline diff or something. The text is the same.

PR #1028, opened 2022-03-22
Adding support in cache package to check if Artifact Cache service is enabled or not

tiwarishub:
This PR introduced a function that can be used to check the presence of Artifact Cache service using the presence of ACTION_CACHE_URL env var. This var we only set in runner if Artifact Cache is present. 

These are changes required on the package consumer side: https://github.com/actions/cache/pull/774

We have tested it for GHES, where AC is only available in version >= 3.5

In 3.4 where AC is not present 
image

In GHES, where AC is present
image

Linked: https://github.com/github/c2c-actions/issues/4065


packages/cache/__tests__/__fixtures__/index.js
```diff
@@ -3,10 +3,10 @@
 const fs = require('fs');
 const os = require('os');
 const filePath = process.env[`GITHUB_ENV`]
-fs.appendFileSync(filePath, `ACTIONS_RUNTIME_URL=${process.env.ACTIONS_RUNTIME_URL}${os.EOL}`, {
+fs.appendFileSync(filePath, `ACTIONS_RUNTIME_TOKEN=${process.env.ACTIONS_RUNTIME_TOKEN}${os.EOL}`, {
     encoding: 'utf8'
 })
-fs.appendFileSync(filePath, `ACTIONS_RUNTIME_TOKEN=${process.env.ACTIONS_RUNTIME_TOKEN}${os.EOL}`, {
+fs.appendFileSync(filePath, `ACTIONS_CACHE_URL=${process.env.ACTIONS_CACHE_URL}${os.EOL}`, {
```
aparna-ravindra, line 9: is this change required?

packages/cache/RELEASES.md
```diff
@@ -49,4 +49,7 @@
   - Use @azure/storage-blob v12.8.0
 
 ### 1.0.10
-- Update `lockfileVersion` to `v2` in `package-lock.json [#1022](https://github.com/actions/toolkit/pull/1022) 
\ No newline at end of file
+- Update `lockfileVersion` to `v2` in `package-lock.json [#1022](https://github.com/actions/toolkit/pull/1022) 
+
+### 1.0.11
+- Added support to check if Artifact cache service feature is available or not.
```
aparna-ravindra, line 55: please link the PR to this line for future reference

packages/cache/__tests__/__fixtures__/index.js
tiwarishub: yes this change is required for CI test to set proper env

packages/cache/RELEASES.md
tiwarishub: done

aparna-ravindra:
LGTM

packages/cache/src/cache.ts
```diff
@@ -43,6 +43,16 @@ function checkKey(key: string): void {
...
 }
 
+/**
+ * isFeatureAvailable to check the presence of Actions cache service
+ *
+ * @returns boolean return true if Actions cache service feature is available, otherwise false
+ */
+
+export function isFeatureAvailable(): boolean {
+  return utils.isFeatureAvailable()
```
brcrista, line 53: Is it useful to put the function in `utils` instead of just inlining it here?

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -123,3 +123,11 @@ export function assertDefined<T>(name: string, value?: T): T {
...
   return value
 }
+
+export function isFeatureAvailable(): boolean {
+  if (process.env['ACTIONS_CACHE_URL']) {
+    return true
+  }
+
+  return false
+}
```
brcrista, line 133:
```suggestion
export function isFeatureAvailable(): boolean {
  return !!process.env['ACTIONS_CACHE_URL']
}
```

☝️ this is a common JavaScript idiom for returning a value's truthiness as a bool

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -31,12 +31,7 @@ import {
 const versionSalt = '1.0'
 
 function getCacheApiUrl(resource: string): string {
-  // Ideally we just use ACTIONS_CACHE_URL
-  const baseUrl: string = (
-    process.env['ACTIONS_CACHE_URL'] ||
-    process.env['ACTIONS_RUNTIME_URL'] ||
```
brcrista, line 37: Breaking change here.  If we're going to remove this, we need to bump the package version to `2.0.0`

tiwarishub: This should not be a breaking change as we are already setting ACTIONS_CACHE_URL in runner based on the presence of Artifact Cache service https://github.com/github/actions-dotnet/blob/9925858070aba79b8c5f8be793f58e1117ecd03d/Actions/Runtime/Sdk/Server/TaskHub.cs#L3484  and https://github.com/actions/runner/blob/main/src/Runner.Worker/Handlers/NodeScriptActionHandler.cs#L53-L56

tiwarishub: @aiqiaoy  for 👀

packages/cache/src/cache.ts
brcrista: Personally I would opt for keeping the app code simple and pushing complexity into the test code, but it's not a big deal.

You can also simplify the current code like

```
export const isFeatureAvailable = utils.isFeatureAvailable
```

packages/cache/src/internal/cacheHttpClient.ts
bishal-pdMSFT: While technically yes it is a breaking change, pretty sure it is not so in practical terms. In hosted Actions, `ACTIONS_CACHE_URL` is set and no-one would be using the fallback. In GHES, caching was not available and hence need not worry about older server/runner versions.
Hence we should be okay to do without extra work of a major version change.

tiwarishub: updated for major release

packages/cache/src/cache.ts
tiwarishub: it makes sense to keep the app code simple. Inlined the function and moved to test cases to the new file.

vsvipul (approved):

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -31,12 +31,7 @@ import {
...
-    ''
-  ).replace('pipelines', 'artifactcache')
+  const baseUrl: string = process.env['ACTIONS_CACHE_URL'] || ''
```
bishal-pdMSFT, line 34: Do we need `|| ''`?

bishal-pdMSFT (approved):

packages/cache/src/internal/cacheHttpClient.ts
tiwarishub: Actually we don't need it but safeguard was there from beginning so i didn't remove it

PR #1030, opened 2022-03-23
Cap the cache download chunk to 2 GB

ashwinsangem:
We see an [issue](https://github.com/actions/cache/issues/773#issuecomment-1075284784)  when the download chunk goes beyond 2 GB with the new version of cache action. So, mitigating it to avoid further escalation.


ashwinsangem:
Before:
image
After: 
image
`npm audit fix` did address good number of issues.

dhadka:
Thanks for putting this together!  Should we add a test for large caches either here (`actions/toolkit`) or in `actions/cache`?  Ideally we should be able to repro the error without this fix and then show the error is resolved here.

ashwinsangem:
The error is resolved when with the above fix.
https://github.com/ashwinsangem/test-cache/runs/5675166034?check_suite_focus=true
image

bishal-pdMSFT (approved):

vsvipul (approved):

dhadka (approved):

PR #1036, opened 2022-03-29
Added cacheSize in ReserveCache API request

t-dedah:
Added an optional parameter ie `cacheSize` in Reserve Cache API which will be sent as part of body if the user is an GHES customer. In all the other cases, `cacheSize` will be missing.

### DIFF
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -44,6 +44,16 @@ function checkKey(key: string): void {
 }
 
 /**
+ * isFeatureAvailable to check the presence of Actions cache service
+ *
+ * @returns boolean return true if Actions cache service feature is available, otherwise false
+ */
+
+ export function isFeatureAvailable(): boolean {
+  return !!process.env['ACTIONS_CACHE_URL']
+}
+
+/**
  * Restores cache from keys
  *
  * @param paths a list of file paths to restore from the cache
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -44,16 +44,6 @@ function checkKey(key: string): void {
 }
 
 /**
- * isFeatureAvailable to check the presence of Actions cache service
- *
- * @returns boolean return true if Actions cache service feature is available, otherwise false
- */
-
-export function isFeatureAvailable(): boolean {
-  return !!process.env['ACTIONS_CACHE_URL']
-}
-
-/**
  * Restores cache from keys
  *
  * @param paths a list of file paths to restore from the cache
@@ -152,17 +142,7 @@ export async function saveCache(
   checkKey(key)
 
   const compressionMethod = await utils.getCompressionMethod()
-
-  core.debug('Reserving Cache')
-  const cacheId = await cacheHttpClient.reserveCache(key, paths, {
-    compressionMethod
-  })
-  if (cacheId === -1) {
-    throw new ReserveCacheError(
-      `Unable to reserve cache with key ${key}, another job may be creating this cache.`
-    )
-  }
-  core.debug(`Cache ID: ${cacheId}`)
+  let cacheId = null
 
   const cachePaths = await utils.resolvePaths(paths)
   core.debug('Cache Paths:')
@@ -185,13 +165,31 @@ export async function saveCache(
     const fileSizeLimit = 10 * 1024 * 1024 * 1024 // 10GB per repo limit
     const archiveFileSize = utils.getArchiveFileSizeInBytes(archivePath)
     core.debug(`File Size: ${archiveFileSize}`)
-    if (archiveFileSize > fileSizeLimit) {
+    if (archiveFileSize > fileSizeLimit && !utils.isGhes()) {
       throw new Error(
         `Cache size of ~${Math.round(
           archiveFileSize / (1024 * 1024)
         )} MB (${archiveFileSize} B) is over the 10GB limit, not saving cache.`
       )
     }
+
+    const cacheSize = utils.isGhes() ? archiveFileSize : undefined
+    core.debug('Reserving Cache')
+    cacheId = await cacheHttpClient.reserveCache(key, paths, {
+      compressionMethod,
+      cacheSize
+    })
+    if (cacheId === -1) {
+      throw new ReserveCacheError(
+        `Unable to reserve cache with key ${key}, another job may be creating this cache.`
+      )
+    }
+    if (cacheId === -2) {
+      throw new ReserveCacheError(
+        `Cache size of ~${Math.round(archiveFileSize / (1024 * 1024))} MB (${archiveFileSize} B) is over the data cap limit, not saving cache.`
+      )
+    }
+    core.debug(`Cache ID: ${cacheId}`)
 
     core.debug(`Saving Cache (ID: ${cacheId})`)
     await cacheHttpClient.saveCache(cacheId, archivePath, options)
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -49,7 +49,7 @@ function checkKey(key: string): void {
  * @returns boolean return true if Actions cache service feature is available, otherwise false
  */
 
- export function isFeatureAvailable(): boolean {
+export function isFeatureAvailable(): boolean {
   return !!process.env['ACTIONS_CACHE_URL']
 }
 
--- a/packages/cache/src/internal/cacheHttpClient.ts
+++ b/packages/cache/src/internal/cacheHttpClient.ts
@@ -32,11 +32,7 @@ const versionSalt = '1.0'
 
 function getCacheApiUrl(resource: string): string {
   // Ideally we just use ACTIONS_CACHE_URL
-  const baseUrl: string = (
-    process.env['ACTIONS_CACHE_URL'] ||
-    process.env['ACTIONS_RUNTIME_URL'] ||
-    ''
-  ).replace('pipelines', 'artifactcache')
+  const baseUrl: string = process.env['ACTIONS_CACHE_URL'] || ''
   if (!baseUrl) {
     throw new Error('Cache Service Url not found, unable to restore cache.')
   }
--- a/packages/cache/src/internal/cacheHttpClient.ts
+++ b/packages/cache/src/internal/cacheHttpClient.ts
@@ -31,7 +31,12 @@ import {
 const versionSalt = '1.0'
 
 function getCacheApiUrl(resource: string): string {
-  const baseUrl: string = process.env['ACTIONS_CACHE_URL'] || ''
+  // Ideally we just use ACTIONS_CACHE_URL
+  const baseUrl: string = (
+    process.env['ACTIONS_CACHE_URL'] ||
+    process.env['ACTIONS_RUNTIME_URL'] ||
+    ''
+  ).replace('pipelines', 'artifactcache')
   if (!baseUrl) {
     throw new Error('Cache Service Url not found, unable to restore cache.')
   }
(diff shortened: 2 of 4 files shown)


packages/cache/src/cache.ts
```diff
@@ -185,14 +175,32 @@ export async function saveCache(
...
+      cacheSize
+    })
+    if (cacheId === -1) {
+      throw new ReserveCacheError(
+        `Unable to reserve cache with key ${key}, another job may be creating this cache.`
+      )
+    }
+    if (cacheId === -2) {
+      throw new ReserveCacheError(
+        `Cache size of ~${Math.round(archiveFileSize / (1024 * 1024))} MB (${archiveFileSize} B) is over the data cap limit, not saving cache.`
```
bishal-pdMSFT, line 199: Can we use the API response error message directly. It has the cap mentioned which would be a valuable info in action log.

packages/cache/src/cache.ts
```diff
@@ -185,14 +175,32 @@ export async function saveCache(
...
+    cacheId = await cacheHttpClient.reserveCache(key, paths, {
+      compressionMethod,
+      cacheSize
+    })
+    if (cacheId === -1) {
+      throw new ReserveCacheError(
+        `Unable to reserve cache with key ${key}, another job may be creating this cache.`
+      )
+    }
+    if (cacheId === -2) {
```
bishal-pdMSFT, line 197: As discussed, instead of `-2`, use the status code of `400`

packages/cache/src/cache.ts
```diff
@@ -185,14 +175,32 @@ export async function saveCache(
     const fileSizeLimit = 10 * 1024 * 1024 * 1024 // 10GB per repo limit
     const archiveFileSize = utils.getArchiveFileSizeInBytes(archivePath)
     core.debug(`File Size: ${archiveFileSize}`)
-    if (archiveFileSize > fileSizeLimit) {
+    if (archiveFileSize > fileSizeLimit && !utils.isGhes()) {
```
bishal-pdMSFT, line 178: Add a comment here as to why we are excluding this check for GHES

packages/cache/src/cache.ts
```diff
@@ -185,14 +175,32 @@ export async function saveCache(
...
       throw new Error(
         `Cache size of ~${Math.round(
           archiveFileSize / (1024 * 1024)
         )} MB (${archiveFileSize} B) is over the 10GB limit, not saving cache.`
       )
     }
 
+    const cacheSize = utils.isGhes() ? archiveFileSize : undefined
```
bishal-pdMSFT, line 186: Can we not send size even for non-GHES? There is no harm and server side can decide whether to process it or not (which currently it will not do). But sending every-time just simplifies the logic a bit here.

bishal-pdMSFT (changes requested):
Left some comments.

packages/cache/src/cache.ts
bishal-pdMSFT: Also, this will allow multiple input validations to show appropriate error message.

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -123,3 +123,10 @@ export function assertDefined<T>(name: string, value?: T): T {
 
   return value
 }
+
+export function isGhes(): boolean {
```
bishal-pdMSFT, line 127: Please update cache actions and other setup-* actions to use this utils instead of their own copies. Not a blocking comment but good to take up as a follow up.

packages/cache/src/cache.ts
anuragc617, line 186: can we move this entire logic of calling to reserve cache and figuring out the right error message to a function.

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -149,14 +149,18 @@ export async function reserveCache(
...
+    version,
+    cacheSize: options?.cacheSize
   }
   const response = await retryTypedResponse('reserveCache', async () =>
     httpClient.postJson<ReserveCacheResponse>(
       getCacheApiUrl('caches'),
       reserveCacheRequest
     )
   )
+  if(response?.statusCode === 400){
```
anuragc617, line 161: As discussed return cacheId and and message from here.

packages/cache/src/cache.ts
```diff
@@ -182,17 +172,31 @@ export async function saveCache(
...
+    const cacheSize = archiveFileSize
+    core.debug('Reserving Cache')
+    let reserveCacheResponse = await cacheHttpClient.reserveCache(key, paths, {
+      compressionMethod,
+      cacheSize
+    })
+
+    if(reserveCacheResponse?.statusCode === 400){
+      throw new ReserveCacheError(
+        `Cache size of ~${Math.round(archiveFileSize / (1024 * 1024))} MB (${archiveFileSize} B) is over the data cap limit, not saving cache.`
```
bishal-pdMSFT, line 187: This is wrong and will cause any 400 status code to show this error message to user.

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -143,21 +143,22 @@ export async function reserveCache(
...
   const httpClient = createHttpClient()
   const version = getCacheVersion(paths, options?.compressionMethod)
 
   const reserveCacheRequest: ReserveCacheRequest = {
     key,
-    version
+    version,
+    cacheSize: options?.cacheSize
   }
   const response = await retryTypedResponse('reserveCache', async () =>
```
bishal-pdMSFT, line 155: you can try using [retryHttpClientResponse](https://github.com/actions/toolkit/blob/b4639928698a6bfe1c4bdae4b2bfdad1cb75016d/packages/cache/src/internal/requestUtils.ts#L123)

packages/cache/src/internal/requestUtils.ts
```diff
@@ -111,7 +111,9 @@ export async function retryTypedResponse<T>(
         return {
           statusCode: error.statusCode,
           result: null,
-          headers: {}
+          headers: {},
+          message: error.message,
+          typeKey: error.result.typeKey
```
bishal-pdMSFT, line 116: why do we need `typeKey`. can we not just use error message for logging an error?

bishal-pdMSFT, line 116: Instead of adding these two separately can we add `error` directly. I think the type is [HttpClientError](https://github.com/actions/http-client/blob/5ea1c5292929dba26848ea46f3b4c2e1ea1e30a2/index.ts#L72)

packages/cache/src/cache.ts
```diff
@@ -182,17 +172,41 @@ export async function saveCache(
...
     }
 
-    const fileSizeLimit = 10 * 1024 * 1024 * 1024 // 10GB per repo limit
     const archiveFileSize = utils.getArchiveFileSizeInBytes(archivePath)
     core.debug(`File Size: ${archiveFileSize}`)
-    if (archiveFileSize > fileSizeLimit) {
-      throw new Error(
-        `Cache size of ~${Math.round(
-          archiveFileSize / (1024 * 1024)
-        )} MB (${archiveFileSize} B) is over the 10GB limit, not saving cache.`
```
bishal-pdMSFT, line 192: Don't remove this. We may need it for backward compatiblity. You can put a check so that it only runs for non-GHES.

packages/cache/src/cache.ts
```diff
@@ -182,17 +172,41 @@ export async function saveCache(
...
+      paths,
+      {
+        compressionMethod,
+        cacheSize
+      }
+    )
+
+    if (
+      reserveCacheResponse?.statusCode === 400 &&
+      reserveCacheResponse?.typeKey === 'InvalidReserveCacheRequestException'
```
bishal-pdMSFT, line 191: Why handle this exception explicitly? IMO just check for any non null `error` and if so, throw a ReserveCacheError and use same message from `error.message`

packages/cache/src/cache.ts
```diff
@@ -182,17 +172,41 @@ export async function saveCache(
...
+    const reserveCacheResponse = await cacheHttpClient.reserveCache(
+      key,
+      paths,
+      {
+        compressionMethod,
+        cacheSize
+      }
+    )
+
+    if (
```
bishal-pdMSFT, line 189: This if block can be moved as else to L204.

packages/cache/src/cache.ts
t-dedah: Done

packages/cache/src/cache.ts
t-dedah: Done

packages/cache/src/cache.ts
t-dedah: Removed this code altogether

packages/cache/src/cache.ts
t-dedah: Yes, now we are sending in both ghes and non ghes case

packages/cache/src/internal/cacheUtils.ts
t-dedah: Removed this method.

packages/cache/src/internal/cacheHttpClient.ts
t-dedah: Done

packages/cache/src/internal/requestUtils.ts
t-dedah: `typeKey` is the name of the exception thrown. This will help us narrow down error and take appropriate steps like 400 with multiple possible values of `typeKey` can be handled differently.

packages/cache/src/cache.ts
t-dedah: > Can we not send size even for non-GHES? There is no harm and server side can decide whether to process it or not (which currently it will not do). But sending every-time just simplifies the logic a bit here.

@bishal-pdMSFT 
If we wish to send `cacheSize` for all the cases wont it be much cleaner to not check anything here?
We have updated the tests accordingly so what else will require backward compatibility as the behaviour for user is still same?

bishal-pdMSFT: Backward compatibility in the sense that the action/toolkit should still do the checking even it someone hits an older service which does not have the service side checking. There are two cases in which such need arises:
1. if the service side change in rolled back for some reason (lets say a critical bug on the service side). Since the client is shipped we can't hide/un-ship it, such clients will not have this check happening. Also applies if service side change in only rolled out to few rings only.
2. If the action is run against an older GHES server (<3.5) then that server will not have server side check. This is actually not a problem for us as cache service does not exist in older GHES versions but I guess the principal for compatibility can still apply.

TBH these are not big reasons in our case but I guess its good to have backward compatibility as a principle.

packages/cache/src/internal/requestUtils.ts
bishal-pdMSFT: Do we have a use case of handling different exception types differently? I can't think of such use case currently.

packages/cache/src/cache.ts
t-dedah: Brought back this code and added a comment.

packages/cache/src/internal/cacheHttpClient.ts
t-dedah: Extended `ITypedResponse` to include `message` for error message. This will allow us to keep using the same method with advanced Response Type. This also gives us additional functionality of having error message as `message` in all the other responses for other API calls.

packages/cache/src/cache.ts
t-dedah: Using error.message is it exists else using a default message with similar information.

packages/cache/src/cache.ts
t-dedah: Done

packages/cache/src/internal/requestUtils.ts
t-dedah: Added `error: HttpClientError` instead of these two fields.

PR #1044, opened 2022-03-31
Added cacheSize in ReserveCache API request

t-dedah:
Added an optional parameter ie cacheSize in Reserve Cache API which will be sent as part of body if the user is an GHES customer. In all the other cases, cacheSize will be missing.

[ADR](https://github.com/github/c2c-actions/blob/c5e9d502a0fb5c1f806366347c8ad10f695e8816/docs/teams/platform/adrs/artifact-cache/0704-set-artifact-cache-size-limit.md)

The original PR got merged by mistake and reverted later on. Link to original PR https://github.com/actions/toolkit/pull/1036


packages/cache/src/cache.ts
```diff
@@ -181,18 +171,45 @@ export async function saveCache(
...
+          )} MB (${archiveFileSize} B) is over the data cap limit, not saving cache.`
+      )
+    } else {
+      throw new ReserveCacheError(
+        `Unable to reserve cache with key ${key}, another job may be creating this cache.`
+      )
+    }
+
+    core.debug(`Cache ID: ${cacheId}`)
     core.debug(`Saving Cache (ID: ${cacheId})`)
```
bishal-pdMSFT, line 213: Do we need both?

packages/cache/src/cache.ts
```diff
@@ -181,18 +171,45 @@ export async function saveCache(
...
+    // For GHES, this check will take place in ReserveCache API with enterprise file size limit
+    if (archiveFileSize > fileSizeLimit && !utils.isGhes()) {
       throw new Error(
         `Cache size of ~${Math.round(
           archiveFileSize / (1024 * 1024)
         )} MB (${archiveFileSize} B) is over the 10GB limit, not saving cache.`
       )
     }
 
+    const cacheSize = archiveFileSize
```
bishal-pdMSFT, line 186: Is this needed? Just keep using `archiveFileSize`

packages/cache/src/cache.ts
```diff
@@ -181,18 +171,45 @@ export async function saveCache(
     if (core.isDebug()) {
       await listTar(archivePath, compressionMethod)
     }
-
     const fileSizeLimit = 10 * 1024 * 1024 * 1024 // 10GB per repo limit
     const archiveFileSize = utils.getArchiveFileSizeInBytes(archivePath)
     core.debug(`File Size: ${archiveFileSize}`)
-    if (archiveFileSize > fileSizeLimit) {
+    // For GHES, this check will take place in ReserveCache API with enterprise file size limit
```
bishal-pdMSFT, line 177: leave an empty line before a comment

packages/cache/src/cache.ts
```diff
@@ -181,18 +171,45 @@ export async function saveCache(
...
+    } else if (reserveCacheResponse?.statusCode === 400) {
+      throw new ReserveCacheError(
+        reserveCacheResponse?.error?.message ??
+          `Cache size of ~${Math.round(
+            archiveFileSize / (1024 * 1024)
+          )} MB (${archiveFileSize} B) is over the data cap limit, not saving cache.`
+      )
+    } else {
+      throw new ReserveCacheError(
+        `Unable to reserve cache with key ${key}, another job may be creating this cache.`
```
bishal-pdMSFT, line 208: Now that we have error message, it can be appended here to give more insights to user.,

```suggestion
        `Unable to reserve cache with key ${key}, another job may be creating this cache. More details: ${reserveCacheResponse?.error?.message}`
```

packages/cache/src/internal/requestUtils.ts
```diff
@@ -111,7 +109,8 @@ export async function retryTypedResponse<T>(
         return {
           statusCode: error.statusCode,
           result: null,
-          headers: {}
+          headers: {},
+          error: error
```
bishal-pdMSFT, line 113: There are tests for [requestUtils](https://github.com/actions/toolkit/blob/e779919848e5a3ee655ab825e53f8ac979390a5c/packages/cache/__tests__/requestUtils.test.ts). Update/add tests to include testing `error` property.

packages/cache/__tests__/saveCache.test.ts
```diff
@@ -60,7 +66,95 @@ test('save with large cache outputs should fail', async () => {
...
+  const cacheSize = 11 * 1024 * 1024 * 1024 //~11GB, over the 10GB limit
+  jest
+    .spyOn(cacheUtils, 'getArchiveFileSizeInBytes')
+    .mockReturnValueOnce(cacheSize)
+  const compression = CompressionMethod.Gzip
+  const getCompressionMock = jest
+    .spyOn(cacheUtils, 'getCompressionMethod')
+    .mockReturnValueOnce(Promise.resolve(compression))
+
+  jest.spyOn(cacheUtils, 'isGhes').mockReturnValueOnce(true)
```
bishal-pdMSFT, line 94: I believe this block may be common code among more than 1 tests. If so it can be refactored into a separate method to keep each test code less verbose.

packages/cache/src/internal/contracts.d.ts
```diff
@@ -1,4 +1,10 @@
 import {CompressionMethod} from './constants'
+import {ITypedResponse} from '@actions/http-client/interfaces'
+import {HttpClientError} from '@actions/http-client'
+
+export interface ITypedResponseWithErrorMessage<T> extends ITypedResponse<T> {
```
bishal-pdMSFT, line 5: Rename to `ITypedResponseWithError`

packages/cache/src/cache.ts
```diff
@@ -181,18 +171,45 @@ export async function saveCache(
     if (core.isDebug()) {
       await listTar(archivePath, compressionMethod)
     }
-
     const fileSizeLimit = 10 * 1024 * 1024 * 1024 // 10GB per repo limit
     const archiveFileSize = utils.getArchiveFileSizeInBytes(archivePath)
     core.debug(`File Size: ${archiveFileSize}`)
-    if (archiveFileSize > fileSizeLimit) {
+    // For GHES, this check will take place in ReserveCache API with enterprise file size limit
+    if (archiveFileSize > fileSizeLimit && !utils.isGhes()) {
```
thboop, line 178: should we check the size on GHES and throw an appropriate error here? May be compromising the GHES experience here if they don't change the cache size. What happens if we try to upload a file that is too big? Does it try to upload and fail.

packages/cache/src/cache.ts
```diff
@@ -181,18 +171,45 @@ export async function saveCache(
...
+        cacheSize
+      }
+    )
+
+    if (reserveCacheResponse?.result?.cacheId) {
+      cacheId = reserveCacheResponse?.result?.cacheId
+    } else if (reserveCacheResponse?.statusCode === 400) {
+      throw new ReserveCacheError(
+        reserveCacheResponse?.error?.message ??
+          `Cache size of ~${Math.round(
```
thboop, line 202: Should the server return a message and we show that here instead? Just assuming every 400 is a Bad Request may not scale well if we return 400's for other issues.

packages/cache/package-lock.json
```diff
@@ -1,3 +1,5 @@
+
+
```
thboop, line 2: I think we should remove these lines

packages/cache/__tests__/saveCache.test.ts
```diff
@@ -60,7 +66,95 @@ test('save with large cache outputs should fail', async () => {
...
+  expect(createTarMock).toHaveBeenCalledTimes(1)
+  expect(createTarMock).toHaveBeenCalledWith(
+    archiveFolder,
+    cachePaths,
+    compression
+  )
+  expect(getCompressionMock).toHaveBeenCalledTimes(1)
+})
+
+test('save with large cache outputs should fail in GHES with error message', async () => {
```
thboop, line 78: 👍  i like the tests!

packages/cache/src/internal/requestUtils.ts
thboop, line 113: we should address the linter issues/test failures before we merge.

packages/cache/src/cache.ts
t-dedah: No, removed the first one.

packages/cache/src/cache.ts
t-dedah: Done

packages/cache/src/cache.ts
t-dedah: Done

packages/cache/src/internal/contracts.d.ts
t-dedah: Done

packages/cache/package-lock.json
t-dedah: Done

packages/cache/__tests__/saveCache.test.ts
t-dedah: 😄

packages/cache/src/internal/requestUtils.ts
t-dedah: Done

packages/cache/src/cache.ts
t-dedah: Wont `reserveCacheResponse?.error?.message ??` handle that? We are first checking if message is present in the response or not. If not only then we fallback to the default value.

packages/cache/src/cache.ts
t-dedah: As we dont have the limit for GHES repos here, we moved the check to actions-dotnet. If the file is too big we will thrown an error which is being handled as 400 here.

[ADR](https://github.com/github/c2c-actions/blob/c5e9d502a0fb5c1f806366347c8ad10f695e8816/docs/teams/platform/adrs/artifact-cache/0704-set-artifact-cache-size-limit.md)
[PR link to this change. ](https://github.com/github/actions-dotnet/pull/11557)

bishal-pdMSFT (approved):

PR #1060, opened 2022-04-25
Diff @actions/http-client changes

brcrista:
**Note: the base branch for this PR is `brcrista/http-client-base`, not `main`.**  This is a branch I created just from copying over the files from https://github.com/actions/http-client without any changes.

The purpose of this PR is to highlight the changes I had to make to get `@actions/http-client` building along with toolkit.  See https://github.com/actions/toolkit/pull/1062 for the actual PR to the mainline.

The biggest change is having to update some exported types in `http-client` to get it to build with `strict`.  For example, null checking is now enforced.  While there are more involved ways to clean this up (for example, to eliminate the possibility of returning nulls), I tried just to describe the current state of things without making any changes to behavior.


packages/artifact/__tests__/retry.test.ts
```diff
@@ -3,7 +3,7 @@ import * as net from 'net'
 import * as core from '@actions/core'
 import * as configVariables from '../src/internal/config-variables'
 import {retry} from '../src/internal/requestUtils'
-import {IHttpClientResponse} from '@actions/http-client/interfaces'
+import {IHttpClientResponse} from '@actions/http-client/lib/interfaces'
```
brcrista, line 6: Having to put `lib` in the module reference seems to be a `lerna` thing.  I tested what happens if you install the actual package through `npm` and it doesn't change.

Other packages in this repo reference each other's submodules through `lib/`.

packages/http-client/__tests__/basics.test.ts
```diff
@@ -39,7 +39,7 @@ describe('basics', () => {
   //     "url": "https://httpbin.org/get"
   //  }
 
-  it('does basic http get request', async done => {
+  it('does basic http get request', async () => {
```
brcrista, line 42: Got an error about passing `done` to an `async` callback.  The Jest version in this project is a couple major versions newer than what `http-client` had, so I figured that was a breaking change that went in.

packages/core/RELEASES.md
```diff
@@ -1,5 +1,8 @@
 # @actions/core Releases
 
+### 1.7.0
```
thboop, line 3: might want to rebase to smooth out this diff

packages/core/package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/core",
-  "version": "1.6.0",
```
thboop, line 3: should be 1.7.1

packages/artifact/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/artifact",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "preview": true,
```
thboop, line 4: We should update all the package-lock.json as well, alongside release notes

packages/http-client/package.json
```diff
@@ -1,37 +1,45 @@
...
-    "test": "jest",
-    "format": "prettier --write *.ts && prettier --write **/*.ts",
-    "format-check": "prettier --check *.ts && prettier --check **/*.ts",
-    "audit-check": "npm audit --audit-level=moderate"
+  "keywords": [
+    "github",
+    "actions",
+    "http"
+  ],
+  "homepage": "https://github.com/actions/toolkit/tree/main/packages/github",
```
thboop, line 10: nit fix this homepage to the the link to the http-client package

packages/http-client/src/auth.ts
```diff
@@ -23,9 +23,9 @@ export class BasicCredentialHandler implements ifm.IRequestHandler {
   handleAuthentication(
     httpClient: ifm.IHttpClient,
     requestInfo: ifm.IRequestInfo,
-    objs
+    objs: any
   ): Promise<ifm.IHttpClientResponse> {
-    return null
+    throw new Error("not implemented")
```
thboop, line 28: What do these changes do?

packages/http-client/src/proxy.ts
```diff
@@ -1,23 +1,23 @@
...
+    return undefined
   }
 
-  let proxyVar: string
-  if (usingSsl) {
-    proxyVar = process.env['https_proxy'] || process.env['HTTPS_PROXY']
-  } else {
-    proxyVar = process.env['http_proxy'] || process.env['HTTP_PROXY']
-  }
+  const proxyVar = (() => {
```
thboop, line 8: Why turn this into a function, shouldn't we set the string? No need to invoke this multiple times

packages/http-client/package.json
```diff
@@ -1,37 +1,45 @@
 {
   "name": "@actions/http-client",
-  "version": "1.0.11",
+  "version": "2.0.0",
```
thboop, line 3: can we add/update the releases.md with information on why we have published a new major version, update the readme, and let users know why they should use it?

packages/artifact/package.json
brcrista: 👍  I think CI is failing because I need to update the `package-lock.json`s.  I'm struggling a bit to figure out how to get `npm` to update them with `lerna` in the mix

packages/core/RELEASES.md
brcrista: I tried that and it made this change go away, but for some reason that made the files in `http-client/src` all show up as "added" instead of "modified" (????).  Since those are the files that really matter here, I switched it back to the pre-rebase commit.

packages/http-client/src/auth.ts
brcrista: Returning `null` fails type checking with `strict: true`.  I could have just added `| null` to the return type, but that doesn't make sense for when the method is actually implemented.  So better to throw an exception for when it's not implemented.  Callers can use `canHandleAuthentication()` to make sure the method is implemented.

brcrista:
This is out-of-date now.  I'll open a new one

michelledeprosse21:
Can you please show me how to retrieve my Ethereum off of Etherscan?

Micdeprosse21 
 
  On Wed, Apr 27, 2022 at 6:52 PM, Brian ***@***.***> wrote:   


Closed #1060.

PR #1061, opened 2022-04-25
Move @actions/http-client into the toolkit

brcrista:
💡 See https://github.com/actions/toolkit/pull/1060 for a better diff!

https://github.com/actions/toolkit contains a variety of packages used for building actions.  https://github.com/actions/http-client is one such package, but lives outside of the toolkit.  Moving it inside of the toolkit would reduce the number of repos we have to fan out maintenance tasks for (such as github/c2c-actions-service#2937).

I checked with @bryanmacfarlane on the historical decision here.  Apparently it was just inertia from before we released the toolkit as multiple packages.

The benefits here are:
- Have one fewer repo to keep track of
- Signal that this is an HTTP client meant for building actions, not for general use.

`@actions/http-client` will continue to be released as its own package.

### DIFF
--- a/.github/workflows/test-http-client.yml
+++ b/.github/workflows/test-http-client.yml
@@ -1,0 +1,51 @@
+name: http-tests
+on: 
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    paths-ignore:
+      - '**.md'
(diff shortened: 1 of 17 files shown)


brcrista:
Wrong branch 🤦‍♂️

PR #1062, opened 2022-04-25
Move @actions/http-client into the toolkit

brcrista:
💡 See https://github.com/actions/toolkit/pull/1060 for a better diff!

https://github.com/actions/toolkit contains a variety of packages used for building actions.  https://github.com/actions/http-client is one such package, but lives outside of the toolkit.  Moving it inside of the toolkit would reduce the number of repos we have to fan out maintenance tasks for (such as github/c2c-actions-service#2937).

I checked with @bryanmacfarlane on the historical decision here.  Apparently it was just inertia from before we released the toolkit as multiple packages.

The benefits here are:
- Have one fewer repo to keep track of
- Signal that this is an HTTP client meant for building actions, not for general use.

`@actions/http-client` will continue to be released as its own package.


scripts/create-package
```diff
@@ -9,5 +9,5 @@ if [[ -z "$name" ]]; then
   exit 1
 fi
 
-lerna create @actions/$name
-cp packages/toolkit/tsconfig.json packages/$name/tsconfig.json
```
brcrista, line 13: This file doesn't exist

thboop (approved):
lgtm

karlhorky:
@thboop @brcrista seems like [moving `tunnel` to `devDependencies`](https://github.com/actions/toolkit/pull/1064/files#diff-8296ac8ec54f53d18350a5a98a041ba1af0307c61d1406d31350740b383d274dL36) caused failures (eg. when upgrading to `@actions/core@1.8.1`):

https://github.com/actions/toolkit/issues/1083

brcrista:
Thanks @karlhorky, should be fixed now https://github.com/actions/toolkit/pull/1085

PR #1063, opened 2022-04-26
Add checksum validation on artifact upload

robherley:
Relevant Issues:
- https://github.com/github/c2c-actions-checks/issues/301
- https://github.com/github/c2c-actions-checks/issues/391

**tl;dr** adds base64 encodings of the CRC64 and MD5 to the headers of a chunk upload:

```
x-actions-results-crc64
x-actions-results-md5
```

Recently we've been running into issues when artifacts are getting corrupted during the upload process. This is extremely rare (seen < 1% during my tests) but happens nonetheless. There still needs to be more investigation as to _why_ these are being corrupted, but we've narrowed the problem areas to be between the upload and file container service handler after isolated testing with md5 checksum headers. 

This PR adds a CRC64 and MD5 checksum to the header. Since NodeJS's crypto library (openssl bindings) doesn't have CRC64, I added a simple implementation based on Go's [`hash/crc64`](https://cs.opensource.google/go/go/+/master:src/hash/crc64/crc64.go) pkg. Also this is tailored to the CRC polynomial [used by azure storage](https://github.com/Azure/azure-storage-net/blob/b0d47bfd4c1c0dcbe60eaf373e2977cd7405aa5f/Lib/Common/Core/Util/Crc64.cs#L25). We're already using this same polynomial places deeper in the stack.

I ran some [benchmarks](https://gist.github.com/robherley/338d90d20ad23fe235a7de69f57c2a75) and this CRC64 implementation is a bit faster than sha256/md5:

```
crc64 x 145,627 ops/sec ±8.33% (79 runs sampled)
md5 x 76,392 ops/sec ±12.44% (69 runs sampled)
sha256 x 65,982 ops/sec ±14.55% (62 runs sampled)
Fastest is crc64
```


packages/artifact/src/internal/upload-http-client.ts
```diff
@@ -406,14 +407,18 @@ export class UploadHttpClient {
     isGzip: boolean,
     totalFileSize: number
   ): Promise<boolean> {
+    // open a new stream and read it to compute the digest
+    const digest = await digestForStream(openStream())
```
robherley, line 411: note: we'll have to read the data chunk twice to calculate the CRC digest. Not sure if there is a more clever solution where we can calculate during chunking (or if it's worth it 🤷)

cc @konradpabjan

packages/artifact/src/internal/utils.ts
```diff
@@ -202,6 +204,9 @@ export function getUploadHeaders(
   if (contentRange) {
     requestOptions['Content-Range'] = contentRange
   }
+  if (digest) {
+    requestOptions['X-GH-Actions-CRC64'] = digest
```
robherley, line 208: I came up with `X-GH-Actions-CRC64` for the header, willing to change if there's a better alternative 👍

packages/artifact/src/internal/upload-http-client.ts
konradpabjan: My initial hunch is this shouldn't be a concern. The majority of the time as part of artifact upload is due to networking so a small overhead to compute the digest should be fine.

packages/artifact/src/internal/utils.ts
konradpabjan: For normal parts of the site I see stuff like this:

image

so how about something like `x-gh-actions-crc64` for consistency with lowercase? I'm also leaning towards `x-actions-result-crc64` (with our future architecture this might make sense)

packages/artifact/src/internal/upload-http-client.ts
```diff
@@ -406,14 +407,18 @@ export class UploadHttpClient {
     isGzip: boolean,
     totalFileSize: number
   ): Promise<boolean> {
+    // open a new stream and read it to compute the digest
```
konradpabjan, line 410: 10 lines up you can add another `@param` with a comment

packages/artifact/src/internal/crc64.ts
```diff
@@ -0,0 +1,300 @@
...
+ * validate that there was no corruption during transmission. The implementation here is based on Go's hash/crc64 pkg,
+ * but without the slicing-by-8 optimization: https://cs.opensource.google/go/go/+/master:src/hash/crc64/crc64.go
+ *
+ * This implementation uses a pregenerated table based on 0x9A6C9329AC4BC9B5 as the polynomial, the same polynomial that
+ * is used for Azure Storage: https://github.com/Azure/azure-storage-net
+ */
+
+// when transpile target is >= ES2020 (after dropping node 12) these can be changed to bigint literals - ts(2737)
+const AZURE_TABLE = [
+  BigInt('0x0000000000000000'),
```
konradpabjan, line 14: I found the original file, pretty cool: https://github.com/Azure/azure-storage-net/blob/cbe605f9faa01bfc3003d75fc5a16b2eaccfe102/Lib/Common/Core/Util/Crc64.cs#L29

How about linking to that directly in the comment?

Also i would change `AZURE_TABLE` to a different name, though I can't think of anything off the top of my head

packages/artifact/src/internal/utils.ts
robherley: I like `x-actions-result-crc64` 👍 Also HTTP Headers are case insensitive so we shouldn't need to worry about that

packages/artifact/src/internal/upload-http-client.ts
robherley: it's not a new param though 🤔

konradpabjan: Whoops, digest in the utils file

packages/artifact/src/internal/crc64.ts
robherley: changed to `PREGEN_POLY_TABLE` in https://github.com/actions/toolkit/pull/1063/commits/b68735e06098244d8f6a11cfac4f0000d6807e4c

konradpabjan (approved):

wesdgray:
I ran into this problem with a service that handled large file uploads and the root cause was that our continuous deployment was killing our containers after a shorter than expected timeout. In our case, during deployment, hosts are taken out of the pool and in-flight requests are allowed to finish but it was not known to us at the time that a timeout was even in effect. 

Also if you're wondering how I even stumbled upon this: https://youtu.be/9qljpi5jiMQ?t=722

PR #1064, opened 2022-04-28
Diff @actions/http-client changes

brcrista:
**Note: the base branch for this PR is `brcrista/http-client-diff`, not `main`.**  This is a branch I created just from copying over the files from https://github.com/actions/http-client without any changes.

The purpose of this PR is to highlight the changes I had to make to get `@actions/http-client` building along with toolkit.  See https://github.com/actions/toolkit/pull/1062 for the actual PR to the mainline.

The biggest change is having to update some exported types in `http-client` to get it to build with `strict`.  For example, null checking is now enforced.  While there are more involved ways to clean this up (for example, to eliminate the possibility of returning nulls), I tried just to describe the current state of things without making any changes to behavior.

There are also many minor changes to satisfy linting rules.


packages/http-client/src/interfaces.ts
```diff
@@ -1,83 +1,84 @@
-import http = require('http')
+import * as http from 'http'
+import * as https from 'https'
 
-export interface IHeaders {
-  [key: string]: any
+export interface Headers {
+  [header: string]: string
```
brcrista, line 5: I looked at existing usage in `http-client` and other toolkit packages and we always use string values for headers.  Node's `http` module tends to use `string | string[] | undefined` for header values, but that would have required more code changes to handle where headers are consumed in `http-client`.

packages/http-client/src/auth.ts
```diff
@@ -38,26 +38,25 @@ export class BearerCredentialHandler implements ifm.IRequestHandler {
...
+  prepareRequest(options: http.RequestOptions): void {
+    if (!options.headers) {
+      throw Error('The request has no headers')
+    }
+    options.headers['Authorization'] = `Bearer ${this.token}`
   }
 
   // This handler cannot handle 401
-  canHandleAuthentication(response: ifm.IHttpClientResponse): boolean {
+  canHandleAuthentication(): boolean {
```
brcrista, line 49: The linter errors on unused parameters.  TypeScript still considers this as implementing the interface method.

packages/http-client/src/interfaces.ts
```diff
@@ -1,83 +1,76 @@
-import http = require('http')
+import * as http from 'http'
+import * as https from 'https'
+import {HttpClientResponse} from './index'
 
-export interface IHeaders {
-  [key: string]: any
-}
```
brcrista, line 5: Deleting this in favor of the standard `http.OutgoingHttpHeaders`.

PR #1066, opened 2022-04-29
Chore: update dev dependencies and workflows to node16

thboop:
Node 12 is EOL soon, so lets go ahead and move to node 16

### DIFF
--- a/.github/workflows/artifact-tests.yml
+++ b/.github/workflows/artifact-tests.yml
@@ -24,10 +24,10 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
-    - name: Set Node.js 12.x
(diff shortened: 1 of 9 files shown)


JamesMGreene (commented):
If you don't want to need to continually update so many spots for this, you could consider using the newer `node-version-file` parameter of `actions/setup-node` combined with a `.node-version` file. 🙃 

https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file

:warning: In the Node version manager world, the only guaranteed format supported for a `.node-version` file is the entire 3-part SemVer version, e.g. `16.2.0`, but the `actions/setup-node` code _looks like_ it would still allow you to put a range like `16.x` in it. 🤷🏻‍♂️

PR #1082, opened 2022-05-10
Update other packages to use http-client v2

brcrista:
We moved `@actions/http-client` to be part of the toolkit in https://github.com/actions/toolkit/pull/1062.  We also made some breaking changes to exported types and released v2.

The biggest change in terms of lines of code affected was to get rid of the `I-` prefix for interfaces since TypeScript doesn't follow this convention.

I bumped the patch version of all packages except for `tool-cache`, where I bumped the major version.  The rationale is explained in the release notes for that package.


packages/cache/__tests__/saveCache.test.ts
```diff
@@ -5,7 +5,7 @@ import * as cacheHttpClient from '../src/internal/cacheHttpClient'
 import * as cacheUtils from '../src/internal/cacheUtils'
 import {CacheFilename, CompressionMethod} from '../src/internal/constants'
 import * as tar from '../src/internal/tar'
-import {ITypedResponse} from '@actions/http-client/interfaces'
+import {TypedResponse} from '@actions/http-client/lib/interfaces'
```
brcrista, line 8: Having to put `/lib/` here is a Lerna thing now that we're using the same version of the package that's in the repo.

packages/tool-cache/RELEASES.md
```diff
@@ -1,5 +1,10 @@
 # @actions/tool-cache Releases
 
+### 2.0.0
+- Update to v2.0.0 of `@actions/http-client`
+- The type of the `headers` parameter in the exported function `downloadTool` has been narrowed from `{ [header: string]: any }` to `{ [header: string]: number | string | string[] | undefined; }` (that is, `http.OutgoingHttpHeaders`).
+    This is strictly a compile-time change for TypeScript consumers. Previous attempts to use a header value of a type other than those now accepted would have resulted in an error at run time.
```
brcrista, line 6: Note the major version bump.  This could be considered a bug fix, but I went with bumping the major version to be conservative.

luketomlinson (approved):
:shipit:

Issue #1083, opened 2022-05-12
Error: Cannot find module 'tunnel' in `@actions/github@5.0.2`


akphi:
**Describe the bug**
In the latest version of `@actions/github 5.0.2`, we got the following error in our pipeline

```
node:internal/modules/cjs/loader:933
  const err = new Error(message);
              ^

Error: Cannot find module 'tunnel'
Require stack:
- /home/runner/work/legend-studio/legend-studio/node_modules/@actions/http-client/lib/index.js
- /home/runner/work/legend-studio/legend-studio/node_modules/@actions/core/lib/oidc-utils.js
- /home/runner/work/legend-studio/legend-studio/node_modules/@actions/core/lib/core.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
...
```

**To Reproduce**

This is our repo
https://github.com/finos/legend-studio/runs/6414726076?check_suite_focus=true

**Expected behavior**

I think this happens because we use `Yarn` to install dependencies and we got a problem with `Yarn` not installing `tunnel` because it is a `devDependency` of `http-client`, but in fact, `tunnel` should be a runtime dependency of `http-client`.

I believe the workaround from our end is not so complicated - just manually installing `tunnel`, but could we make this a `dependency` instead?

Thank you so much!

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

arminrosu:
>  not installing tunnel because it is a devDependency of http-client, but in fact, tunnel should be a runtime dependency of http-client.

> I believe the workaround from our end is not so complicated - just manually installing tunnel, but could we make this a dependency instead?

Looks like when `actions/http-client` was merged into `actions/toolkit` in #1062 the `tunnel` [dependency](https://github.com/actions/http-client/blob/602cbafdf5b9de6e5a68995f6d8ccf4f318c5fde/package.json#L36) was changed to a [devDependency](https://github.com/actions/toolkit/pull/1062/files#diff-8296ac8ec54f53d18350a5a98a041ba1af0307c61d1406d31350740b383d274dR44). That's when `@actions/http-client@2.0.0` and `@actions/core@1.8.1` were released. Likely an oversight.

karlhorky:
Seems like @gfx opened https://github.com/actions/toolkit/pull/1084 to fix this, watching over there now

PR #1084, opened 2022-05-13
Add a missing dependency "tunnel" to @actions/http-client

gfx:
`tunnel` is not a dev dependency, but a runtime dependency.

Fix https://github.com/actions/toolkit/issues/1083

### DIFF
--- a/packages/http-client/package.json
+++ b/packages/http-client/package.json
@@ -38,9 +38,11 @@
   "bugs": {
     "url": "https://github.com/actions/toolkit/issues"
   },
+  "dependencies" {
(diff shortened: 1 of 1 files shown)


hom-bahrani (approved):

packages/http-client/package.json
```diff
@@ -38,9 +38,11 @@
   "bugs": {
     "url": "https://github.com/actions/toolkit/issues"
   },
+  "dependencies" {
+    "tunnel": "0.0.6"
```
akphi, line 42: I suppose you need to update the `package-lock.json` file as well

PR #1108, opened 2022-06-01
Adding fix for tar exiting with code 1 issue

kotewar:
Fix for https://github.com/actions/cache/issues/689

The issue is mainly happening when `~/` (i.e. the home directory) is used in the path for caching. Internally we use a sub-directory of `home` directory as our temporary storage for our tar to get created. So if our temporary folder is a sub-directory of path that needs to be cached, we fall into an issue where the tar cannot tar itself and finds a change in its contents while it is trying to tar, which leads to the program throwing the following error (as mentioned in the issue) -

```
Post job cleanup.
/usr/bin/tar --posix --use-compress-program zstd -T0 -cf cache.tzst -P -C /home/runner/work/xxx/xxx--files-from manifest.txt
/usr/bin/tar: ../../../work/_temp/6483fbb9-3579-4fb0-a582-3316b336d5a9/cache.tzst: file changed as we read it
Warning: Tar failed with error: The process '/usr/bin/tar' failed with exit code 1
```

To fix this, we've added an explicit exclude for the tar file in the compression command.

Also some of our package-lock.json files are outdated, have committed them as well.


kotewar:
Pending items - 
- Test cases fixes
- Syncing up package.json and package-lock.json

kotewar:
The test run for verifying the same is added here - https://github.com/kotewar/learning-actions/actions/runs/2434060218

vsvipul (changes requested):
Please remove non cache related package-lock.json changes.

anuragc617:
> The test run for verifying the same is present here - https://github.com/kotewar/learning-actions/actions/runs/2434060218

I tried to run the same workflow for cache@v3, and it is still [passing](https://github.com/anuragc617/acr-build-helloworld-node/actions/runs/2446030768/workflow) for me. Are you sure that you have tested the correct scenario?

kotewar:
@anuragc617 Updated the URL with correct test link.  Please refer - https://github.com/kotewar/learning-actions/runs/6751011211?check_suite_focus=true

anuragc617:
Yes, @kotewar my bad restore has happened for ubuntu-latest.

vsvipul (approved):

kotewar:
@anuragc617 are you good with the changes?

anuragc617 (approved):

PR #1112, opened 2022-06-09
Add debug info for check os version id during manifest findMatch

dsame:
Currently the setup-python action fails silently if run the linux distribution other than Ubuntu.
The PR adds more debug to detect the situation the client uses non-ubuntu self-hosted aggent

### DIFF
--- a/packages/tool-cache/src/manifest.ts
+++ b/packages/tool-cache/src/manifest.ts
@@ -92,6 +92,11 @@ export async function _findMatch(
             chk = true
           } else {
             chk = semver.satisfies(osVersion, item.platform_version)
+            if (!chk) {
+              debug(
+                `os version: "${osVersion}" does not match version python is built for: "${item.platform_version}"`
+              )
+            }
           }
         }
 


packages/tool-cache/src/manifest.ts
```diff
@@ -92,6 +92,11 @@ export async function _findMatch(
             chk = true
           } else {
             chk = semver.satisfies(osVersion, item.platform_version)
+            if (!chk) {
+              debug(
+                `os version: "${osVersion}" does not match version python is built for: "${item.platform_version}"`
```
marko-zivic-93, line 97: You can put capital 'OS' instead of lowercase 'os'

marko-zivic-93 (approved):

packages/tool-cache/src/manifest.ts
dsame: Done

vsafonkin (approved):

IvanZosimov (approved):

marko-zivic-93 (approved):

panticmilos (approved):

vsafonkin (approved):

marko-zivic-93 (approved):

PR #1114, opened 2022-06-12
Fix typo in AnnotationProperties doc

r-antonio:
Just that, I was using `core` and checking the docs when I noticed the annotation typo


Pablo1q (approved):
Ok

PR #1118, opened 2022-06-16
[Artifacts] Add more extensions to the gzip compression exception list

hectorm:
Extended the gzip compression exception list with the addition of the extensions recognized by [GNU Tar](https://git.savannah.gnu.org/cgit/tar.git/tree/src/suffix.c?h=release_1_34#n31).

### DIFF
--- a/packages/artifact/src/internal/upload-gzip.ts
+++ b/packages/artifact/src/internal/upload-gzip.ts
@@ -26,6 +26,7 @@ const gzipExemptFileExtensions = [
   '.xz', // XZ
   '.txz', // XZ
   '.zst', // ZSTD
+  '.zstd', // ZSTD
   '.tzst', // ZSTD
   '.zip', // ZIP
   '.7z' // 7ZIP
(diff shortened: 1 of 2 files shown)


packages/artifact/src/internal/upload-gzip.ts
```diff
@@ -9,15 +9,26 @@ const stat = promisify(fs.stat)
  * If any of these types of files are encountered then on-disk gzip creation will be skipped and the original file will be uploaded as-is
  */
 const gzipExemptFileExtensions = [
-  '.gzip',
-  '.zip',
-  '.tar.lz',
-  '.tar.gz',
-  '.tar.bz2',
-  '.tar.zst',
-  '.tar.zstd',
```
konradpabjan, line 18: Thanks for the great PR @hectorm! (really appreciate the tests) 🙇‍♂️ 

Requesting just a small change, could you add `.zstd`, currently it would be removed. Recently it was added as part of another PR https://github.com/actions/toolkit/pull/1184

konradpabjan (approved):

dionrhys:
Hey folks, would it be possible to cut a new `@actions/artifact` release that includes this and other PRs (#1184 #1190 #1125 #1309 #1241) since Jan? I can open an issue for this if you prefer.

I'm using [`upload-artifact`](https://github.com/actions/upload-artifact/tree/main) right now with `.tzst` files, but it's spending time trying to re-compress the `.tzst` as gzip during uploads. It'd save us some compute if it didn't try to do this for already-compressed archives.

PR #1122, opened 2022-06-20
Graceful handling of error (non-validation one) 

tiwarishub:
We recently saw due to unavailability of Actions cache service, some of the setup-* start failing because of their caching feature. Ideally from actions-cache team it is recommended to handle cache service error (non-validation) gracefully and not to fail to workflow. This PR moves that dependency to caches packages, so that we will not throw error and throw the error as warning/info to stderr/stdout.

### DIFF
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -5,7 +5,6 @@ import * as cacheHttpClient from './internal/cacheHttpClient'
 import {createTar, extractTar, listTar} from './internal/tar'
 import {DownloadOptions, UploadOptions} from './options'
 
-
 export class ValidationError extends Error {
   constructor(message: string) {
     super(message)
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -5,6 +5,7 @@ import * as cacheHttpClient from './internal/cacheHttpClient'
 import {createTar, extractTar, listTar} from './internal/tar'
 import {DownloadOptions, UploadOptions} from './options'
 
+
 export class ValidationError extends Error {
   constructor(message: string) {
     super(message)
@@ -128,7 +129,7 @@ export async function restoreCache(
     return cacheEntry.cacheKey
   } catch (error) {
     // Supress all cache related errors because caching should be optional
-    core.warning(`Fail to restore: ${error}`)
+    core.warning(`Fail to restore: ${(error as Error).message}`)
   } finally {
     // Try to delete the archive to save space
     try {
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -128,7 +128,7 @@ export async function restoreCache(
     return cacheEntry.cacheKey
   } catch (error) {
     // Supress all cache related errors because caching should be optional
-    core.warning(`Fail to restore: ${(error as Error).message}`)
+    core.warning(`Failed to restore: ${(error as Error).message}`)
   } finally {
     // Try to delete the archive to save space
     try {
@@ -226,9 +226,9 @@ export async function saveCache(
   } catch (error) {
     const typedError = error as Error
     if (typedError.name === ReserveCacheError.name) {
-      core.info(`Fail to save: ${typedError.message}`)
+      core.info(`Failed to save: ${typedError.message}`)
     } else {
-      core.warning(`Fail to save: ${typedError.message}`)
+      core.warning(`Failed to save: ${typedError.message}`)
     }
   } finally {
     // Try to delete the archive to save space
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -4,6 +4,7 @@ import * as utils from './internal/cacheUtils'
 import * as cacheHttpClient from './internal/cacheHttpClient'
 import {createTar, extractTar, listTar} from './internal/tar'
 import {DownloadOptions, UploadOptions} from './options'
+import { ArtifactCacheEntry } from './internal/contracts'
 
 export class ValidationError extends Error {
   constructor(message: string) {
@@ -86,23 +87,24 @@ export async function restoreCache(
   }
 
   const compressionMethod = await utils.getCompressionMethod()
-
-  // path are needed to compute version
-  const cacheEntry = await cacheHttpClient.getCacheEntry(keys, paths, {
-    compressionMethod
-  })
-  if (!cacheEntry?.archiveLocation) {
-    // Cache not found
-    return undefined
-  }
-
-  const archivePath = path.join(
-    await utils.createTempDirectory(),
-    utils.getCacheFileName(compressionMethod)
-  )
-  core.debug(`Archive Path: ${archivePath}`)
-
+  let archivePath = ""
   try {
+      // path are needed to compute version
+      const cacheEntry = await cacheHttpClient.getCacheEntry(keys, paths, {
+      compressionMethod
+    })
+
+    if (!cacheEntry?.archiveLocation) {
+      // Cache not found
+      return undefined
+    }
+  
+      archivePath = path.join(
+      await utils.createTempDirectory(),
+      utils.getCacheFileName(compressionMethod)
+    )
+    core.debug(`Archive Path: ${archivePath}`)
+  
     // Download the cache from the cache entry
     await cacheHttpClient.downloadCache(
       cacheEntry.archiveLocation,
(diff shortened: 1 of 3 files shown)


packages/cache/src/cache.ts
```diff
@@ -86,23 +86,24 @@ export async function restoreCache(
   }
 
   const compressionMethod = await utils.getCompressionMethod()
+  let archivePath = ''
+  try {
```
bishal-pdMSFT, line 90: So this will essentially suppress all failures including temp directory creation and tar extraction failures. Essentially we are trading accuracy for continuity here. We definitely need a monitor which will tell us how many users are hitting such failures.

packages/cache/src/cache.ts
```diff
@@ -217,6 +223,13 @@ export async function saveCache(
 
     core.debug(`Saving Cache (ID: ${cacheId})`)
     await cacheHttpClient.saveCache(cacheId, archivePath, options)
+  } catch (error) {
+    const typedError = error as Error
+    if (typedError.name === ReserveCacheError.name) {
```
bishal-pdMSFT, line 228: Should this check happen in action itself? I see that we do some [checks](https://github.com/actions/cache/blob/c3f1317a9e7b1ef106c153ac8c0f00fed3ddbc0d/src/restore.ts#L57) in action itself. Should we also not do this check there?

packages/cache/src/cache.ts
```diff
@@ -152,7 +158,7 @@ export async function saveCache(
   checkKey(key)
 
   const compressionMethod = await utils.getCompressionMethod()
-  let cacheId = null
+  let cacheId = -1
```
bishal-pdMSFT, line 161: why this change? IIRC this was made `null` recently. May @t-dedah can chime in on why it was made null.

packages/cache/src/cache.ts
```diff
@@ -86,23 +86,24 @@ export async function restoreCache(
   }
 
   const compressionMethod = await utils.getCompressionMethod()
+  let archivePath = ''
+  try {
+    // path are needed to compute version
+    const cacheEntry = await cacheHttpClient.getCacheEntry(keys, paths, {
```
aparna-ravindra, line 92: can the try block be just around the http client to handle errors from the server?

packages/cache/src/cache.ts
```diff
@@ -123,6 +124,11 @@ export async function restoreCache(
 
     await extractTar(archivePath, compressionMethod)
     core.info('Cache restored successfully')
+
+    return cacheEntry.cacheKey
+  } catch (error) {
+    // Supress all cache related errors because caching should be optional
+    core.warning(`Fail to restore: ${error}`)
```
bishal-pdMSFT, line 131:
```suggestion
    core.warning(`Failed to restore: ${error}`)
```

packages/cache/src/cache.ts
bishal-pdMSFT: I guess the idea here is to ensure that _any_ failure is suppressed so that user does not get blocked due to cache action failures.
A bit scary thought but I guess it would be good for users' workflows to continue executing as in reality there is no blocker due to cache action failure. This may mean that users may not even notice things like archive/unarchive failing, and that's why I feel we should have [monitors](https://github.com/actions/toolkit/pull/1122#discussion_r901311622) to ability to detect any anomaly in such failures.

packages/cache/src/cache.ts
tiwarishub: Okay, let me check with @t-dedah  because this function return type ``Number`` https://github.com/actions/toolkit/blob/main/packages/cache/src/cache.ts#L150 so in case of exception returning null might not be correct

packages/cache/src/cache.ts
tiwarishub: done

packages/cache/src/cache.ts
tiwarishub: the condition which you mentioned in actions will be removed after this change and basically idea here is we will also need to add any checks like this in other actions setup-* or wherever this toolkit  cache package  is getting used

packages/cache/src/cache.ts
tiwarishub: This suppression is already there https://github.com/actions/cache/blob/c3f1317a9e7b1ef106c153ac8c0f00fed3ddbc0d/src/restore.ts#L55-L62  and https://github.com/actions/cache/blob/main/src/save.ts#L51-L59 , the only thing is now it is moved to toolkit instead of Github Actions.

tiwarishub: > We definitely need a monitor which will tell us how many users are hitting such failures.

I will check how we can monitors for this

packages/cache/src/cache.ts
tiwarishub: Let me explain more why this change is required, earlier even though Function return type is number and ``cacheId`` was set null but in the code we were making sure this will never return null https://github.com/actions/toolkit/blob/main/packages/cache/src/cache.ts#L204 otherwise we  were throwing error so this line https://github.com/actions/toolkit/blob/main/packages/cache/src/cache.ts#L229 will never get execute (in case of exception). So compiler was happy. but now as i have added catch block https://github.com/actions/toolkit/blob/main/packages/cache/src/cache.ts#L229 line can be executed (in case of no error or exception is consumed) so compiler is throwing error that ``return type can only be number not number | null

tiwarishub:
@bishal-pdMSFT / @aparna-ravindra, Let me know if you this PR is good to go

packages/cache/src/cache.ts
bishal-pdMSFT: @tiwarishub but I don't see you catching the ValidationError and handling it. Does that still happen in action while rest happens in toolkit?

tiwarishub: no, in the given try block we are not throwing ValidationError exception https://github.com/actions/toolkit/pull/1122/files#diff-34e72845d1658a8a9b9fdfe6b5b9261952e7f0496a0a3fcec87dae722718a64aL151-L152. So ValidationError will be still thrown as a error

bishal-pdMSFT: Got it. Though I feel we should also add the ValidationError check here as well. We do not have a specific Validate method and that means the validation can potentially be added in many places in future and without a check all those validations will get suppressed with this global suppression and will be difficult to understand why that is happening.

packages/cache/src/cache.ts
bishal-pdMSFT: Yep, that seems to be the case.

packages/cache/src/cache.ts
bishal-pdMSFT: This is same problem as what @anuragc617 would be looking into w.r.t getting warnings from cache action so that we can have monitor on those.

bishal-pdMSFT:
Will this be a breaking change as we are changing the error contract. We are now not throwing errors which were previously being thrown.

bishal-pdMSFT:
Otherwise looks good to me.

packages/cache/src/cache.ts
tiwarishub: added

tiwarishub:
> Will this be a breaking change as we are changing the error contract. We are now not throwing errors which were previously being thrown.
Yes, i have seen in restore, Github actions generally checks for cache returned key so there we are good.

For actions/cache:
https://github.com/actions/cache/blob/main/src/restore.ts#L33
For Setup-java:
Restore cache: https://github.com/actions/setup-java/blob/main/src/cache.ts#L101-L104
For Setup-node:
https://github.com/actions/setup-node/blob/main/src/cache-restore.ts#L44
For Setup-python:
https://github.com/actions/setup-python/blob/main/src/cache-distributions/cache-distributor.ts#L38

But for save, Actions generally trust to exception to get raised if there is issues to stop execution of their current function. So there we need to check to change to check return cacheId as well.

For actions/cache: https://github.com/actions/cache/blob/main/src/save.ts#L48
For Setup-java: https://github.com/actions/setup-java/blob/main/src/cache.ts#L127
For Setup-python: https://github.com/actions/setup-python/blob/main/src/cache-save.ts#L47
For setup-node: https://github.com/actions/setup-node/blob/main/src/cache-save.ts#L53

@bishal-pdMSFT , should we go ahead with major release then ?

bishal-pdMSFT:
New major version looks good to me

vsafonkin (approved):

PR #1132, opened 2022-07-13
Fix Empty Cache save when the path is github workspace directory

pdotl:
This PR fixes the [#833](https://github.com/actions/cache/issues/833) issue in actions/cache.

**Issue Summary**: Cache being saved is empty in case of path being current directory or github.workspace. So github.workspace directory can never get cached.

**Root cause**:  The resolvePaths function returns empty since the relative path of the path provided in cache's path param is calculated against the workspace directory itself. 

**Resolution**: We manually add '.' to paths list since the list will be empty in the above scenario.

Feedback is welcome regarding the implementation and possible alternatives. 👀

### DIFF
--- a/packages/cache/__tests__/cacheUtils.test.ts
+++ b/packages/cache/__tests__/cacheUtils.test.ts
@@ -1,6 +1,5 @@
 import {promises as fs} from 'fs'
 import * as path from 'path'
-import { coerce } from 'semver'
 import * as cacheUtils from '../src/internal/cacheUtils'
 
 test('getArchiveFileSizeInBytes returns file size', () => {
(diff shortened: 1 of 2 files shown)


aparna-ravindra:
Please check the failing test cases?

packages/cache/__tests__/cacheUtils.test.ts
```diff
@@ -32,3 +32,8 @@ test('assertDefined throws if undefined', () => {
 test('assertDefined returns value', () => {
   expect(cacheUtils.assertDefined('test', 5)).toBe(5)
 })
+
+test('resolvePaths works on current directory', async () => {
+  const paths = await cacheUtils.resolvePaths(['.'])
```
aparna-ravindra, line 37: it would be better to set the workspace path as the param here and check that the result is not empty.

tiwarishub (approved):
LGTM. If you have time, let's also create a PR audit fix

pdotl:
I'll create a separate PR for npm audit fix. 👍

PR #1140, opened 2022-07-29
Introducing timeout in segment download to avoid the download stuck problem

bishal-pdMSFT:
This PR focuses on fixing the download stuck problem that had been reported by several users in the past, especially when caches are of larger sizes (>= 2GB).

To avoid the problem we've introduced a timeout that aborts the stuck download if not completed within 1 hour.

We have kept the timeout as 1 hour because the cache sizes can be very huge and on self hosted runners, the downloads might not be as fast as GitHub hosted runners, hence to accommodate for everyone and avoid inconvenience for some users the timeout has been set to 1 hour. 

PS - `lerna` has been upgraded to version 5.x because 4.x has some vulnerability that was being caught by the audit action. The upgrade has been made using npm's `audit fix` tool.

### DIFF
--- a/packages/cache/src/internal/downloadUtils.ts
+++ b/packages/cache/src/internal/downloadUtils.ts
@@ -13,6 +13,8 @@ import {SocketTimeout} from './constants'
 import {DownloadOptions} from '../options'
 import {retryHttpClientResponse} from './requestUtils'
 
+import { AbortController } from "@azure/abort-controller";
+
 /**
  * Pipes the body of a HTTP response to a stream
  *
@@ -248,7 +250,8 @@ export async function downloadCacheStorageSDK(
 
     try {
       downloadProgress.startDisplayTimer()
-
+      const controller = new AbortController();
+      const abortSignal = controller.signal;
       while (!downloadProgress.isDone()) {
         const segmentStart =
           downloadProgress.segmentOffset + downloadProgress.segmentSize
@@ -264,14 +267,16 @@ export async function downloadCacheStorageSDK(
             segmentStart,
             segmentSize,
             {
+              abortSignal: abortSignal,
               concurrency: options.downloadConcurrency,
               onProgress: downloadProgress.onProgress()
             }
           ),
-          timer.setTimeout(60 * 60 * 1000, 'timeout')]);
+          timer.setTimeout(options.abortTimeInMs, 'timeout')]);
           
         if(result === 'timeout') {
-          throw new Error("Segment download timed out");
+          controller.abort();
+          throw new Error("Download aborted, segment download timed out.");
         }
 
         fs.writeFileSync(fd, result)
--- a/packages/cache/src/internal/downloadUtils.ts
+++ b/packages/cache/src/internal/downloadUtils.ts
@@ -249,9 +249,9 @@ export async function downloadCacheStorageSDK(
 
     try {
       downloadProgress.startDisplayTimer()
-      const abortTimeInMs =
-        options.abortTimeInMs === undefined ? 3600000 : options.abortTimeInMs
-      const abortSignal = AbortController.timeout(abortTimeInMs)
+      const segmentTimeoutInMs =
+        options.segmentTimeoutInMs === undefined ? 3600000 : options.segmentTimeoutInMs
+      const abortSignal = AbortController.timeout(segmentTimeoutInMs)
       abortSignal.addEventListener('abort', () => {
         core.warning('Cache download aborted, segment download timed out.')
       })
--- a/packages/cache/src/internal/downloadUtils.ts
+++ b/packages/cache/src/internal/downloadUtils.ts
@@ -250,7 +250,9 @@ export async function downloadCacheStorageSDK(
     try {
       downloadProgress.startDisplayTimer()
       const segmentTimeoutInMs =
-        options.segmentTimeoutInMs === undefined ? 3600000 : options.segmentTimeoutInMs
+        options.segmentTimeoutInMs === undefined
+          ? 3600000
+          : options.segmentTimeoutInMs
       const abortSignal = AbortController.timeout(segmentTimeoutInMs)
       abortSignal.addEventListener('abort', () => {
         core.warning('Cache download aborted, segment download timed out.')
--- a/packages/cache/src/internal/downloadUtils.ts
+++ b/packages/cache/src/internal/downloadUtils.ts
@@ -249,13 +249,11 @@ export async function downloadCacheStorageSDK(
 
     try {
       downloadProgress.startDisplayTimer()
-      const segmentTimeoutInMs =
-        options.segmentTimeoutInMs === undefined
-          ? 3600000
-          : options.segmentTimeoutInMs
-      const abortSignal = AbortController.timeout(segmentTimeoutInMs)
+      const abortSignal = AbortController.timeout(
+        options.segmentTimeoutInMs || 3600000
+      )
       abortSignal.addEventListener('abort', () => {
-        core.warning('Cache download aborted, segment download timed out.')
+        core.warning('Aborting cache download as it exceeded the timeout.')
       })
       while (!downloadProgress.isDone()) {
         const segmentStart =
@@ -277,7 +275,6 @@ export async function downloadCacheStorageSDK(
             onProgress: downloadProgress.onProgress()
           }
         )
-
         fs.writeFileSync(fd, result)
       }
     } finally {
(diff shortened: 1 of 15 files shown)


packages/cache/src/internal/downloadUtils.ts
```diff
@@ -258,21 +260,41 @@ export async function downloadCacheStorageSDK(
...
-        const result = await client.downloadToBuffer(
-          segmentStart,
-          segmentSize,
-          {
+        const abortTimeInMs =
+          options.abortTimeInMs === undefined ? 3600000 : options.abortTimeInMs
+        const result = await promiseWithTimeout(
+          abortTimeInMs,
+          client.downloadToBuffer(segmentStart, segmentSize, {
+            abortSignal: controller.signal,
```
bishal-pdMSFT, line 268: The doc seems to suggest that the async work will get aborted on timeout.
```
const signal = AbortController.timeout(1000);
doAsyncWork({ abortSignal: signal });
```

Do we still need to handle timeout from our side?

packages/cache/src/internal/downloadUtils.ts
```diff
@@ -258,21 +260,41 @@ export async function downloadCacheStorageSDK(
...
+  timeoutMs: number,
+  promise: Promise<Buffer>
+): Promise<unknown> => {
+  let timeoutHandle: NodeJS.Timeout
+  const timeoutPromise = new Promise(resolve => {
+    timeoutHandle = setTimeout(() => resolve('timeout'), timeoutMs)
+  })
+
+  return Promise.race([promise, timeoutPromise]).then(result => {
+    clearTimeout(timeoutHandle)
```
bishal-pdMSFT, line 297: The promise timeout implementation looks very similar to what we were trying earlier but saw that promise still kept on awaiting till timeout was done. Is `clearTimeout` fixing that here?

packages/cache/src/internal/downloadUtils.ts
```diff
@@ -258,21 +260,41 @@ export async function downloadCacheStorageSDK(
...
 
         downloadProgress.nextSegment(segmentSize)
-
-        const result = await client.downloadToBuffer(
-          segmentStart,
-          segmentSize,
-          {
+        const abortTimeInMs =
+          options.abortTimeInMs === undefined ? 3600000 : options.abortTimeInMs
+        const result = await promiseWithTimeout(
```
bishal-pdMSFT, line 265: Add unit tests if possible

packages/cache/src/internal/downloadUtils.ts
kotewar: Yes, current implementation of [Timers promises API](https://nodejs.org/api/timers.html#timers-promises-api) only have an option to `setTimeout` but not clear it. So due to this, once the download was completed, there was no way to clear the timeout and the process would run for 1 hour straight. `clearTimeout` here will clear the timeout as soon as any of the promise is complete.

packages/cache/src/internal/downloadUtils.ts
kotewar: Didn't come across this, let me give it a try! 👍🏽

bishal-pdMSFT: [This](https://www.npmjs.com/package/@azure/abort-controller) is doc I referred above.

packages/cache/src/internal/downloadUtils.ts
bishal-pdMSFT: That's interesting! I did not think not clearing timer will keep the thread locked

bishal-pdMSFT: Ah! it indeed is the case! https://nodejs.org/api/timers.html#class-timeout

> By default, when a timer is scheduled using either [setTimeout()](https://nodejs.org/api/timers.html#settimeoutcallback-delay-args) or [setInterval()](https://nodejs.org/api/timers.html#setintervalcallback-delay-args), the Node.js event loop will continue running as long as the timer is active.

bishal-pdMSFT: [Unref](https://nodejs.org/api/timers.html#timeoutunref) can be used to not block https://stackoverflow.com/a/43255345

kotewar: Yes! Btw, I am trying `AbortController.timeout(timeoutInMs)` right now and it is looking good for now. Working on error handling as the [error message](https://github.com/bbq-beets/test-cache-load/runs/7689143074?check_suite_focus=true#step:2:34) is not that good.

packages/cache/src/internal/downloadUtils.ts
kotewar: Added

packages/cache/src/internal/downloadUtils.ts
```diff
@@ -247,7 +249,12 @@ export async function downloadCacheStorageSDK(
 
     try {
       downloadProgress.startDisplayTimer()
-
+      const abortTimeInMs =
```
bishal-pdMSFT, line 252: Better to handle undefined `abortTimeInMs` when instantiating `options`. This way all defaulting is in one place.

packages/cache/src/internal/downloadUtils.ts
```diff
@@ -247,7 +249,12 @@ export async function downloadCacheStorageSDK(
...
+        options.abortTimeInMs === undefined ? 3600000 : options.abortTimeInMs
+      const abortSignal = AbortController.timeout(abortTimeInMs)
+      abortSignal.addEventListener('abort', () => {
+        core.warning('Cache download aborted, segment download timed out.')
```
bishal-pdMSFT, line 256: Don't we have to throw error here or will `downloadToBuffer` throw? If latter then do we really need to log warning here or will the error contains timeout information?

packages/cache/src/options.ts
```diff
@@ -46,6 +46,13 @@ export interface DownloadOptions {
    * @default 30000
    */
   timeoutInMs?: number
+
+  /**
+   * Time after which download should be aborted if stuck
+   *
+   * @default 3600000
+   */
+  abortTimeInMs?: number
```
bishal-pdMSFT, line 55: This is also a timeout. Previous timeout is a chunk timeout  hence we can name this as `segmentTimeoutInMs`?

packages/cache/src/internal/downloadUtils.ts
kotewar: Throwing new error here [stops the workflow with errors](https://github.com/bbq-beets/test-cache-load/runs/7689586482?check_suite_focus=true#step:2:34). Hence just throwing warning to let the rest of the workflow [succeed, but with warnings](https://github.com/bbq-beets/test-cache-load/actions/runs/2803164508)

tiwarishub: @kotewar , that error will be catched here right https://github.com/actions/toolkit/blob/main/packages/cache/src/cache.ts#L129-L136 and set as warning?
THrowing error might be useful here so that partial download dont get untarred [here](https://github.com/actions/toolkit/blob/main/packages/cache/src/cache.ts#L125)

kotewar: Its not getting caught as a warning, instead the execution is stopped. Please check [this output](https://github.com/bbq-beets/test-cache-load/runs/7689586482?check_suite_focus=true#step:2:34)

packages/cache/src/internal/downloadUtils.ts
kotewar: We are doing the instantiation [here](https://github.com/actions/toolkit/blob/segment-download-timeout/packages/cache/src/options.ts#L95) while getting the options, but since the member is [declared optional](https://github.com/actions/toolkit/blob/segment-download-timeout/packages/cache/src/options.ts#L55) the undefined check is required.
Screenshot 2022-08-08 at 10 22 37 AM

packages/cache/src/internal/downloadUtils.ts
kotewar: @tiwarishub I tested by added a new debug log in my repo - https://github.com/kotewar/cache/blob/main/src/restore.ts#L48

This log should have printed in the workflow file, but since the `cacheKey` is not found, the restore run is exiting there. Check [this run](https://github.com/bbq-beets/test-cache-load/runs/7719216695?check_suite_focus=true#step:2:36)

tiwarishub (approved):

smorimoto:
The bad thing about this is that PR is implemented on a fairly strong assumption that the existence of the cache is always optional. Some software doesn't make it optional, and it's technically wrong for caching, but in that case this really doesn't work anyway. You don't need to do this for this fairly low-level package, and this work should definitely be done only on the actions/cache side. It's really enough to leave a reference to that implementation in the documentation of this package.

smorimoto:
In short, it's not a bad thing to abort, but it should throw an error anyway.

PR #891, opened 2021-08-23
chore(ci): update dependencies

mmorel-35:
Hi ! I have updated the current dependencies with the help of dependabot.

PR #888, opened 2021-08-20
Fix endColumn documentation

martincostello:
Change "start" to "end".

PR #1152, opened 2022-08-13
Fix zstd failing on windows when used with the gnu tar workaround

pdotl:
This PR aims to resolve the issues [#888](https://github.com/actions/cache/issues/891) and [#891](https://github.com/actions/cache/issues/891) in `actions/cache`.

#### Context 
We had introduced zstd aliases in `actions/cache` `v3.0.6` to fix issues with zstd failing for tar version < 1.30. That change was made in understanding that we have zstd disabled on windows(bsd tar). However, due to dismal gzip performance on windows runners, developers have been using a workaround to use gnu tar along with zstd. The above release broke this workaround. 

#### Approach
We have switched to using short flag versions(ex. `zstd -d`) of zstd command for windows on gnu tar instead of the alias versions(ex. `unzstd`). This restores the workaround functionality same as before the release `v3.0.6`. This approach has been implemented in this PR.

The downsides of using this approach:
- The change is quite specific in nature and doesn't cover other potential issues we might face with using zstd aliases.

#### Alternative Approach
We also considered reverting the change done in `v3.0.6` and instead implement zstd aliases based on tar versions. This would also work correctly as it has been working before. We only use zstd aliases for tar version < 1.30. 

The downside of using this approach:
- Extra system call to tar version to fetch version for every cache call.
- Parsing of tar version stdout output. The tar version is not in semantic versioning format so could be brittle to parse and test. 
- Also tar version parsing needs to be implemented.

Feedback is welcome for choosing between both approaches and code review as well. 👀

### DIFF
--- a/packages/cache/src/internal/tar.ts
+++ b/packages/cache/src/internal/tar.ts
@@ -3,7 +3,9 @@ import * as io from '@actions/io'
 import {existsSync, writeFileSync} from 'fs'
 import * as path from 'path'
 import * as utils from './cacheUtils'
-import {CompressionMethod} from './constants'
+import {CompressionMethod, TarOperation} from './constants'
+
+const IS_WINDOWS = process.platform === 'win32'
 
 async function getTarPath(
   args: string[],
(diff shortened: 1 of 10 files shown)


packages/cache/src/internal/tar.ts
```diff
@@ -133,15 +147,22 @@ export async function listTar(
   archivePath: string,
   compressionMethod: CompressionMethod
 ): Promise<void> {
-  // --d: Decompress.
+  // -d: Decompress.
+  // unzstd is equivalent to 'zstd -d'
   // --long=#: Enables long distance matching with # bits.
   // Maximum is 30 (1GB) on 32-bit OS and 31 (2GB) on 64-bit.
   // Using 30 here because we also support 32-bit self-hosted runners.
   function getCompressionProgram(): string[] {
```
bishal-pdMSFT, line 155: This function seems repeated in two places. Shall we define a common function and use in both places.

pdotl: Yes we can do that. Currently it is an internal function defined in three places. I'll make the changes 👍

packages/cache/src/internal/tar.ts
```diff
@@ -61,14 +61,21 @@ export async function extractTar(
...
   await io.mkdirP(workingDirectory)
-  // --d: Decompress.
+  // -d: Decompress.
+  // unzstd is equivalent to 'zstd -d'
   // --long=#: Enables long distance matching with # bits. Maximum is 30 (1GB) on 32-bit OS and 31 (2GB) on 64-bit.
   // Using 30 here because we also support 32-bit self-hosted runners.
   function getCompressionProgram(): string[] {
     switch (compressionMethod) {
       case CompressionMethod.Zstd:
+        if (process.platform === 'win32') {
```
bishal-pdMSFT, line 71: May want to use a const to keep this simpler to read
`const IS_WINDOWS = process.platform === 'win32'`

bishal-pdMSFT (approved):

PR #1155, opened 2022-08-16
Added custom user inputted timeout

kotewar:
Added an env var `SEGMENT_DOWNLOAD_TIMEOUT_MINS` that overrides the 1 hour default timeout for decreasing the cache download timeout.

[A successful run ](https://github.com/bbq-beets/test-cache-load/actions/runs/2865816461)

### DIFF
--- a/packages/cache/README.md
+++ b/packages/cache/README.md
@@ -44,6 +44,6 @@ const cacheKey = await cache.restoreCache(paths, key, restoreKeys)
 
 ##### Cache segment restore timeout
 
-Version `v3.0.4` of cache package introduces a segment download timeout. A cache gets downloaded in multiple segments of fixed sizes (`1GB` for a `32-bit` runner and `2GB` for a `64-bit` runner). Sometimes, a segment download gets stuck which causes the workflow job to be stuck forever and fail. The segment download timeout will allow the segment download to get aborted and hence allow the job to proceed with a cache miss.
+A cache gets downloaded in multiple segments of fixed sizes (`1GB` for a `32-bit` runner and `2GB` for a `64-bit` runner). Sometimes, a segment download gets stuck which causes the workflow job to be stuck forever and fail. Version `v3.0.4` of cache package introduces a segment download timeout. The segment download timeout will allow the segment download to get aborted and hence allow the job to proceed with a cache miss.
 
-Default value of this timeout is 1 hour and can be customized by specifying an [environment variable](https://docs.github.com/en/actions/learn-github-actions/environment-variables) named `SEGMENT_DOWNLOAD_TIMEOUT_MINS`.
+Default value of this timeout is 60 minutes and can be customized by specifying an [environment variable](https://docs.github.com/en/actions/learn-github-actions/environment-variables) named `SEGMENT_DOWNLOAD_TIMEOUT_MINS` with timeout value in minutes.
--- a/packages/cache/README.md
+++ b/packages/cache/README.md
@@ -42,3 +42,8 @@ const restoreKeys = [
 const cacheKey = await cache.restoreCache(paths, key, restoreKeys)
 ```
 
+##### Cache segment restore timeout
+
+Starting `v3.0.5` of `actions/toolkit`, in case any issue occurs while downloading the cache, the download will be aborted by default within 1 hour if any `segment` doesn't download completely. A `segment` is limited to size of `1GB` for a `32-bit` runner and `2GB` for a `64-bit` runner. So for any cache that exceeds the size of one segment, multiple segments will be downloaded in sequence to complete the download.
+
+To customise the `segment` download timeout, an environment variable `SEGMENT_DOWNLOAD_TIMEOUT_MINS` needs to be set with the timeout minutes. This way the download wouldn't get stuck forever and proceed to next step in the workflow without any problem.
--- a/packages/cache/README.md
+++ b/packages/cache/README.md
@@ -44,6 +44,6 @@ const cacheKey = await cache.restoreCache(paths, key, restoreKeys)
 
 ##### Cache segment restore timeout
 
-Starting `v3.0.5` of `actions/toolkit`, in case any issue occurs while downloading the cache, the download will be aborted by default within 1 hour if any `segment` doesn't download completely. A `segment` is limited to size of `1GB` for a `32-bit` runner and `2GB` for a `64-bit` runner. So for any cache that exceeds the size of one segment, multiple segments will be downloaded in sequence to complete the download.
+Version `v3.0.4` of cache package introduces a segment download timeout. A cache gets downloaded in multiple segments of fixed sizes (`1GB` for a `32-bit` runner and `2GB` for a `64-bit` runner). Sometimes, a segment download gets stuck which causes the workflow job to be stuck forever and fail. The segment download timeout will allow the segment download to get aborted and hence allow the job to proceed with a cache miss.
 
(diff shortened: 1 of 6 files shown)


packages/cache/RELEASES.md
```diff
@@ -80,4 +80,7 @@
 - Added 1 hour timeout for the download stuck issue [#810](https://github.com/actions/cache/issues/810).
 
 ### 3.0.3
-- Bug fixes for download stuck issue [#810](https://github.com/actions/cache/issues/810).
\ No newline at end of file
+- Bug fixes for download stuck issue [#810](https://github.com/actions/cache/issues/810).
+
+### 3.0.4
+- Allowing users to provide a custom timeout as input for aborting cache download using an environment variable `CACHE_DOWNLOAD_TIMEOUT_MINS`. Default is 1 hour.
```
bishal-pdMSFT, line 86:
```suggestion
- Allowing users to provide a custom timeout as input for aborting download of a cache segment using an environment variable `CACHE_DOWNLOAD_TIMEOUT_MINS`. Default is 1 hour.
```

packages/cache/src/options.ts
```diff
@@ -112,10 +112,21 @@ export function getDownloadOptions(copy?: DownloadOptions): DownloadOptions {
       result.segmentTimeoutInMs = copy.segmentTimeoutInMs
     }
   }
-
+  const customDownloadTimeoutMins = process.env['CACHE_DOWNLOAD_TIMEOUT_MINS']
```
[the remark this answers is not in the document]
kotewar: Since the users are not aware of the internal working of the download client, hence kept it a generic one that everyone would understand. If we want to keep it `SEGMENT_DOWNLOAD_TIMEOUT_MINS`, we'll have to also explain more about what a segment is, how the download segment size differs in 32 bit and 64 bit systems, etc.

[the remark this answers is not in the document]
bishal-pdMSFT: That's a good point. It's just that customer might come back complaining that I set a timeout and the download timed out only after a longer time period. 
If it is helpful, we can document the process to download in segments and mention the timeout there. But agree that this is additional complexity.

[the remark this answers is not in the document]
kotewar: Will work on the documenting part then...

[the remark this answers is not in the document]
bishal-pdMSFT: @aparna-ravindra @anuragc617 any thoughts here?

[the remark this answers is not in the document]
tiwarishub: ``CACHE_DOWNLOAD_TIMEOUT_MINS`` This is not complete cache download timeout, It is timeout for 2GB Should we rename to reflect this ? like CACHE_DOWNLOAD_TIMEOUT_PER_GB and we can compute the timeout for 2gb internally.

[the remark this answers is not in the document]
kotewar: Does [readme](https://github.com/actions/toolkit/blob/4b6b45fe18c542cabfb32eb24f78a67377aa5255/packages/cache/README.md#cache-segment-restore-timeout) work for the same? @tiwarishub

packages/cache/README.md
```diff
@@ -42,3 +42,8 @@ const restoreKeys = [
 const cacheKey = await cache.restoreCache(paths, key, restoreKeys)
 ```
 
+##### Cache segment restore timeout
+
+Starting `v3.0.5` of `actions/toolkit`, in case any issue occurs while downloading the cache, the download will be aborted by default within 1 hour if any `segment` doesn't download completely. A `segment` is limited to size of `1GB` for a `32-bit` runner and `2GB` for a `64-bit` runner. So for any cache that exceeds the size of one segment, multiple segments will be downloaded in sequence to complete the download.
```
bishal-pdMSFT, line 47: Put the version of cache action as that is what is visible to users. Toolkit is an internal concept.

kotewar: I plan to put a similar readme in `actions/cache` where we can put the actions/cache version. I can remove the version from readme if its not needed. I just felt documenting the version in readme will be helpful for the readers of this readme file.

bishal-pdMSFT, line 47: Version vXXX of cache action introduces a segment download timeout. A cache gets downloaded in multiple segments of fixed sizes (`1GB` for a `32-bit` runner and `2GB` for a `64-bit` runner). Sometime a segment download gets stuck which causes the workflow job to be stuck. The timeout will allow the segment download to get aborted and hence allowing the job to proceed with cache miss.

Default value of this timeout is 1 hour and can be customized by specifying an [environment variable](https://docs.github.com/en/actions/learn-github-actions/environment-variables) named `SEGMENT_DOWNLOAD_TIMEOUT_MINS`.

bishal-pdMSFT: Ah this is toolkit README. My bad. This is good. But we should add same section in cache action readme as well.

kotewar: Yes!

packages/cache/RELEASES.md
```diff
@@ -80,4 +80,7 @@
 - Added 1 hour timeout for the download stuck issue [#810](https://github.com/actions/cache/issues/810).
 
 ### 3.0.3
-- Bug fixes for download stuck issue [#810](https://github.com/actions/cache/issues/810).
\ No newline at end of file
+- Bug fixes for download stuck issue [#810](https://github.com/actions/cache/issues/810).
+
+### 3.0.4
+-  Allowing users to provide a custom timeout as input for aborting download of a cache segment using an environment variable `CACHE_DOWNLOAD_TIMEOUT_MINS`. Default is 1 hour.
```
bishal-pdMSFT, line 86:
```suggestion
-  Allowing users to provide a custom timeout as input for aborting download of a cache segment using an environment variable `SEGMENT_DOWNLOAD_TIMEOUT_MIN`. Default is 1 hour.
```

packages/cache/README.md
kotewar: Made the changes

packages/cache/README.md
```diff
@@ -42,3 +42,8 @@ const restoreKeys = [
 const cacheKey = await cache.restoreCache(paths, key, restoreKeys)
 ```
 
+##### Cache segment restore timeout
+
+Version `v3.0.4` of cache package introduces a segment download timeout. A cache gets downloaded in multiple segments of fixed sizes (`1GB` for a `32-bit` runner and `2GB` for a `64-bit` runner). Sometimes, a segment download gets stuck which causes the workflow job to be stuck forever and fail. The segment download timeout will allow the segment download to get aborted and hence allow the job to proceed with a cache miss.
```
bishal-pdMSFT, line 47: A small change
```suggestion
A cache gets downloaded in multiple segments of fixed sizes (`1GB` for a `32-bit` runner and `2GB` for a `64-bit` runner). Sometimes, a segment download gets stuck which causes the workflow job to be stuck forever and fail. Version `v3.0.4` of cache package introduces a segment download timeout. The segment download timeout will allow the segment download to get aborted and hence allow the job to proceed with a cache miss.
```

packages/cache/README.md
```diff
@@ -42,3 +42,8 @@ const restoreKeys = [
...
+
+Default value of this timeout is 1 hour and can be customized by specifying an [environment variable](https://docs.github.com/en/actions/learn-github-actions/environment-variables) named `SEGMENT_DOWNLOAD_TIMEOUT_MINS`.
```
bishal-pdMSFT, line 49:
```suggestion
Default value of this timeout is 60 minutes and can be customized by specifying an [environment variable](https://docs.github.com/en/actions/learn-github-actions/environment-variables) named `SEGMENT_DOWNLOAD_TIMEOUT_MINS` with timeout value in minutes.
```

bishal-pdMSFT (approved):

PR #1178, opened 2022-09-13
Add save-state and set-output file commands

rentziass:
This adds support for the file command version of `save-state` and `set-output` that are being added to the runner in [#2118](https://github.com/actions/runner/pull/2118).

In the case that these commands are being used on a runner that doesn't support the file-command versions they will fallback to the pre-existing workflow commands.

### DIFF
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -1,10 +1,12 @@
 import {issue, issueCommand} from './command'
-import {issueCommand as issueFileCommand} from './file-command'
+import {
+  issueCommand as issueFileCommand,
+  prepareKeyValueMessage as prepareCommandValue
+} from './file-command'
 import {toCommandProperties, toCommandValue} from './utils'
 
 import * as os from 'os'
 import * as path from 'path'
-import {v4 as uuidv4} from 'uuid'
 
 import {OidcClient} from './oidc-utils'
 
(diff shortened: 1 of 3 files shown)


packages/core/src/core.ts
```diff
@@ -1,10 +1,12 @@
 import {issue, issueCommand} from './command'
-import {issueCommand as issueFileCommand} from './file-command'
+import {
+  issueCommand as issueFileCommand,
+  prepareKeyValueMessage as prepareCommandValue
```
thboop, line 4: super nit but I find it easier to keep names consistent rather then renaming them across files just for search-ability later one

packages/core/src/file-command.ts
```diff
@@ -22,3 +23,25 @@ export function issueCommand(command: string, message: any): void {
     encoding: 'utf8'
   })
 }
+
+export function prepareKeyValueMessage(key: string, value: any): string {
+  const delimiter = `ghadelimiter_${uuidv4()}`
```
thboop, line 28: 👍 on the refactor

packages/core/src/core.ts
rentziass: I've renamed `issueCommand -> issueFileCommand` at the source too to avoid confusion with the existing `issueCommand`, what do you think?

thboop (approved):
lgtm

ytarsha (approved):

bradyclifford:
When will this find its way into `actions/github-script@v6`?

packages/core/src/file-command.ts
```diff
@@ -22,3 +23,25 @@ export function issueCommand(command: string, message: any): void {
...
+  const convertedValue = toCommandValue(value)
+
+  // These should realistically never happen, but just in case someone finds a
+  // way to exploit uuid generation let's not allow keys or values that contain
+  // the delimiter.
+  if (key.includes(delimiter)) {
+    throw new Error(
+      `Unexpected input: name should not contain the delimiter "${delimiter}"`
+    )
+  }
```
skeggse, line 38: The check for the `delimiter` in the `key` feels a little misleading to me - if the `key` is user-controlled, it can still be used to add additional outputs if that's the risk being mitigated here; the delimiter doesn't capture the `key`, it just captures the `value`.
```suggestion
```

PR #1180, opened 2022-09-16
Resolve auditing issues

rentziass:
- Our tools audits are currently failing to a known [`parse-url < 8.1.0` vulnerability](https://github.com/advisories/GHSA-j9fq-vwqv-2fm2). `parse-url` is a Lerna dependency and the issue [has been reported](https://github.com/lerna/lerna/issues/3312).
- Some of the internal dependencies were outdate, causing the audit to fail


thboop (approved):
lgtm!

PR #1181, opened 2022-09-21
Add additionalPlugins parameter to getOctokit method

luketomlinson:
This allows clients to configure their Octokit/Github instance with OctoKit [plugins](https://github.com/octokit/core.js/#plugins). This is a variadic parameter, so it should not break existing calls to `getOctokit`

### DIFF
--- a/packages/github/src/github.ts
+++ b/packages/github/src/github.ts
@@ -2,7 +2,7 @@ import * as Context from './context'
 import {GitHub, getOctokitOptions} from './utils'
 
 // octokit + plugins
-import {OctokitOptions} from '@octokit/core/dist-types/types'
+import {OctokitOptions, OctokitPlugin} from '@octokit/core/dist-types/types'
 
 export const context = new Context.Context()
 
@@ -14,7 +14,12 @@ export const context = new Context.Context()
  */
 export function getOctokit(
   token: string,
-  options?: OctokitOptions
+  options?: OctokitOptions,
+  ...additionalPlugins: OctokitPlugin[]
 ): InstanceType<typeof GitHub> {
+  if (additionalPlugins.length) {
+    const GitHubWithPlugins = GitHub.plugin(...additionalPlugins)
+    return new GitHubWithPlugins(getOctokitOptions(token, options))
+  }
   return new GitHub(getOctokitOptions(token, options))
 }


packages/github/src/github.ts
```diff
@@ -14,7 +14,12 @@ export const context = new Context.Context()
  */
 export function getOctokit(
   token: string,
-  options?: OctokitOptions
+  options?: OctokitOptions,
+  ...additionalPlugins: OctokitPlugin[]
```
benwells, line 18: Since `options` is optional what happens here if the caller doesn't pass anything in as far as that argument goes? (e.g. `getOctokit(token)`).

Maybe we could set a default value to remedy and make `additionalPlugins` optional with no default as the new "last arg"?

```ts
options?: OctokitOptions = {},
additionalPlugins?: OctokitPlugin[]
```
(I'm sure the typing isn't right there with `OctokitOptions`, but you get the idea).

packages/github/src/github.ts
```diff
@@ -14,7 +14,12 @@ export const context = new Context.Context()
...
 ): InstanceType<typeof GitHub> {
+  if (additionalPlugins.length) {
```
benwells, line 20: Depending on how the args are ultimately structured, may be worth adding a null check here via optional chaining, in case additionalPlugins is left out entirely?  (to prevent "cannot read property length of undefined" type errors?).

```suggestion
  if (additionalPlugins?.length) {
```

packages/github/src/github.ts
luketomlinson: So reading some [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters#using_rest_parameters_in_combination_with_ordinary_parameters), the default for the "rest parameters" aka "variadic parameters" is an empty array. Typescript won't actually let you make this `...additionalPlugins?: OctokitPlugin[]`

packages/github/src/github.ts
cory-miller, line 20: Do you need to check if `additionalPlugins` has any values?

If additionalPlugins is empty, then it seems like an empty list[1] will be appended to the defaults[2].

1: https://github.com/octokit/core.js/blob/ebd1baf5fc5bb9652504990c669f358630221381/src/index.ts#L64

2: https://github.com/actions/toolkit/blob/64c334f0e54c45f0ee319b5707281d9df0c43a3d/packages/github/src/utils.ts#L20

cory-miller (approved):

packages/github/src/github.ts
benwells: My bad, I misunderstood the usage to be something like

```js
const plugins = [Plugin1, Plugin2];
getOctokit(token, options, plugins);
```

instead of 

```js
getOctokit(token, options, Plugin1, Plugin2);
```

Thanks for clarifying!

packages/github/src/github.ts
luketomlinson: oh so you're saying I can just skip the check altogether?

cory-miller: Not a TS expert, but the following looks like it will be functionally the same as checking `additionalPlugins` for length.

```
{
  const GitHub = GitHub.plugin(...additionalPlugins);
  return new GitHub(getOctokitOptions(token, options));
}
```

luketomlinson: yeah that's a fair point

PR #1183, opened 2022-09-25
feat(core): handle comma seperated lists in getMultilineInput

rickstaa:
This PR makes sure that the core.getMultilineInput function can also parse comma-separated lists (i.e. `[val1,val2,val3]`` and `val1,val2,val3`).


DmitriyBobrovskiy:
any update on this?

shoogle:
It would be better to put this in a new function `core.getCommaSeparatedInput()` or ` core.getJsonInput()`.

People might already be using `core.getMultilineInput()` with list items that contain commas, so it's not safe to start treating commas as a list separator retrospectively.

```yaml
- uses: myaction
  with:
    multiline-param: |
      item1, with comma
      item2 no comma
      item3, with, several, commas
```

This list has 3 items according to the current implementation, but your code would parse it as 7 items, potentially breaking the workflow.

iamstarkov:
@rickstaa I implemented the `core.getListInput (comma or new lines separated)` in #1926, so if you dont have time to implement it yourself you can review the mentioned PR

PR #1184, opened 2022-09-27
[artifact] exempt .tar.zst files from compression

lucacasonato:
These files are already compressed with zstd - no need to attempt re-compression.


joshwlewis:
It looks like there is some usage of `.tzst` in this repo already:

https://github.com/actions/toolkit/blob/fc005283374f1cea5e03ea1caf959cd9ff12fdc2/packages/cache/src/internal/constants.ts#L3

Since `.tar.zstd` and `.tzst` are synonyms, maybe it makes sense to add both here?

PR #1185, opened 2022-09-28
fix(core): `getMultilineInput` not trimming whitespace

### LINKED ISSUE #1186: core: `getMultilineInput` not trimming whitespace
**Describe the bug**
core.getInput is [trimming whitespace](https://github.com/actions/toolkit/blob/main/packages/core/src/core.ts#L17) by default, but core.getMultilineInput is not.
Also, core.getMultilineInput ignores trimWhitespace option.

**To Reproduce**

action.yml

```yml
# ...
inputs:
  option:
    description: a multiline option
```

index.js

```js
import * as core from "@actions/core"

const option = core.getMultilineInput("option", {
  trimWhitespace: true,
})

core.info(option)
```

workflow.yml

```yml
# ...
jobs:
  test:
    # ...
    steps:
      - uses: # ...
        with:
          option: |-
             a
             b 
             c 
```

-> `["a ", " b ", " c"]`

**Expected behavior**
-> `["a", "b", "c"]`

bicstone:
Fixed #1186

`core.getInput` is [trimming whitespace](https://github.com/actions/toolkit/blob/main/packages/core/src/core.ts#L17) by default, but `core.getMultilineInput` is not.
Also, `core.getMultilineInput` ignores `trimWhitespace` option.

This pull request has been fixed so that `core.getMultilineInput` will trim whitespace in the same way as `core.getInput`.

### DIFF
--- a/packages/core/__tests__/core.test.ts
+++ b/packages/core/__tests__/core.test.ts
@@ -33,8 +33,8 @@ const testEnvVars = {
   INPUT_BOOLEAN_INPUT_FALSE3: 'FALSE',
   INPUT_WRONG_BOOLEAN_INPUT: 'wrong',
   INPUT_WITH_TRAILING_WHITESPACE: '  some val  ',
-
   INPUT_MY_INPUT_LIST: 'val1\nval2\nval3',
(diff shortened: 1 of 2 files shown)


rentziass (approved):
Thank you for contributing this! Looking good to me, merging 🚀

PR #1203, opened 2022-10-11
Document another likely reason for problem matcher patterns failing

AlexTMjugador:
Input lines are passed as-is to the regular expressions used by problem matcher patterns. This can lead to unexpected failures when trying to, for example, match colored terminal output, which relies on [ANSI escape sequences](https://en.wikipedia.org/wiki/ANSI_escape_code). The cause is not that obvious at a first glance because such characters are stripped from the GitHub log UI, and I think that it should be documented.

### DIFF
--- a/docs/problem-matchers.md
+++ b/docs/problem-matchers.md
@@ -142,6 +142,8 @@ Some of the starter actions are already using problem matchers, for example:
 
 Use ECMAScript regular expression syntax when testing patterns.
 
+The patterns match the input lines as-is, without stripping ANSI escape codes or other extraneous characters.  If the subject strings can contain such characters, the regular expression must account for them.
+
 ### File property getting dropped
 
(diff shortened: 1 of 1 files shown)


AlexTMjugador:
I've noticed that GitHub runners contain code to strip out the aforementioned escape sequences:

https://github.com/actions/runner/blob/44d4d076fec3fddaf68c876cbea2217110d48892/src/Runner.Worker/Handlers/OutputManager.cs#L99-L100

However, not taking into account such sequences did not match the app error shown in the following workflow run log: https://github.com/ComunidadAylas/PackSquash-action/actions/runs/3226864102/jobs/5280915179

Its runner version is 2.298.2.

AlexTMjugador:
It's been a while since I opened this PR. Are there any plans to review it? It's still relevant to this day. Thank you.

PR #1206, opened 2022-10-13
Patch @actions/cache to use @action/core version 1.10.0

pdotl:
Update @action/cache package to use the version `1.10.0` of @actions/core.

### DIFF
--- a/packages/cache/package-lock.json
+++ b/packages/cache/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@actions/cache",
-      "version": "3.0.4",
+      "version": "3.0.5",
       "license": "MIT",
       "dependencies": {
         "@actions/core": "^1.10.0",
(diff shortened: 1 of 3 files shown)


bishal-pdMSFT:
Doesn't it need a code change to remove usage of `save-state` and `set-output`?

bishal-pdMSFT:
> Doesn't it need a code change to remove usage of `save-state` and `set-output`?

Looks like that is handled internally https://github.com/actions/toolkit/blob/1bc93f3cdf638ff423193f992119be08b7bb0666/packages/core/src/core.ts#L355

pdotl:
> 

The code changes happened on the `actions/cache` side [here](https://github.com/actions/cache/pull/950).

packages/cache/package-lock.json
```diff
@@ -1,625 +1,13 @@
 {
   "name": "@actions/cache",
-  "version": "3.0.4",
-  "lockfileVersion": 2,
```
Phantsure, line 4: Why did we change lockfile version?

packages/cache/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/cache",
-  "version": "3.0.4",
```
aparna-ravindra, line 3: please update RELEASES.md as well

packages/cache/package-lock.json
pdotl: This was changed on using `npm version patch` command. Reverting.

Phantsure (approved):
lgtm

PR #1207, opened 2022-10-13
Update cache: add getCacheEntry to check if a cache entry exists without having to download it

MartijnHols:
This adds a `getCacheEntry` export to the `cache` package that can be used to determine if a cache entry exists. When merged, this method can be used to add a flag to the the cache-action to only check if a cache entry exists for a key. Its `cache-hit` output can then by used to speed up workflows that use big caches and to skip work if the result of the work is already in the cache.

For context: I am using the cache action to share my build across workflow steps. If the build already exists, there is no point downloading it or building it again and the later steps can just reuse the old cached result.

Remake of #659 which I closed because it was ignored for a year. Rebased to latest working commit on main

### DIFF
--- a/packages/cache/__tests__/getCacheEntry.test.ts
+++ b/packages/cache/__tests__/getCacheEntry.test.ts
@@ -1,0 +1,169 @@
+import * as core from '@actions/core'
+import {getCacheEntry} from '../src/cache'
+import * as cacheHttpClient from '../src/internal/cacheHttpClient'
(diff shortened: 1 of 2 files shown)


cdce8p:
@MartijnHols I've opened #1286 which would (if merged) provide a new `dryRun` option for the `restore` function. I believe this might archive the same result as your `getCacheEntry` function while being a bit easier to implement and maintain. The cache key checks wouldn't need to be duplicated.

Would love to get your feedback.

PR #1223, opened 2022-10-25
Standardize behaviour of no_proxy environmental variable

### LINKED ISSUE #1172: no_proxy matching agains parts of the host name is not supported
It is quite common to define the `no_proxy` environment variable like this:

```
NO_PROXY=.part.of.some.domain.com,some-other-host.domain.com
```

In practice, this matches all contained hosts for many HTTP clients. But for action runners, it does not.  
Consider a GitHub host like `github.part.of.some.domain.com`. I would expect this host to be excluded from the HTTP proxy, but the Actions runner only recognizes this if I put the variable like this:

```
NO_PROXY=github.part.of.some.domain.com,.part.of.some.domain.com,some-other-host.domain.com
```

The relevant code is https://github.com/actions/toolkit/blob/main/packages/http-client/src/proxy.ts#L50

Perhaps the no_proxy matching could be broadened a bit?

felixlut:
The behaviour of the `no_proxy` environmental variable is [famously inconsistent](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/), but this change should follow how it usually works (i.e., that `no_proxy=domain.com` matches its subdomains, such as `some.domain.com`). 

Resolves #1172

### DIFF
--- a/packages/http-client/__tests__/proxy.test.ts
+++ b/packages/http-client/__tests__/proxy.test.ts
@@ -157,6 +157,12 @@ describe('proxy', () => {
     expect(bypass).toBeFalsy()
   })
 
+  it('checkBypass returns false if no_proxy is part of domain', () => {
+    process.env['no_proxy'] = 'myserver.com'
+    const bypass = pm.checkBypass(new URL('https://evilmyserver.com'))
+    expect(bypass).toBeFalsy()
+  })
+
   // Do not strip leading dots as per https://github.com/actions/runner/blob/97195bad5870e2ad0915ebfef1616083aacf5818/docs/adrs/0263-proxy-support.md
   it('checkBypass returns false if host with leading dot in no_proxy', () => {
     process.env['no_proxy'] = '.myserver.com'
--- a/packages/http-client/__tests__/proxy.test.ts
+++ b/packages/http-client/__tests__/proxy.test.ts
@@ -164,6 +164,12 @@ describe('proxy', () => {
     expect(bypass).toBeFalsy()
   })
 
+  it('checkBypass returns true if host with subdomain in no_proxy defined with leading "."', () => {
+    process.env['no_proxy'] = '.myserver.com'
+    const bypass = pm.checkBypass(new URL('https://sub.myserver.com'))
+    expect(bypass).toBeTruthy()
+  })
+
   // Do not match wildcard ("*") as per https://github.com/actions/runner/blob/97195bad5870e2ad0915ebfef1616083aacf5818/docs/adrs/0263-proxy-support.md
   it('checkBypass returns true if no_proxy is "*"', () => {
     process.env['no_proxy'] = '*'
--- a/packages/http-client/__tests__/proxy.test.ts
+++ b/packages/http-client/__tests__/proxy.test.ts
@@ -152,7 +152,7 @@ describe('proxy', () => {
   })
 
   it('checkBypass returns true if host with leading dot in no_proxy', () => {
-    process.env['no_proxy'] = '.myserver.com'
+    process.env['no_proxy'] = '.myserver.com' 
     const bypass = pm.checkBypass(new URL('https://myserver.com'))
     expect(bypass).toBeTruthy()
   })
--- a/packages/http-client/__tests__/proxy.test.ts
+++ b/packages/http-client/__tests__/proxy.test.ts
@@ -151,10 +151,11 @@ describe('proxy', () => {
     expect(bypass).toBeTruthy()
   })
 
-  it('checkBypass returns true if host with leading dot in no_proxy', () => {
+  // Do not strip leading dots as per https://github.com/actions/runner/blob/97195bad5870e2ad0915ebfef1616083aacf5818/docs/adrs/0263-proxy-support.md
+  it('checkBypass returns false if host with leading dot in no_proxy', () => {
     process.env['no_proxy'] = '.myserver.com'
     const bypass = pm.checkBypass(new URL('https://myserver.com'))
-    expect(bypass).toBeTruthy()
+    expect(bypass).toBeFalsy()
   })
 
   it('checkBypass returns false if no_proxy is subdomain', () => {
@@ -163,10 +164,11 @@ describe('proxy', () => {
     expect(bypass).toBeFalsy()
   })
 
+  // Do not match wildcard ("*") as per https://github.com/actions/runner/blob/97195bad5870e2ad0915ebfef1616083aacf5818/docs/adrs/0263-proxy-support.md
   it('checkBypass returns true if no_proxy is "*"', () => {
     process.env['no_proxy'] = '*'
     const bypass = pm.checkBypass(new URL('https://anything.whatsoever.com'))
-    expect(bypass).toBeTruthy()
+    expect(bypass).toBeFalsy()
   })
 
   it('HttpClient does basic http get request through proxy', async () => {
--- a/packages/http-client/__tests__/proxy.test.ts
+++ b/packages/http-client/__tests__/proxy.test.ts
@@ -151,16 +151,16 @@ describe('proxy', () => {
     expect(bypass).toBeTruthy()
   })
 
+  it('checkBypass returns false if no_proxy is subdomain', () => {
+    process.env['no_proxy'] = 'myserver.com'
+    const bypass = pm.checkBypass(new URL('https://myserver.com.evil.org'))
+    expect(bypass).toBeFalsy()
+  })
+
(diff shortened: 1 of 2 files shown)


packages/http-client/src/proxy.ts
```diff
@@ -51,7 +51,7 @@ export function checkBypass(reqUrl: URL): boolean {
     .split(',')
     .map(x => x.trim().toUpperCase())
     .filter(x => x)) {
-    if (upperReqHosts.some(x => x === upperNoProxyItem)) {
+    if (upperReqHosts.some(x => x.includes(upperNoProxyItem))) {
```
aibaars, line 54: I think it would be safer to use `endsWith` instead of `includes` to avoid bad actors from bypassing the proxy. For example if we set `no_proxy=mycompany.com`  then a domain like `mycompany.com.evil.org` should not bypass the proxy.

```suggestion
    if (upperReqHosts.some(x => x.endsWith(upperNoProxyItem))) {
```
The above would still let domains like `evilmycompany.com` bypass the proxy. Perhaps a better formulation would be
 
```suggestion
    if (upperReqHosts.some(x => x === upperNoProxyItem || x.endsWith('.' + upperNoProxyItem))) {
```

In addition we might want to strip off a leading `.`  to ensure that `no_proxy=example.com` and `no_proxy=.example.com` are treated the same.

fhammerl: +1 for @aibaars suggestions too, ensWith + '.' sounds good
NIT: I'd consider a small refactoring of the block from line 54, it's becoming a bit dense with filters

@felixlut Could you add a test-case that covers the change in https://github.com/actions/toolkit/blob/main/packages/http-client/__tests__/proxy.test.ts ?

PS the PR validation currently fails due to some older packages, I'll take a look in a different PR

felixlut: I like your suggestions, and sure I can add a test-case. I'll try to find time over the weekend!

felixlut: I've implemented the suggested changes (in addition to some other suggestions from the [article](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/#standardizing-no_proxy) I posted above), as well as test-cases for each of them. Namely the following was added:
* `*` matching all hosts
* Strip leading dots (`.domain.com` --> `domain.com`)

aibaars (approved):
This looks good to me. @fhammerl  , could you have a look at this?

fhammerl:
@TingluoHuang `actions/runner` for sure, can you think of any other services where we use noproxy?

TingluoHuang:
@fhammerl 

> @TingluoHuang `actions/runner` for sure, can you think of any other services where we use noproxy?

No

aibaars:
@fhammerl Could you update and merge https://github.com/actions/toolkit/pull/1226 ? I think that PR is meant to unblock the audit failure.

felixlut:
Haven't seen any activity in here for a couple of days. Just checking, there is nothing I should do now? This PR is just blocked until #1230 is merged?

felixlut:
Haven't seen any activity in here for a couple of days. Just checking, there is nothing I should do now? This PR is just blocked until #1230 is merged?

fhammerl (approved):
Great tests, thanks for the refactoring too.

LGTM

We're holding off with merging it for now as the 'audit' check is failing. 
We'll fix that first on main, then merge this branch.

felixlut:
@fhammerl any update?

fhammerl:
We're migrating the project over to Node16, will merge this PR right after.

gendergap:
> We're migrating the project over to Node16, will merge this PR right after.

Could you please reference the PR for that migration so that people waiting for this can track the progress?

vmjoseph:
@felixlut Would you mind committing one more time so that our checks can re-run? Once those clear we can get this ready to merge.

felixlut:
@vmjoseph Any update on this?

felixlut:
Any update on this? @fhammerl @aibaars

vmjoseph (approved):

vmjoseph (changes requested):
Looks like there's a linting error.

packages/http-client/__tests__/proxy.test.ts
```diff
@@ -145,6 +145,30 @@ describe('proxy', () => {
...
   })
 
+  it('checkBypass returns true if host with subdomain in no_proxy', () => {
+    process.env['no_proxy'] = 'myserver.com'
+    const bypass = pm.checkBypass(new URL('https://sub.myserver.com'))
+    expect(bypass).toBeTruthy()
+  })
+
+  it('checkBypass returns true if host with leading dot in no_proxy', () => {
+    process.env['no_proxy'] = '.myserver.com'
```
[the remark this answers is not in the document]
felixlut: I added some whitespace in my last commit to re-trigger the checks, that's probably it since they passed before! I've reverted it, but am not in a position to run the `format` or `lint-fix` right now. I'll take a look at that later tonight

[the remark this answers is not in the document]
felixlut: Or it works I guess :D 

Want me to run the `npm audit fix` and fix the bad dependencies?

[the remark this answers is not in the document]
fhammerl: Done it, should pass audit now

fhammerl:
Hey @felixlut 
This is a long lived PR now, thanks for sticking by!

We'd like to follow the no_proxy system we've established in [actions/runner](https://github.com/actions/runner/blob/main/docs/adrs/0263-proxy-support.md)

The runner rules today:

1. ✅ Support matching for subdomains (the meat of this PR)
2. ❌ Do NOT strip leading period in no_proxy entry (aka `no_proxy=.github.com` should bypass proxy for sub.github.com, but not for vanilla github.com urls)
3. ❌ Do NOT support `*` wildcard

This would bring the behaviour to that of `wget`'s if I understand the table correctly.

With those changes, I can get this PR merged ASAP.

Let me know if that works for you

felixlut:
Sounds good to me, as you say, the subdomains was the reason i opened this! I can make the required rollbacks later tonight, or more likely tomorrow!

packages/http-client/__tests__/proxy.test.ts
```diff
@@ -145,6 +145,32 @@ describe('proxy', () => {
...
+    expect(bypass).toBeFalsy()
+  })
+
+  // Do not match wildcard ("*") as per https://github.com/actions/runner/blob/97195bad5870e2ad0915ebfef1616083aacf5818/docs/adrs/0263-proxy-support.md
+  it('checkBypass returns true if no_proxy is "*"', () => {
+    process.env['no_proxy'] = '*'
+    const bypass = pm.checkBypass(new URL('https://anything.whatsoever.com'))
+    expect(bypass).toBeFalsy()
+  })
+
```
fhammerl, line 173: Can We add one more test to chech a subdomain is supported when no_proxy has a leading dot?

```
  it('checkBypass returns true if host with subdomain in no_proxy defined with leading "."', () => {
    process.env['no_proxy'] = '.myserver.com'
    const bypass = pm.checkBypass(new URL('https://sub.myserver.com'))
    expect(bypass).toBeTruthy()
  })
```

felixlut: Ye sure! 

Another question also, since you want the `no_proxy` to be the same as in the [`actions/runner`](https://github.com/actions/runner), do you also want me to re-structure the code to be more similar to how they handle it? 
Obviously it's not the same language, but might be easier to keep consistency between them if they are implemented using the same logic, just with different languages.  

Might be overkill, but I'm willing to do it if that makes it easier maintainable for you guys!

fhammerl: I'd prefer to keep PRs for the functional changes and refactoring changes separate, but really appreciate you asking 👍

Also, [here's](https://github.com/actions/runner/blob/5421fe3f7107f770c904ed4c7e506ae7a5cde2c2/src/Test/L0/RunnerWebProxyL0.cs#L342) a good set of unit tests from the runner code if needed

felixlut: I''ve added the [test you wrote](https://github.com/actions/toolkit/pull/1223#discussion_r1102785194), as well a new one for checking partial domains (such that `no_proxy=myserver.com` won't let `evilmyserver.com` through)

packages/http-client/__tests__/proxy.test.ts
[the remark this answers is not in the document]
felixlut: The lining and auditing is fixed now!

fhammerl:
@felixlut Thanks again for this, I'll ping you as we publish it 👍

PR #1232, opened 2022-11-07
Add GNUtar as default in windows

Phantsure:
GNUtar is present in windows hosted runners to default to it for archiving


packages/cache/src/internal/cacheUtils.ts
```diff
@@ -67,7 +67,7 @@ export async function unlinkFile(filePath: fs.PathLike): Promise<void> {
   return util.promisify(fs.unlink)(filePath)
 }
 
-async function getVersion(app: string): Promise<string> {
+export async function getVersion(app: string): Promise<string> {
```
Phantsure, line 70: Don't need to export this.

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -113,6 +114,21 @@ export async function getCacheEntry(
   const cacheResult = response.result
   const cacheDownloadUrl = cacheResult?.archiveLocation
   if (!cacheDownloadUrl) {
+    // List cache for primary key only if cache miss occurs
```
Phantsure, line 117: Is there a way to run this when debug step is setup?

packages/cache/__tests__/tar.test.ts
```diff
@@ -75,7 +82,7 @@ test('gzip extract tar', async () => {
 
   expect(mkdirMock).toHaveBeenCalledWith(workspace)
   const tarPath = IS_WINDOWS
-    ? `${process.env['windir']}\\System32\\tar.exe`
+    ? `${process.env['PROGRAMFILES']}\\Git\\usr\\bin\\tar.exe`
```
pdotl, line 85: Can we reduce the repetition of writing the complete path everywhere? Like use variable that can work with the extrapolation.

packages/cache/__tests__/tar.test.ts
```diff
@@ -87,14 +94,16 @@ test('gzip extract tar', async () => {
       '-P',
       '-C',
       IS_WINDOWS ? workspace?.replace(/\\/g, '/') : workspace
-    ].concat(IS_MAC ? ['--delay-directory-restore'] : []),
+    ]
```
pdotl, line 97: Same with this, if this can be extracted out in a single utility/setup method.

packages/cache/src/internal/tar.ts
```diff
@@ -14,14 +14,20 @@ async function getTarPath(
   switch (process.platform) {
     case 'win32': {
       const systemTar = `${process.env['windir']}\\System32\\tar.exe`
-      if (compressionMethod !== CompressionMethod.Gzip) {
+      const gnuTar = `${process.env['PROGRAMFILES']}\\Git\\usr\\bin\\tar.exe`
+      if (existsSync(gnuTar)) {
```
pdotl, line 18: Just curious, why does call to `existsSync` here does not have a `fs.` prefix but is present in the above `isGnuTarInstalled()` function?

packages/cache/src/internal/tar.ts
```diff
@@ -14,14 +14,20 @@ async function getTarPath(
...
+        // Use GNUtar as default on windows
+        args.push('--force-local')
+        return gnuTar
+      } else if (
+        compressionMethod !== CompressionMethod.Gzip ||
```
pdotl, line 23: This line might short circuit to add `--force-local` for `bsdtar`. The `else if` condition is reached when `existsSync(gnuTar)` is false. Now if the compression method is not `gzip`, `await utils.isGnuTarInstalled()` might not be evaluated.

packages/cache/src/internal/tar.ts
```diff
@@ -14,14 +14,20 @@ async function getTarPath(
...
+        (await utils.isGnuTarInstalled())
+      ) {
         // We only use zstandard compression on windows when gnu tar is installed due to
         // a bug with compressing large files with bsdtar + zstd
         args.push('--force-local')
       } else if (existsSync(systemTar)) {
```
pdotl, line 29: Nitpicking here. Can simplify the logic to either use `multiple ifs with returns` or use `if-else-if` logic. Having both make readability a bit confusing.

pdotl (commented):
Some minor comments.

packages/cache/__tests__/tar.test.ts
Phantsure: Let me try that

packages/cache/__tests__/tar.test.ts
Phantsure: I think this can be left like this as combination might need changing with different conditions

packages/cache/src/internal/tar.ts
Phantsure: Due to the way it is imported

packages/cache/src/internal/tar.ts
Phantsure: Thats why I have still not removed the condition check if only bsdtar exists then use gzip. Can still remove that condition as per our discussion as this is not a release per se but I would still want to keep things not breaking.

Phantsure: talking about this condition: https://github.com/actions/toolkit/blob/main/packages/cache/src/internal/cacheUtils.ts#L93

packages/cache/src/internal/tar.ts
Phantsure: Sure can do

Phantsure: Would this reduce the speed of the code?

Phantsure: Done

pdotl: I don't think it would, and will not be significant either way. It would make it easier to follow though IMO.

packages/cache/src/internal/tar.ts
pdotl: That makes sense. But we should still fix the logic here so it is not possible to have `--force-local` added incase of `bsdtar`. Since this might happen as soon as we remove the aforementioned condition in subsequent PRs.

Phantsure: Yes. That should be handled when we remove the condition then.

packages/cache/src/internal/tar.ts
pdotl: Again, just nitpicking here. I wanted to know should we make this consistent? Not sure about conventions in the javascript world, but it is considered good practice to avoid blanket `*` imports.

packages/cache/src/internal/tar.ts
pdotl: Sounds good. The subsequent PR should remove all `gzip` related logic.

pdotl (approved):
LGTM. A minor comment related to imports. Please take care of failing test cases other than the audit one.

packages/cache/src/internal/tar.ts
Phantsure: Looks like it is not a problem otherwise there should have been a caution in docs. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#namespace_import

Phantsure:
Done

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -117,8 +117,11 @@ export function getCacheFileName(compressionMethod: CompressionMethod): string {
 }
 
 export async function isGnuTarInstalled(): Promise<boolean> {
+  const gnuTar = `${process.env['PROGRAMFILES']}\\Git\\usr\\bin\\tar.exe`
```
bishal-pdMSFT, line 120: This path is used in multiple files. Move to `constants` file

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -117,8 +117,11 @@ export function getCacheFileName(compressionMethod: CompressionMethod): string {
 }
 
 export async function isGnuTarInstalled(): Promise<boolean> {
```
bishal-pdMSFT, line 119: Looks like this function is specifically for windows. Then rename it `isGnuTarInstalledOnWindows` just to ensure it is not used for linux by mistake.

packages/cache/src/internal/tar.ts
bishal-pdMSFT: >But we should still fix the logic here so it is not possible to have --force-local added incase of bsdtar

I came here to add same comment. This is adding `'--force-local'` for bsdtar. I would suggest to keep appropriate behaviour even though you want to change it in next PR.

packages/cache/src/internal/tar.ts
```diff
@@ -11,17 +11,24 @@ async function getTarPath(
...
+      const gnuTar = `${process.env['PROGRAMFILES']}\\Git\\usr\\bin\\tar.exe`
       const systemTar = `${process.env['windir']}\\System32\\tar.exe`
-      if (compressionMethod !== CompressionMethod.Gzip) {
+      if (existsSync(gnuTar)) {
+        // Use GNUtar as default on windows
+        args.push('--force-local')
+        tarPath = gnuTar
+      } else if (
+        compressionMethod !== CompressionMethod.Gzip ||
+        (await utils.isGnuTarInstalled())
```
bishal-pdMSFT, line 25: Please explain why this is added. The prior if already checks similar thing.

packages/cache/src/internal/tar.ts
```diff
@@ -11,17 +11,24 @@ async function getTarPath(
...
+  let tarPath = await io.which('tar', true)
   switch (process.platform) {
     case 'win32': {
+      const gnuTar = `${process.env['PROGRAMFILES']}\\Git\\usr\\bin\\tar.exe`
       const systemTar = `${process.env['windir']}\\System32\\tar.exe`
-      if (compressionMethod !== CompressionMethod.Gzip) {
+      if (existsSync(gnuTar)) {
+        // Use GNUtar as default on windows
+        args.push('--force-local')
+        tarPath = gnuTar
```
bishal-pdMSFT, line 22: Here `${process.env['PROGRAMFILES']}\\Git\\usr\\bin\\tar.exe` is used as hard-coded path but isGnuTarInstalled() also looks at direct tar command as well. I feel this is an inconsistency. can we use a common logic to find gnutar path and reuse it everywhere?

bishal-pdMSFT: My suggestion would be to have a common function named `getGnuTarPath` and return path else null. Use this function both here as well as in isGnuTarInstalled

packages/cache/src/internal/tar.ts
Phantsure: Currently incase of bsdtar this condition won't get hit as bsdtar only does compression with gzip

packages/cache/src/internal/cacheUtils.ts
Phantsure: Good point

packages/cache/src/internal/tar.ts
Phantsure: Okay. That makes sense

packages/cache/src/internal/cacheUtils.ts
Phantsure: Renamed to `getGnuTarPathOnWindows` inline with below comments

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -116,9 +120,12 @@ export function getCacheFileName(compressionMethod: CompressionMethod): string {
     : CacheFilename.Zstd
 }
 
-export async function isGnuTarInstalled(): Promise<boolean> {
+export async function getGnuTarPathOnWindows(): Promise<string> {
```
Phantsure, line 123: This returns any GNU tar present in path or returns empty string is equivalent to false

packages/cache/src/internal/constants.ts
```diff
@@ -21,3 +21,6 @@ export const DefaultRetryDelay = 5000
 // over the socket during this period, the socket is destroyed and the download
 // is aborted.
 export const SocketTimeout = 5000
+
+// The default path of GNUtar on hosted Windows runners
+export const GnuTarPathOnWindows = 'C:\\Program Files\\Git\\usr\\bin\\tar.exe'
```
bishal-pdMSFT, line 26: You should still be able to use the `env` for program files path here.

packages/cache/src/internal/cacheUtils.ts
bishal-pdMSFT, line 123: Looks like this function may be called multiple times. We can cache the result in a variable in this file.

packages/cache/src/internal/tar.ts
```diff
@@ -7,21 +7,18 @@ import {CompressionMethod} from './constants'
 
 const IS_WINDOWS = process.platform === 'win32'
 
-async function getTarPath(
-  args: string[],
-  compressionMethod: CompressionMethod
-): Promise<string> {
+async function getTarPath(args: string[]): Promise<string> {
+  let tarPath = await io.which('tar', true)
```
bishal-pdMSFT, line 11: I think this change does not look necessary now.

packages/cache/src/internal/tar.ts
```diff
@@ -7,21 +7,18 @@ import {CompressionMethod} from './constants'
...
     case 'win32': {
+      const gnuTar = await utils.getGnuTarPathOnWindows()
       const systemTar = `${process.env['windir']}\\System32\\tar.exe`
-      if (compressionMethod !== CompressionMethod.Gzip) {
-        // We only use zstandard compression on windows when gnu tar is installed due to
-        // a bug with compressing large files with bsdtar + zstd
+      if (gnuTar) {
+        // Use GNUtar as default on windows
         args.push('--force-local')
+        tarPath = gnuTar
```
bishal-pdMSFT, line 19: If you don't return and just break then L40 will override it again. Isn't it?

packages/cache/src/internal/tar.ts
bishal-pdMSFT: Ah it may still be needed as L40 is also changed.

packages/cache/src/internal/tar.ts
bishal-pdMSFT, line 19: Should we log debug message that tar is being overridden

packages/cache/src/internal/tar.ts
Phantsure: Right. It was added to easy readability but now all if conditions are returning so let me remove that

packages/cache/src/internal/cacheUtils.ts
Phantsure: It is called only two times

packages/cache/src/internal/tar.ts
Phantsure: Changed it back to previous way

bishal-pdMSFT (approved):

packages/cache/src/internal/cacheUtils.ts
bishal-pdMSFT: Ok

packages/cache/src/internal/tar.ts
bishal-pdMSFT: @Phantsure it still makes sense to add info log on which tar is being used.

PR #1234, opened 2022-11-14
Add logs for cache version on miss

Phantsure:
This PR should show cache version details on cache miss. Sample test workflow run: [here](https://github.com/bbq-beets/test-zstd/actions/runs/3710689378/jobs/6290420326#step:7:52)

### DIFF
--- a/packages/cache/src/internal/cacheHttpClient.ts
+++ b/packages/cache/src/internal/cacheHttpClient.ts
@@ -143,7 +143,7 @@ async function printCachesListForDiagnostics(
       core.debug(
         `No matching cache found for cache key '${key}', version '${version} and scope ${process.env['GITHUB_REF']}. There exist one or more cache(s) with similar key but they have different version or scope. See more info on cache matching here: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key \nOther caches with similar key:`
       )
-      for (const cacheEntry of cacheListResult.artifactCaches || []) {
+      for (const cacheEntry of cacheListResult?.artifactCaches || []) {
         core.debug(
           `Cache Key: ${cacheEntry?.cacheKey}, Cache Version: ${cacheEntry?.cacheVersion}, Cache Scope: ${cacheEntry?.scope}, Cache Created: ${cacheEntry?.creationTime}`
         )
--- a/packages/cache/src/internal/cacheHttpClient.ts
+++ b/packages/cache/src/internal/cacheHttpClient.ts
@@ -107,7 +107,7 @@ export async function getCacheEntry(
   if (response.statusCode === 204) {
     // List cache for primary key only if cache miss occurs
     if (core.isDebug()) {
-      await listCache(keys[0], httpClient, version)
+      await printCachesListForDiagnostics(keys[0], httpClient, version)
     }
     return null
   }
@@ -127,7 +127,7 @@ export async function getCacheEntry(
   return cacheResult
 }
 
-async function listCache(
+async function printCachesListForDiagnostics(
   key: string,
   httpClient: HttpClient,
   version: string
@@ -143,7 +143,7 @@ async function listCache(
       core.debug(
         `No matching cache found for cache key '${key}', version '${version} and scope ${process.env['GITHUB_REF']} but there are ${totalCount} existing version of the cache for this key. More info on versioning can be found here: https://github.com/actions/cache#cache-version \nOther versions are as follows:`
       )
-      for (const cacheEntry of cacheListResult?.artifactCaches || []) {
+      for (const cacheEntry of cacheListResult.artifactCaches || []) {
         core.debug(
           `Cache Key: ${cacheEntry?.cacheKey}, Cache Version: ${cacheEntry?.cacheVersion}, Cache Scope: ${cacheEntry?.scope}, Cache Created: ${cacheEntry?.creationTime}`
         )
--- a/packages/cache/src/internal/cacheHttpClient.ts
+++ b/packages/cache/src/internal/cacheHttpClient.ts
@@ -115,7 +115,7 @@ export async function getCacheEntry(
   const cacheDownloadUrl = cacheResult?.archiveLocation
   if (!cacheDownloadUrl) {
     // List cache for primary key only if cache miss occurs
-    await listCache(keys[0], httpClient, version)
+    await listCache(keys[0], httpClient, version, cacheResult?.scope)
     throw new Error('Cache not found.')
   }
   core.setSecret(cacheDownloadUrl)
@@ -128,7 +128,8 @@ export async function getCacheEntry(
 async function listCache(
   key: string,
   httpClient: HttpClient,
-  version: string
+  version: string,
+  scope?: string
 ): Promise<void> {
   const resource = `caches?key=${encodeURIComponent(key)}`
   const response = await retryTypedResponse('listCache', async () =>
@@ -139,7 +140,7 @@ async function listCache(
     const totalCount = cacheListResult?.totalCount
     if (totalCount && totalCount > 0) {
       core.info(
-        `Cache miss occurred on the cache key '${key}' and version '${version} but there is ${totalCount} existing version of the cache for this key. More info on versioning can be found here: https://github.com/actions/cache#cache-version`
+        `No matching cache found for cache key '${key}', version '${version} and scope ${scope} but there is ${totalCount} existing version of the cache for this key. More info on versioning can be found here: https://github.com/actions/cache#cache-version`
       )
       core.debug(`Other versions are as follows:`)
       cacheListResult?.artifactCaches?.forEach(cacheEntry => {
--- a/packages/cache/src/internal/cacheHttpClient.ts
+++ b/packages/cache/src/internal/cacheHttpClient.ts
@@ -105,6 +105,10 @@ export async function getCacheEntry(
     httpClient.getJson<ArtifactCacheEntry>(getCacheApiUrl(resource))
   )
   if (response.statusCode === 204) {
+    // List cache for primary key only if cache miss occurs
+    if (core.isDebug()) {
+      await listCache(keys[0], httpClient, version, response)
+    }
     return null
   }
   if (!isSuccessStatusCode(response.statusCode)) {
@@ -114,8 +118,6 @@ export async function getCacheEntry(
   const cacheResult = response.result
   const cacheDownloadUrl = cacheResult?.archiveLocation
   if (!cacheDownloadUrl) {
-    // List cache for primary key only if cache miss occurs
-    await listCache(keys[0], httpClient, version, cacheResult?.scope)
     throw new Error('Cache not found.')
   }
   core.setSecret(cacheDownloadUrl)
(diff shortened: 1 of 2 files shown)


packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -113,6 +114,21 @@ export async function getCacheEntry(
   const cacheResult = response.result
   const cacheDownloadUrl = cacheResult?.archiveLocation
   if (!cacheDownloadUrl) {
+    // List cache for primary key only if cache miss occurs
+    const resource = `caches?key=${encodeURIComponent(keys[0])}`
```
bishal-pdMSFT, line 118: It is debatable if we should use cache key or the last of the restore keys. If restore keys are specified by user then it means we did not find a matching cache even with restore key. So should we restrict to only cache key for showing candidate caches or use the least restrictive restore key.
One problem with least restrictive restore key would that the list of candidate caches can be reallly long and make the logs too verbose.

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -113,6 +114,21 @@ export async function getCacheEntry(
...
+    const response = await httpClient.getJson<ArtifactCacheList>(getCacheApiUrl(resource))
```
bishal-pdMSFT, line 119: So we are using AC endpoint directly and not GitHub REST API. I feel it should be fine but tagging @anuragc617 for his review.

packages/cache/src/internal/cacheHttpClient.ts
bishal-pdMSFT, line 118: Lets refactor these changes to a separate function for readability

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -113,6 +114,21 @@ export async function getCacheEntry(
...
+    if(response.statusCode === 204) {
```
bishal-pdMSFT, line 120: Handle null

```suggestion
    if(response?.statusCode === 204) {
```

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: I think we should stick to primary key as cache miss means miss on primary key. Even if cache is found for restore keys it is still said cache not hit

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: Other functions were doing the same so I thought should be fine. Little unsure about permissions need for such call to AC.

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: Makes sense

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -113,6 +114,21 @@ export async function getCacheEntry(
...
+      const cacheListResult = response.result
+      const totalCount = cacheListResult?.totalCount
+      if(totalCount && totalCount > 0) {
+        core.info(`Cache miss occurred on the cache key '${keys[0]}' and version '${version} but there is ${totalCount} existing version of the cache for this key. More info on versioning can be found here: https://github.com/actions/cache#cache-version`)
+        core.debug(`Other versions are as follows:`)
```
bishal-pdMSFT, line 125: I don't think count is much valuable info here and can be removed.

```suggestion
        core.info(`No matching cache found for cache key '${keys[0]}' with version '${version}. More info on cache matching can be found here: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key.`)
        core.debug(`Candidate caches which were tried:`)
```

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -113,6 +114,21 @@ export async function getCacheEntry(
...
   const cacheDownloadUrl = cacheResult?.archiveLocation
   if (!cacheDownloadUrl) {
+    // List cache for primary key only if cache miss occurs
+    const resource = `caches?key=${encodeURIComponent(keys[0])}`
+    const response = await httpClient.getJson<ArtifactCacheList>(getCacheApiUrl(resource))
+    if(response.statusCode === 204) {
+      const cacheListResult = response.result
+      const totalCount = cacheListResult?.totalCount
+      if(totalCount && totalCount > 0) {
+        core.info(`Cache miss occurred on the cache key '${keys[0]}' and version '${version} but there is ${totalCount} existing version of the cache for this key. More info on versioning can be found here: https://github.com/actions/cache#cache-version`)
```
bishal-pdMSFT, line 124: We should also mention the current scope if possible. Is it available in Context?

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -113,6 +114,8 @@ export async function getCacheEntry(
   const cacheResult = response.result
   const cacheDownloadUrl = cacheResult?.archiveLocation
   if (!cacheDownloadUrl) {
+    // List cache for primary key only if cache miss occurs
```
aparna-ravindra, line 117: This method exists in line 108 if the status code is 204. 
In which case will there be cacheResult without archiveLocation?

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: Makes sense. I added cache version docs to tell that version is the reason of mismatch. Candidate caches were not tried. They are other existing caches with same key but different versions

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -122,6 +125,32 @@ export async function getCacheEntry(
   return cacheResult
 }
 
+async function listCache(
+  key: string,
+  httpClient: HttpClient,
+  version: string
+): Promise<void> {
+  const resource = `caches?key=${encodeURIComponent(key)}`
```
aparna-ravindra, line 133: does the actions run time token have permissions to call this API?

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: Yes

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: From my testing no: [workflow run](https://github.com/Phantsure/test-repo/actions/runs/3444578591/jobs/5747330201). But is there a way to have those permissions? or is it not recommended from security perspective?

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: https://github.com/actions/toolkit/blob/main/packages/cache/src/cache.ts#L96
It checks after returning from here.

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: Removed "cache miss"

packages/cache/src/internal/cacheHttpClient.ts
aparna-ravindra: rephrasing the question:
If there is a cache miss, then the method exists in line 108 and this line is not reachable. Is that expected?

packages/cache/src/internal/cacheHttpClient.ts
bishal-pdMSFT: We will have to give explicit permission to the token and I think it may not be desirable to do so. Unfortunately we only have a "manage" permission and not more granular like "read".

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: In case of cache miss condition at line 107 is false as statusCode wouldn't be 204. Then it moves ahead to check if status is not success. Then if it is success but not 204 then it moves ahead to check more.

Phantsure: I have moved the logic up to line 108 when it returns null for cache miss

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -122,6 +127,31 @@ export async function getCacheEntry(
...
+  const response = await retryTypedResponse('listCache', async () =>
+    httpClient.getJson<ArtifactCacheList>(getCacheApiUrl(resource))
+  )
+  if (response.statusCode === 200) {
+    const cacheListResult = response.result
+    const totalCount = cacheListResult?.totalCount
+    if (totalCount && totalCount > 0) {
+      core.debug(
+        `No matching cache found for cache key '${key}', version '${version} and scope ${process.env['GITHUB_REF']} but there are ${totalCount} existing version of the cache for this key. More info on versioning can be found here: https://github.com/actions/cache#cache-version \nOther versions are as follows:`
```
bishal-pdMSFT, line 144: This is focusing only on `version`. mismatch may happen because of scope as well. Also the count seems to not an useful informatioj.
```suggestion
        `No matching cache found for cache key '${key}', version '${version} and scope ${process.env['GITHUB_REF']}. There exist one or more cache(s) with similar key but they have different version or scope. More info on cache matching can be found here: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key \nOther caches with similar key:`
```

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -104,6 +105,10 @@ export async function getCacheEntry(
     httpClient.getJson<ArtifactCacheEntry>(getCacheApiUrl(resource))
   )
   if (response.statusCode === 204) {
+    // List cache for primary key only if cache miss occurs
+    if (core.isDebug()) {
+      await listCache(keys[0], httpClient, version)
```
bishal-pdMSFT, line 110:
```suggestion
      await printCachesListForDiagnostics(keys[0], httpClient, version)
```

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -122,6 +127,31 @@ export async function getCacheEntry(
...
+    httpClient.getJson<ArtifactCacheList>(getCacheApiUrl(resource))
+  )
+  if (response.statusCode === 200) {
+    const cacheListResult = response.result
+    const totalCount = cacheListResult?.totalCount
+    if (totalCount && totalCount > 0) {
+      core.debug(
+        `No matching cache found for cache key '${key}', version '${version} and scope ${process.env['GITHUB_REF']}. There exist one or more cache(s) with similar key but they have different version or scope. See more info on cache matching here: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key \nOther caches with similar key:`
+      )
+      for (const cacheEntry of cacheListResult?.artifactCaches || []) {
```
bishal-pdMSFT, line 146: check for null is not needed as L142 already checks that there are more than 0 entries.
```suggestion
      for (const cacheEntry of cacheListResult.artifactCaches || []) {
```

bishal-pdMSFT (approved):
Looks good apart from few minor comments.

bishal-pdMSFT (approved):

PR #1237, opened 2022-11-16
Bsdtar + zstd fallback to prevent zstd hanging issue on windows

pdotl:
This PR implements the `bsdtar + zstd` fallback for windows to prevent the `bsdtar + zstd` hanging on windows issue highlighted in [#301](https://github.com/actions/cache/issues/301)

## Context and Motivation

`bsdtar` on windows runners tends to hang when called with `--use-compress-program zstd`. During testing we found the issue can be circumvented by calling `bsdtar` and `zstd` as separate processes such as `tar [OPTIONS] && zstd [OPTIONS]`. If `tar` fails then `zstd` won't run.

For compression, command would be similar to `tar [OPTIONS] && zstd [OPTIONS]` and for decompression, command would be similar to `zstd -d [OPTIONS] && tar [OPTIONS]`.

This PR implements this approach as a fallback method since from #1239 `gnutar` will be default on windows where this problem does not exist.

## Requesting Feedback

The ordering of arguments while forming the `tar` command that executes the archiving and compression is what has changed, so the change has larger exposure than required. Feel free to suggest better ways to implement this fallback where the impact can be minimized.

## Relevant Documentation
- Parent Public issue: https://github.com/actions/cache/issues/984


packages/cache/src/internal/tar.ts
```diff
@@ -3,18 +3,21 @@ import * as io from '@actions/io'
...
+// Function also mutates the args array. For non-mutation call with passing an empty array.
 async function getTarPath(args: string[]): Promise<string> {
   switch (process.platform) {
     case 'win32': {
       const gnuTar = await utils.getGnuTarPathOnWindows()
       const systemTar = `${process.env['windir']}\\System32\\tar.exe`
       if (gnuTar) {
         // Use GNUtar as default on windows
-        args.push('--force-local')
+        if (args.length > 0) {
```
Phantsure, line 18: Do we need this?

packages/cache/src/internal/tar.ts
```diff
@@ -49,19 +54,28 @@ function getWorkingDirectory(): string {
...
   // -d: Decompress.
   // unzstd is equivalent to 'zstd -d'
   // --long=#: Enables long distance matching with # bits. Maximum is 30 (1GB) on 32-bit OS and 31 (2GB) on 64-bit.
   // Using 30 here because we also support 32-bit self-hosted runners.
+  const tarPath = await getTarPath([])
+  const BSD_TAR_ZSTD = IS_WINDOWS && tarPath === SystemTarPathOnWindows
   switch (compressionMethod) {
     case CompressionMethod.Zstd:
+      if (BSD_TAR_ZSTD) {
+        return ['-a'] // auto-detect compression
```
Phantsure, line 67: This is done to make it future proof?

packages/cache/src/internal/tar.ts
```diff
@@ -49,19 +54,28 @@ function getWorkingDirectory(): string {
...
+      }
       return [
         '--use-compress-program',
-        IS_WINDOWS ? 'zstd -d --long=30' : 'unzstd --long=30'
+        IS_WINDOWS ? 'zstd -d --long=30' : 'unzstd --long=30',
+        '-tf'
```
Phantsure, line 72: `-t` is used for?

packages/cache/src/internal/tar.ts
Phantsure: Do we not have to pipe here?

pdotl: Future proof as well as simpler to reason about as the ordering of args might get brittle with different platforms.

packages/cache/src/internal/tar.ts
pdotl: -t is used for listing the cache contents. I had to move it from the listTar function to getCompressionProgram due to ordering

Phantsure: Okay. Makes sense

packages/cache/src/internal/tar.ts
Phantsure: Won't we still need pipe?

pdotl: We should not need to. Unless this also exhibits the hanging issue.

packages/cache/src/internal/tar.ts
Phantsure: Okay

packages/cache/src/internal/tar.ts
Phantsure: Have to check it

packages/cache/__tests__/tar.test.ts
```diff
@@ -149,17 +149,17 @@ test('zstd create tar', async () => {
...
-      '-cf',
-      IS_WINDOWS ? CacheFilename.Zstd.replace(/\\/g, '/') : CacheFilename.Zstd,
       '--exclude',
       IS_WINDOWS ? CacheFilename.Zstd.replace(/\\/g, '/') : CacheFilename.Zstd,
       '-P',
       '-C',
       IS_WINDOWS ? workspace?.replace(/\\/g, '/') : workspace,
       '--files-from',
-      'manifest.txt'
+      'manifest.txt',
```
Phantsure, line 158: Why the reorder?

pdotl: Tests were failing so the order needs be the same here as well.

packages/cache/src/internal/tar.ts
```diff
@@ -117,31 +132,43 @@ export async function createTar(
...
     '--exclude',
     cacheFileName.replace(new RegExp(`\\${path.sep}`, 'g'), '/'),
     '-P',
     '-C',
     workingDirectory.replace(new RegExp(`\\${path.sep}`, 'g'), '/'),
     '--files-from',
-    manifestFilename
+    manifestFilename,
+    ...(await getCompressionProgram()),
+    cacheFileName.replace(new RegExp(`\\${path.sep}`, 'g'), '/')
```
bishal-pdMSFT, line 171: Does moving this args order have any negative effect on gnutar case or bsdtar on macos case? Essentially it would be good to idea to test all these existing scenarios for regression.

packages/cache/src/internal/tar.ts
```diff
@@ -117,31 +132,43 @@ export async function createTar(
   // --long=#: Enables long distance matching with # bits. Maximum is 30 (1GB) on 32-bit OS and 31 (2GB) on 64-bit.
   // Using 30 here because we also support 32-bit self-hosted runners.
   // Long range mode is added to zstd in v1.3.2 release, so we will not use --long in older version of zstd.
-  function getCompressionProgram(): string[] {
+  async function getCompressionProgram(): Promise<string[]> {
+    const tarPath = await getTarPath([])
+    const BSD_TAR_ZSTD = IS_WINDOWS && tarPath === SystemTarPathOnWindows
```
bishal-pdMSFT, line 137: Naming it `BSD_TAR_WINDOWS` may be better as it coveys that this applies only for windows and uses bsdtar. zstd is already evident from switch case below.

packages/cache/src/internal/tar.ts
```diff
@@ -49,18 +54,28 @@ function getWorkingDirectory(): string {
...
   // -d: Decompress.
   // unzstd is equivalent to 'zstd -d'
   // --long=#: Enables long distance matching with # bits. Maximum is 30 (1GB) on 32-bit OS and 31 (2GB) on 64-bit.
   // Using 30 here because we also support 32-bit self-hosted runners.
+  const tarPath = await getTarPath([])
+  const BSD_TAR_ZSTD = IS_WINDOWS && tarPath === SystemTarPathOnWindows
   switch (compressionMethod) {
     case CompressionMethod.Zstd:
+      if (BSD_TAR_ZSTD) {
+        return ['-a'] // auto-detect compression
```
bishal-pdMSFT, line 69: Will auto-detection take care of `--long=30` part?

packages/cache/src/internal/tar.ts
Phantsure: I think no. `-a` seems to be useful for compression not decompression. I got the below error while running it:
`Option -a is not permitted in mode -x`. Workflow run: https://github.com/bbq-beets/test-zstd/actions/runs/3512158872/jobs/5883582602
freebsd docs also stats the same: https://www.freebsd.org/cgi/man.cgi?query=bsdtar&sektion=1&format=htm

packages/cache/src/internal/tar.ts
Phantsure: It seems to be used for compression where information is taken from tar file extension. So I think it can't take care of that

packages/cache/src/internal/tar.ts
Phantsure: Makes sense

packages/cache/src/internal/tar.ts
Phantsure: Generally it shouldn't affect. In windows case due to pipe symbol it is needed. I would take a look at it

packages/cache/src/internal/tar.ts
```diff
@@ -49,18 +54,28 @@ function getWorkingDirectory(): string {
...
 // Common function for extractTar and listTar to get the compression method
-function getCompressionProgram(compressionMethod: CompressionMethod): string[] {
+async function getCompressionProgram(
+  compressionMethod: CompressionMethod
+): Promise<string[]> {
   // -d: Decompress.
   // unzstd is equivalent to 'zstd -d'
   // --long=#: Enables long distance matching with # bits. Maximum is 30 (1GB) on 32-bit OS and 31 (2GB) on 64-bit.
   // Using 30 here because we also support 32-bit self-hosted runners.
+  const tarPath = await getTarPath([])
```
bishal-pdMSFT, line 64: getTarPath is only used for getting to know if `SystemTarPathOnWindows` is used or not. This can be simplified so that we do not need to fit `args` array.
For example either:
1. Create a separate function to just get tar path
2. Or move settings the args to `getCompressionProgram` where anyways we are setting more arguments

bishal-pdMSFT:
@Phantsure @pdotl overall I feel that the logic to find right command for various combinations is very spread-out and hence difficult to maintain. Should we think about simplifying it (in this or a separate PR)?
1. Decide tar and its path
2. For that tar decide compression 
3. A map for create/extract command which has three pivots - tar, compression and OS.

packages/cache/src/internal/tar.ts
Phantsure: args set in getTarPath are only the ones related to particular os which also helps in deciding which tar to use.

Phantsure:
I agree with you how the logic is spread at different places. 

> A map for create/extract command which has three pivots - tar, compression and OS.

@pdotl also pointed this out. Let me try to work on reducing the number of pivots.

packages/cache/src/internal/tar.ts
bishal-pdMSFT: In that case we can break into two function - one to get tar path and another to to get archive command.

packages/cache/src/internal/tar.ts
```diff
@@ -3,44 +3,148 @@ import * as io from '@actions/io'
...
+  const compressionArgs =
+    type !== 'create'
+      ? await getDecompressionProgram(tarPath, compressionMethod, archivePath)
+      : await getCompressionProgram(tarPath, compressionMethod)
+  const BSD_TAR_ZSTD =
+    tarPath.type === ArchiveToolType.BSD &&
+    compressionMethod !== CompressionMethod.Gzip &&
+    IS_WINDOWS
+  if (BSD_TAR_ZSTD && type !== 'create') {
+    return [...compressionArgs, ...tarArgs].join(' ')
```
bishal-pdMSFT, line 145: Is this specific needed only for BSD on decompression?

packages/cache/src/internal/tar.ts
```diff
@@ -49,19 +153,84 @@ function getWorkingDirectory(): string {
...
+async function getDecompressionProgram(
+  tarPath: ArchiveTool,
+  compressionMethod: CompressionMethod,
+  archivePath: string
+): Promise<string[]> {
   // -d: Decompress.
   // unzstd is equivalent to 'zstd -d'
   // --long=#: Enables long distance matching with # bits. Maximum is 30 (1GB) on 32-bit OS and 31 (2GB) on 64-bit.
   // Using 30 here because we also support 32-bit self-hosted runners.
+  const tarFile = 'cache.tar'
```
bishal-pdMSFT, line 165: this looks same across different functions. Can be moved to constants?

packages/cache/src/internal/tar.ts
```diff
@@ -3,44 +3,148 @@ import * as io from '@actions/io'
...
+
+// Return arguments for tar as per tarPath, compressionMethod, method type and os
+async function getTarArgs(
+  tarPath: ArchiveTool,
+  compressionMethod: CompressionMethod,
+  type: string,
+  archivePath = ''
+): Promise<string[]> {
+  const args = [tarPath.path]
+  const manifestFilename = 'manifest.txt'
```
bishal-pdMSFT, line 58: this looks same across different functions. Can be moved to constants?

bishal-pdMSFT (approved):

packages/cache/src/internal/tar.ts
Phantsure: Only for BSDtar on windows with zstd. So very specific

packages/cache/src/internal/tar.ts
Phantsure: Done

packages/cache/src/internal/tar.ts
Phantsure: Done

PR #1249, opened 2022-12-06
Add fallback to gzip compression if cache not found

Phantsure:
This would make sure that caches done previously using `gzip` are restorable but any new caches or restore done using restore keys would be saved using `zstd` on windows if available.
### Changes made:
When cache not found on windows and compression method is not `gzip` then search again with `gzip`. If cache found with `gzip` compression cache version then continue as usual else throw error.

### DIFF
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -106,9 +106,8 @@ export async function restoreCache(
         process.platform === 'win32' &&
         compressionMethod !== CompressionMethod.Gzip
       ) {
-        // On windows, we will try to download the cache entry with the same key
-        // but with different compression method. This is to support the old cache entry created
-        // by the old version of the cache action.
+        // This is to support the old cache entry created
+        // by the old version of the cache action on windows.
         compressionMethod = CompressionMethod.Gzip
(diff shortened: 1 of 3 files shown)


packages/cache/src/cache.ts
```diff
@@ -85,17 +87,38 @@ export async function restoreCache(
...
+        process.platform === 'win32' &&
+        compressionMethod !== CompressionMethod.Gzip
+      ) {
+        // On windows, we will try to download the cache entry with the same key
+        // but with different compression method. This is to support the old cache entry created
+        // by the old version of the cache action.
+        compressionMethod = CompressionMethod.Gzip
+        cacheEntry = await cacheHttpClient.getCacheEntry(keys, paths, {
+          compressionMethod
+        })
```
bishal-pdMSFT, line 115: This can be moved to a well named function for the specific case for ease of understanding, and not take too much space in the overall function.

```suggestion
      cacheEntry = await getCacheEntryOnWindowsForFallbackCompression(keys, paths, compressionMethod);
```

bishal-pdMSFT:
Please add tests changes as well

bishal-pdMSFT (approved):

Issue #461, opened 2020-05-16
@actions/exec doesn't support multiple commands


JasonEtco:
**Describe the bug**

Hello! I ran this script through `@actions/exec`:

```js
await exec.exec('npm ci && npm run build --if-present')
```

It was expanded and run:

![](https://user-images.githubusercontent.com/10660468/82132143-8df37a80-97aa-11ea-8b6c-8f019f379923.png)

So it looks like the first `npm` command was prefixed with `/usr/local/bin`, but the second wasn't. It also didn't throw an error, which made this difficult to debug. What's the right way to 

cc https://github.com/JasonEtco/build-and-tag-action/issues/4

**To Reproduce**

Create an action:

**package.json**:

```json
{
  "scripts": {
    "build": "echo \"Hello!\""
  }
}
```

**The action:**

```js
const exec = require('@actions/exec')

async function main () {
  await exec.exec('npm ci && npm run build')
}

main()
```

Observe that the `build` command isn't properly run.

**Expected behavior**

I'd expect the above command to "just work" - but if not, it should throw an error instead of fail silently.

eine:
You can use [which](https://github.com/actions/toolkit/tree/master/packages/io#which) to get the full path to npm.

JasonEtco:
I would expect `@actions/exec` to do that under the hood, not just for the first command but for all commands chained with `&&`s.

eine:
@JasonEtco, AFAIK `exec` is not a bash interpreter.

NorseGaud:
I work around this (since I know my action should only be used on machines with bash) using
```
    var commandsToRun = commands.replace( /&& (\\\n)?/g, '\n' ).split("\n");
    var commandArrayLength = commandsToRun.length;
    for (var i = 0; i < commandArrayLength; i++) {
        await exec.exec(commandsToRun[i], [], options);
    }
```

Though, there are still limitations like `echo \"Test2\" > /tmp/test` working. 

I think the main thing to communicate to users of an action (using toolkit/exec) is that it's not a bash interpreter. Finding a way to force bash interpretation is my next play.

NorseGaud:
Sure enough, this works great:

```
    var commandsToRun = commands.replace( /&& (\\\n)?/g, '\n' ).split("\n");
    var commandArrayLength = commandsToRun.length;
    for (var i = 0; i < commandArrayLength; i++) {
        await exec.exec('bash', ['-c', commandsToRun[i]], options);
    }
```

[...]

PR #1260, opened 2022-12-09
Fix known issues for cache

Phantsure:
Fix known issue with gzip fallback and bsd tar + zstd combination. 
- gzip fallback was failing as getCacheEntry return null when cache not found and throw no error.
- bsd tar + zstd was failing as `exec.exec` doesn't support `&&` so passed commands through `cmd` for that case only as suggested [here](https://github.com/actions/toolkit/issues/461)

### DIFF
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -91,18 +91,13 @@ export async function restoreCache(
   let compressionMethod = await utils.getCompressionMethod()
   let archivePath = ''
   try {
-    try {
-      console.log('before first get cache entry')
-      // path are needed to compute version
-      cacheEntry = await cacheHttpClient.getCacheEntry(keys, paths, {
-        compressionMethod
-      })
-      console.log('after first get cache entry')
-      console.log(cacheEntry)
-    } catch (error) {
+    // path are needed to compute version
+    cacheEntry = await cacheHttpClient.getCacheEntry(keys, paths, {
+      compressionMethod
+    })
+    if (!cacheEntry?.archiveLocation) {
       // This is to support the old cache entry created
       // by the old version of the cache action on windows.
-      console.log('in first catch block')
       if (
         process.platform === 'win32' &&
         compressionMethod !== CompressionMethod.Gzip
@@ -114,19 +109,15 @@ export async function restoreCache(
         cacheEntry = await cacheHttpClient.getCacheEntry(keys, paths, {
           compressionMethod
         })
-        console.log(cacheEntry)
         if (!cacheEntry?.archiveLocation) {
-          throw error
+          return undefined
         }
       } else {
-        throw error
+        // Cache not found
+        return undefined
       }
     }
 
-    if (!cacheEntry?.archiveLocation) {
-      // Cache not found
-      return undefined
-    }
     archivePath = path.join(
       await utils.createTempDirectory(),
       utils.getCacheFileName(compressionMethod)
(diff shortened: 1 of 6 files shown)


packages/cache/src/cache.ts
```diff
@@ -91,12 +91,11 @@ export async function restoreCache(
   let compressionMethod = await utils.getCompressionMethod()
   let archivePath = ''
   try {
```
bishal-pdMSFT, line 93: Don't we still need to have catch for the try, so as to continue the old behavior if `getCacheEntry` throws?

packages/cache/src/internal/tar.ts
```diff
@@ -142,11 +144,18 @@ async function getArgs(
...
   if (BSD_TAR_ZSTD && type !== 'create') {
-    return [...compressionArgs, ...tarArgs].join(' ')
+    args = [...compressionArgs, ...tarArgs].join(' ')
   } else {
-    return [...tarArgs, ...compressionArgs].join(' ')
+    args = [...tarArgs, ...compressionArgs].join(' ')
+  }
+
+  if (BSD_TAR_ZSTD) {
+    args = ['cmd /c "', args, '"'].join(' ')
```
bishal-pdMSFT, line 155: Instead of using cmd, try if it is possible to run two exec commands as there is no pipe between `tar` and `zstd`.

packages/cache/src/cache.ts
Phantsure: We don't need to catch again. It is caught in below catch statement. I have updated the comments for better understanding in future.

packages/cache/src/internal/tar.ts
```diff
@@ -142,11 +147,18 @@ async function getArgs(
...
-    return [...compressionArgs, ...tarArgs].join(' ')
+    args = [[...compressionArgs].join(' '), [...tarArgs].join(' ')]
   } else {
-    return [...tarArgs, ...compressionArgs].join(' ')
+    args = [[...tarArgs].join(' '), [...compressionArgs].join(' ')]
   }
+
+  if (BSD_TAR_ZSTD) {
+    return args
+  }
```
bishal-pdMSFT, line 159: Instead of this, just return directly from L152 and L154

packages/cache/src/internal/tar.ts
```diff
@@ -183,17 +194,17 @@ async function getDecompressionProgram(
     case CompressionMethod.ZstdWithoutLong:
       return BSD_TAR_ZSTD
         ? [
-            'zstd -d -o',
+            'zstd -d --force -o',
```
bishal-pdMSFT, line 197: Is this intentional? Why?

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -113,6 +114,7 @@ export async function getCacheEntry(
   const cacheResult = response.result
   const cacheDownloadUrl = cacheResult?.archiveLocation
   if (!cacheDownloadUrl) {
+    // Cache achiveLocation not found
```
bishal-pdMSFT, line 117:
```suggestion
    // Cache achiveLocation not found. This should never happen, and hence bail out.
```

bishal-pdMSFT (approved):

packages/cache/src/internal/tar.ts
Phantsure: It returns only in case of bsd tar. Else it will go down

packages/cache/src/internal/tar.ts
Phantsure: removed

PR #1278, opened 2022-12-20
Migrate dev environment and workflows to node16 

fhammerl:
To fix critical audit vulnerabilities, we needed to upgrade (dev) dependencies to versions that are no longer compatible with our current workflows using node12.

Node12 and node16 are the versions used by the runner. With Node12 being deprecated, we've targeted node16 for this project as well. It makes sense to keep in sync with the runner, as most eventual applications of `actions/tookit`'s packages end up being executed by the runner.

Also contains numerous fixes to satisfy `audit-all`

### DIFF
--- a/.github/workflows/artifact-tests.yml
+++ b/.github/workflows/artifact-tests.yml
@@ -24,10 +24,10 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
-    - name: Set Node.js 12.x
+    - name: Set Node.js 16.x
       uses: actions/setup-node@v1
       with:
-        node-version: 12.x
+        node-version: 16.x
 
     # In order to upload & download artifacts from a shell script, certain env variables need to be set that are only available in the
     # node context. This runs a local action that gets and sets the necessary env variables that are needed
--- a/.github/workflows/artifact-tests.yml
+++ b/.github/workflows/artifact-tests.yml
@@ -55,16 +55,16 @@ jobs:
     - name: Create files that will be uploaded
       run: |
         mkdir artifact-path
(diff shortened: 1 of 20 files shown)


packages/artifact/src/internal/download-http-client.ts
```diff
@@ -219,6 +219,13 @@ export class DownloadHttpClient {
       fileDownloadPath: string
     ): Promise<void> => {
       destinationStream.close()
+      // await until file is created at downloadpath; node15 and up fs.createWriteStream had not created a file yet
```
fhammerl, line 222: The single code change we need in node16: awaiting `destinationStream.on('close',` ensures that the file has been created so we can `rmFile` delete it on line 229.

.github/workflows/artifact-tests.yml
```diff
@@ -24,10 +24,10 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
-    - name: Set Node.js 12.x
+    - name: Set Node.js 16.x
       uses: actions/setup-node@v1
```
Phantsure, line 28: setup-node can also be updated to v3?

rentziass (approved):

packages/artifact/src/internal/download-http-client.ts
donatocayurin: ❤️🙏

.github/workflows/artifact-tests.yml
donatocayurin: ❤️🙏

Maison360:
k@t@m

PR #1279, opened 2022-12-21
Address bugbash issues

Phantsure:
There were few issues which came up during internal bugbash:
1. With debug logs on restore was failing saying `zstd: cache.tar already exists; not overwritten:`. 
Possible solutions:
- Add `--force` to zstd command. Docs say that they overwrite the file with this.
- Delete `cache.tar` file after every list, create and extract operation
**Final solution:** Add --force to zstd command as problem is seen when `list` runs which happen in debug mode thus justifying extra time taken in deleting and creating `cache.tar` file using --force.
3. Change debug log to info log when falling back to gzip for restoration of cache on windows.

### DIFF
--- a/.github/workflows/cache-windows-test.yml
+++ b/.github/workflows/cache-windows-test.yml
@@ -78,7 +78,6 @@ jobs:
       run: |
         rm -rf test-cache
         rm -rf ~/test-cache
-        rm -f cache.tar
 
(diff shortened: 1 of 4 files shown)


pdotl (approved):
LGTM. It would be great to add some context in description as to why you chose the `--force` method instead of the other suggested ones.

PR #1253, opened 2022-12-07
feat: allow to tar files from root

blaggacao:

PR #1281, opened 2022-12-22
Cache package release for compression change in windows

Phantsure:
This branch has changes related to compression on windows. Add
1. Change default archiver to GNU tar and compression algorithm to zstd on windows
2. Allowed fallback to BSD tar and zstd in case GNU tar is not present in path or at default location on windows.
3. Allowed fallback to old caching archiving-compression algorithm, i.e. BSD tar and gzip if old cache on windows is found.


packages/cache/src/internal/tar.ts
```diff
@@ -3,153 +3,292 @@ import * as io from '@actions/io'
...
+        '-tf',
+        BSD_TAR_ZSTD
+          ? tarFile
+          : archivePath.replace(new RegExp(`\\${path.sep}`, 'g'), '/'),
+        '-P'
+      )
+      break
+  }
+
+  // Platform specific args
```
pdotl, line 112: Appreciate how you have handled platform specific args in a separate method. Much cleaner and readable.

packages/cache/src/internal/tar.ts
```diff
@@ -3,153 +3,292 @@ import * as io from '@actions/io'
...
+
+  const tarPath = await getTarPath()
+  const tarArgs = await getTarArgs(
+    tarPath,
+    compressionMethod,
+    type,
+    archivePath
+  )
+  const compressionArgs =
+    type !== 'create'
```
pdotl, line 143: Nit⛏️ :Would using a case statement instead of ternary operator make this more readable?

packages/cache/src/internal/tar.ts
```diff
@@ -3,153 +3,292 @@ import * as io from '@actions/io'
...
+      ? await getDecompressionProgram(tarPath, compressionMethod, archivePath)
+      : await getCompressionProgram(tarPath, compressionMethod)
+  const BSD_TAR_ZSTD =
+    tarPath.type === ArchiveToolType.BSD &&
+    compressionMethod !== CompressionMethod.Gzip &&
```
pdotl, line 148: Nit⛏️ : Same as above. Since the variable name says zstd but we are actually checking if it not gzip.

packages/cache/src/internal/tar.ts
```diff
@@ -3,153 +3,292 @@ import * as io from '@actions/io'
...
+    IS_WINDOWS
+
+  if (BSD_TAR_ZSTD && type !== 'create') {
```
pdotl, line 151: Nit⛏️ Same as above. Would case statements make more sense?

packages/cache/src/internal/tar.ts
```diff
@@ -3,153 +3,292 @@ import * as io from '@actions/io'
...
+        ? [
+            'zstd -d --long=30 --force -o',
+            TarFilename,
+            archivePath.replace(new RegExp(`\\${path.sep}`, 'g'), '/')
+          ]
+        : [
+            '--use-compress-program',
+            IS_WINDOWS ? '"zstd -d --long=30"' : 'unzstd --long=30'
+          ]
+    case CompressionMethod.ZstdWithoutLong:
```
pdotl, line 194: We can try to set `--long=30` with string intrapolation to see if we can reduce duplicity in this method. From a view it is the only difference between two cases.

packages/cache/src/internal/tar.ts
```diff
@@ -3,153 +3,292 @@ import * as io from '@actions/io'
...
-      ]
+      return BSD_TAR_ZSTD
+        ? [
+            'zstd -T0 --long=30 --force -o',
+            cacheFileName.replace(new RegExp(`\\${path.sep}`, 'g'), '/'),
+            TarFilename
+          ]
+        : [
+            '--use-compress-program',
+            IS_WINDOWS ? '"zstd -T0 --long=30"' : 'zstdmt --long=30'
```
pdotl, line 232: Same as above.

pdotl (changes requested):

packages/cache/src/internal/tar.ts
```diff
@@ -3,153 +3,292 @@ import * as io from '@actions/io'
...
+  archivePath = ''
+): Promise<string[]> {
+  const args = [`"${tarPath.path}"`]
+  const cacheFileName = utils.getCacheFileName(compressionMethod)
+  const tarFile = 'cache.tar'
+  const workingDirectory = getWorkingDirectory()
+  // Speficic args for BSD tar on windows for workaround
+  const BSD_TAR_ZSTD =
+    tarPath.type === ArchiveToolType.BSD &&
+    compressionMethod !== CompressionMethod.Gzip &&
```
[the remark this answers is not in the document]
Phantsure: There are two zstd variants, with and without long so did with gzip

packages/cache/src/internal/tar.ts
Phantsure: Thanks 🙇🏼

packages/cache/src/internal/tar.ts
Phantsure: Answered above

packages/cache/src/internal/tar.ts
Phantsure: Why so?

packages/cache/src/internal/tar.ts
Phantsure: I wanted to keep entire zstd command in one place because then we would have `"` end double quotes in another string.

packages/cache/src/internal/tar.ts
```diff
@@ -3,153 +3,292 @@ import * as io from '@actions/io'
...
+    case CompressionMethod.Zstd:
+      return BSD_TAR_ZSTD
+        ? [
+            'zstd -d --long=30 --force -o',
+            TarFilename,
+            archivePath.replace(new RegExp(`\\${path.sep}`, 'g'), '/')
+          ]
+        : [
+            '--use-compress-program',
+            IS_WINDOWS ? '"zstd -d --long=30"' : 'unzstd --long=30'
```
Phantsure, line 192:
```suggestion
        : [
            '--use-compress-program',
            IS_WINDOWS ? '"zstd -d' : '"unzstd',
            '--long=30"'
```
You are suggesting something like this?

packages/cache/src/internal/tar.ts
Phantsure: This is how you are suggesting? https://github.com/actions/toolkit/pull/1281/files/c76b5b3a03849f2d13e9311d7541d4b56072cd5a#r1055295832

packages/cache/src/internal/tar.ts
Phantsure: In this case no

packages/cache/src/internal/tar.ts
Phantsure: It is just single if true then this other wise this case. Had there been multiple like in `list`, `create` and `extract` then it would make more sense then

packages/cache/src/internal/tar.ts
pdotl: This with introducing a IS_ZSTD_LONG ternary operator which can expand to add `--long` accordingly?

Phantsure: Yes

Phantsure: Does this feel okay with quotes in another line?

pdotl (approved):
LGTM

blaggacao:
Does this PR mean that the following cache keys:

```yaml
    path: |
      /nix
```

Are now restored to `GITHUB_WORKSPACE` and we need to post-process to move it into the right place?

See also: https://github.com/actions/toolkit/pull/1253

PR #1286, opened 2022-12-23
Add cache restore `lookupOnly` option

cdce8p:
This PR adds a `lookupOnly` option to the cache restore function.
It can be used to check if a cache exists without the need to download it.

Especially useful if e.g. dependency install and actual tests are split across multiple jobs. If the cache restore step for the dependency install has an exact hit, the remaining steps in the job can be skipped. Similarly if the cache is created from scratch and it's just necessary to know if a cache entry already exists.

Would help address
https://github.com/actions/cache/issues/901
https://github.com/actions/cache/issues/831
https://github.com/actions/cache/discussions/1020#discussioncomment-4332778
https://github.com/actions/cache/discussions/1020#discussioncomment-4341251

Alternative to #1207

CC: @kotewar

### DIFF
--- a/packages/cache/RELEASES.md
+++ b/packages/cache/RELEASES.md
@@ -118,3 +118,6 @@
 
 ### 3.1.4
 - Fix zstd not being used due to `zstd --version` output change in zstd 1.5.4 release. See [#1353](https://github.com/actions/toolkit/pull/1353).
+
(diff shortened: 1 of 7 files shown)


kotewar (approved):
Looks good to me.

PR #1291, opened 2022-12-29
Cache package release for compression change in windows with symlink fix

Phantsure:
- Adding changes related to compression and cross os functionality back on windows. 
- Adding symlink fix for Linux cache to restore on windows. Check https://github.com/actions/cache/issues/1010 for more details regarding issue.
- Making cross-os functionality as opt-in instead of default

Please check commits after first commit to understand the changes done
Original PR: actions/toolkit#1281


packages/cache/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/cache",
-  "version": "3.1.1",
```
bishal-pdMSFT, line 3: Version should bump up and not down

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -73,13 +73,18 @@ function createHttpClient(): HttpClient {
...
-      : [compressionMethod]
-  )
+  const components = paths
+    .concat(
+      !compressionMethod || compressionMethod === CompressionMethod.Gzip
+        ? []
+        : [compressionMethod]
+    )
+    .concat(
+      process.platform !== 'win32' || crossOsEnabled ? [] : ['windows-only']
```
bishal-pdMSFT, line 86: Clever change but I am lacking to understand if this is a sufficient change? If I am not wrong, we will end up using gnu tar even when `crossOsEnabled` set to false.

Phantsure: Yes. So GNU tar on windows to windows has not shown any problems. The problems encountered is mostly on cross-os. 
If we go for bsd tar then old problems would still persist. I feel we should still fix those issues by using GNU tar

packages/cache/package.json
Phantsure: Yes.

packages/cache/src/internal/cacheHttpClient.ts
bishal-pdMSFT: This means we would go to gnu tar + gzip, as zstd is not present on hosted windows. Isn't it? Have we tested that?

bishal-pdMSFT: correcttion. zstd is installed. Earlier it was explicitly ignored.

Phantsure: For self hosted windows runners, yes. And we have tested that as when we try to restore old caches done using gzip, we use gnutar + gzip combination.

bishal-pdMSFT, line 86: This change also means that in case of windows and cross os disabled, the existing caches would get invalidated due to version mismatch. Is that correct?

Phantsure: > correcttion. zstd is installed. Earlier it was explicitly ignored.

Yes, for github hosted runner we have zstd

bishal-pdMSFT: I agree with this approach.

bishal-pdMSFT: @pdotl eyes from you please!

Phantsure: Yes. There would be cache invalidation

Phantsure: And if that is acceptable then probably we can remove gzip fallback to restore old caches. Can update that to account for version difference but seems like there would be too many if-else

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -73,13 +73,18 @@ function createHttpClient(): HttpClient {
...
-  const components = paths.concat(
-    !compressionMethod || compressionMethod === CompressionMethod.Gzip
-      ? []
-      : [compressionMethod]
-  )
+  const components = paths
+    .concat(
+      !compressionMethod || compressionMethod === CompressionMethod.Gzip
+        ? []
+        : [compressionMethod]
```
bishal-pdMSFT, line 83: Should we remove this as we are not doing any special casing for gzip?

bishal-pdMSFT: It will cause cache invalidation, but that will get masked by invalidation due to `windows-only`

packages/cache/src/cache.ts
```diff
@@ -61,13 +61,15 @@ export function isFeatureAvailable(): boolean {
...
  * @param restoreKeys an optional ordered list of keys to use for restoring the cache if no cache hit occurred for key
+ * @param crossOsEnabled an optional boolean enabled to restore on windows any cache created on any platform
  * @param downloadOptions cache download options
  * @returns string returns the key for the cache hit, otherwise returns undefined
  */
 export async function restoreCache(
   paths: string[],
   primaryKey: string,
   restoreKeys?: string[],
+  crossOsEnabled = false,
```
bishal-pdMSFT: Thinking about the name - is `crossOsCompatibility` better name? "Enabled" does convey that this is kind of a feature flag (which it is!) but from user perspective compatibility is what is important and how it is rolled out.
Thoughts?

packages/cache/src/internal/cacheHttpClient.ts
bishal-pdMSFT: I was thinking the opposite in another comment. Keep the fallback but remove the version differentiation for gzip.

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: I think this is there so that cache created by gzip is only restorable by gzip. If self-hosted runners don't have zstd then it falls back to gzip. 
We can remove this if we callout that having zstd in path is a requirement for cache to work

Phantsure: Then in that case we can remove gzip from our code. And that seems like a big change

packages/cache/src/cache.ts
Phantsure: `crossOsEnabled` feels better, as compatibility is something largely user has to think about when enabling this as to if the cache contents are compatible or not

packages/cache/src/internal/cacheHttpClient.ts
bishal-pdMSFT: Won't the fallback logic take care of this?

Phantsure: We need different versions of cache for different compression algorithms as same compression algo should be used to compress(save) and decompress(restore)

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: We can't remove version differentiation because:
> We need different versions of cache for different compression algorithms as same compression algo should be used to compress(save) and decompress(restore)

packages/cache/src/internal/cacheHttpClient.ts
bishal-pdMSFT: @Phantsure yes, and that is happening as we are putting `[compressionMethod]` in the version (L83)

Phantsure: @bishal-pdMSFT Yes, but you were asking if we should remove it. Right? We need to keep some component of compression algo in version

packages/cache/src/internal/cacheHttpClient.ts
pdotl: I agree we cannot remove version differentiation here and also that we can remove the fallback since the caches are getting invalidated anyways. However, I would just suggest we update the logic/comment to go together for consistency.

pdotl: I also agree that we should use gnutar for windows to windows. Thanks for the explicit callout here @bishal-pdMSFT.

packages/cache/src/internal/cacheHttpClient.ts
pdotl: So just to understand, if we remove this statement, cache invalidation will happen because the newer versions will include `gzip` in their version hash and that won't match for older gzip caches. So even in this case, shouldn't gzip caches only be restorable by using gzip. Previously they were identified by not having `gzip` in version but moving forward will have `gzip` in version. I think this is what @bishal-pdMSFT is referring to.

pdotl (commented):
Added comments

packages/cache/src/cache.ts
pdotl: My two cents here, I agree with not using compatibility since it elludes to making the contents compatible. While `crossOsEnabled` does work, it feels unconventional since we prefer using present/present-perfect tense. How about `enableCrossOs`?

bishal-pdMSFT: I see your points. But I also feel that `enableCrossOs` is not giving complete clarity on what it is doing. That's why I was adding `Compatibility`. How about we call it what is precisely is - `enableCrossOsArchive`?

packages/cache/src/internal/cacheHttpClient.ts
bishal-pdMSFT: >Previously they were identified by not having gzip in version but moving forward will have gzip in version.

Yes, this is what I meant. We need not put empty compression method here for gzip and let it be handled just like any other compression method.

packages/cache/src/cache.ts
pdotl: `enableCrossOsArchive` is much more precise yes 👍

packages/cache/src/internal/cacheHttpClient.ts
bishal-pdMSFT: Yes agree, we can remove the fallback.

bishal-pdMSFT, line 86: Don't we need to handle Windows -> linux case as well? Would it be better to name this `non-cross-os`?

bishal-pdMSFT: Actually this will get handled automatically as windows will create cache with this suffix version

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: > Previously they were identified by not having gzip in version but moving forward will have gzip in version.

Then we can do that. Remove empty compression case here.

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: Agreed we can remove gzip fallback for old caches, if that's what you mean

Phantsure: Done

packages/cache/src/internal/cacheHttpClient.ts
Phantsure: Empty string still remains if no compression method is passed. Removed gzip case. This was present since initial commit. Let me check if it is really needed

packages/cache/src/cache.ts
Phantsure: Done

pdotl (commented):
Minor comments

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -73,13 +73,16 @@ function createHttpClient(): HttpClient {
...
+  compressionMethod?: CompressionMethod,
+  enableCrossOsArchive = false
 ): string {
-  const components = paths.concat(
-    !compressionMethod || compressionMethod === CompressionMethod.Gzip
-      ? []
-      : [compressionMethod]
-  )
+  const components = paths
+    .concat([!compressionMethod ? '' : compressionMethod])
```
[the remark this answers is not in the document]
Phantsure: As `compressionMethod` can be undefined.

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -73,13 +73,16 @@ function createHttpClient(): HttpClient {
...
+    .concat(
+      process.platform !== 'win32' || enableCrossOsArchive
+        ? []
+        : ['windows-only']
+    ) // Only check for windows platforms if enableCrossOsArchive is false
```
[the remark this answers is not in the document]
Phantsure: Umm, can we push to `const`?

packages/cache/src/internal/cacheHttpClient.ts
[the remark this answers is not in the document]
pdotl: But previously we were concatenating an empty array which is different than adding an empty string.

packages/cache/src/internal/cacheHttpClient.ts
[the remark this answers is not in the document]
pdotl: We are doing so in [L#88](https://github.com/actions/toolkit/blob/4679d0cfe64af09ec136b2467b5fdc1a81fdf25d/packages/cache/src/internal/cacheHttpClient.ts#L88) too. I guess it would be copying the whole thing into a new array then.

packages/cache/src/internal/cacheHttpClient.ts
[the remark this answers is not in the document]
Phantsure: Good point. Reverting

packages/cache/src/internal/cacheHttpClient.ts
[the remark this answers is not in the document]
Phantsure: Right. But why is it better?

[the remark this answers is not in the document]
pdotl: Readability mostly.  `concat` is used to merge arrays whereas `push` is used to push single items onto it.  While reading through it made sense that concat was creating the components and then we are adding items to it using push. Although `concat` also has performance implications and `push` is generally faster.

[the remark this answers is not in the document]
Phantsure: Got it. Added

pdotl (approved):
LGTM 🚀

packages/cache/src/internal/tar.ts
```diff
@@ -1,155 +1,296 @@
...
+import {
+  CompressionMethod,
+  SystemTarPathOnWindows,
+  ArchiveToolType,
+  TarFilename,
+  ManifestFilename
+} from './constants'
 
 const IS_WINDOWS = process.platform === 'win32'
+exportVariable('MSYS', 'winsymlinks:nativestrict')
```
dscho, line 17: This is a grave mistake. This code is supposed to be a library and not leave any remnants in the environment after it ran.

This can cause failures in GitHub workflow runs (and indeed, it did over here). You basically changed the behavior of every workflow run that involves MSYS2 (or for that matter, Git for Windows).

pdotl: Hi @dscho, I hope this was addressed in this [PR](https://github.com/actions/toolkit/pull/1329)?

pdotl: You are correct. We were in the process of getting it merged, will be creating a release to get it live.

dscho: Thank you.

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -73,13 +73,21 @@ function createHttpClient(): HttpClient {
...
-    !compressionMethod || compressionMethod === CompressionMethod.Gzip
-      ? []
-      : [compressionMethod]
-  )
+  const components = paths
+
+  // Add compression method to cache version to restore
+  // compressed cache as per compression method
+  if (compressionMethod) {
+    components.push(compressionMethod)
```
eregon, line 84: This is a bug, it mutates the user-provided argument `paths`, fix in https://github.com/actions/toolkit/pull/1378

PR #1318, opened 2023-01-24
Pass in the current workspace as a parameter to hashFiles

johnsudol:
Follow up work to allow the runner to use the hashFiles implementation in glob

### DIFF
--- a/packages/glob/__tests__/hash-files.test.ts
+++ b/packages/glob/__tests__/hash-files.test.ts
@@ -51,6 +51,7 @@ describe('globber', () => {
     )
   })
 
+  const emptyDirectory = ''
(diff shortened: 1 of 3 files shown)


packages/glob/__tests__/hash-files.test.ts
```diff
@@ -4,6 +4,7 @@ import {hashFiles} from '../src/glob'
 import {promises as fs} from 'fs'
 
 const IS_WINDOWS = process.platform === 'win32'
+const emptyDirectory = ''
```
johnsudol, line 7: This allows the workspace to default to `process.env['GITHUB_WORKSPACE']`

rentziass (approved):

0xsarawut (approved):

PR #1327, opened 2023-01-26
cache restore: suppress timestamp warnings

ripatel-fd:
Adds GNU tar `--warning=no-timestamp` flag when restoring a cache.

Some popular build tools like Bazel deliberately create files with timestmaps far into the future.

By default, GNU tar will warn about this.
This creates a lot of spam in the cache restore logs.

More information in the GNU tar manual:
https://www.gnu.org/software/tar/manual/html_section/warnings.html


ripatel-fd:
Closing due to inactivity.

PR #1329, opened 2023-01-29
Don't set the MSYS env var globally

### LINKED ISSUE #1312: cache: Sets global MSYS env var, interfering with MSYS2 on Windows runners
The recent >=3.1.x updates introduced a change overwriting globally the "MSYS" env var: https://github.com/actions/toolkit/blob/6c1f9eaae833355a0b212b66c5f2e3ac366de185/packages/cache/src/internal/tar.ts#L17

This env vars is used for configuring MSYS2 behavior and setting it overrides config provided by the user, or changes the default behavior.

It would be nice if this env var was only set for what it is needed and not globally.

@Phantsure

lazka:
b2d865f18051b21447 introduced a call to exportVariable() to export the MSYS env var, which configures the symlink strategy for MSYS2/cygwin binaries it calls.

By setting the env var globally, this also changes the behaviour of other MSYS2 using tools in a CI job, and also overrides MSYS configuration set by the user, which I think was not intended.

To avoid this leakage set the MSYS env var only for the commands which @actions/cache calls.

Fixes #1312

### DIFF
--- a/packages/cache/__tests__/tar.test.ts
+++ b/packages/cache/__tests__/tar.test.ts
@@ -1,5 +1,4 @@
 import * as exec from '@actions/exec'
-import {exportVariable} from '@actions/core'
 import * as io from '@actions/io'
 import * as path from 'path'
 import {
@@ -15,8 +14,6 @@ import * as utils from '../src/internal/cacheUtils'
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 import fs = require('fs')
 
-exportVariable('MSYS', 'winsymlinks:nativestrict')
-
 jest.mock('@actions/exec')
 jest.mock('@actions/io')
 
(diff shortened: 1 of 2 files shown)


lazka:
@Phantsure ^

I don't have much experience with typescript, feedback welcome.

The process.env cast ist needed since exec() is too strict to take process.env. I couldn't find any code setting env vars elsewhere.

packages/cache/src/internal/tar.ts
```diff
@@ -250,7 +248,10 @@ async function getCompressionProgram(
 async function execCommands(commands: string[], cwd?: string): Promise<void> {
   for (const command of commands) {
     try {
-      await exec(command, undefined, {cwd})
+      await exec(command, undefined, {
+        cwd,
+        env: {...(process.env as object), MSYS: 'winsymlinks:nativestrict'}
```
Phantsure, line 253: This looks good.

packages/cache/__tests__/tar.test.ts
```diff
@@ -25,6 +22,8 @@ const IS_MAC = process.platform === 'darwin'
 
 const defaultTarPath = IS_MAC ? 'gtar' : 'tar'
 
+const defaultEnv = {MSYS: 'winsymlinks:nativestrict'};
```
Phantsure, line 25: Semi colon not needed

PR #1345, opened 2023-02-18
Fix zstd version parsing [cache]

### LINKED ISSUE #1341: cacheUtils.getCompressionMethod broke with zstd 1.5.4
The windows-2022 runners have just updated to zstd v1.5.4, this changes [the string](https://github.com/facebook/zstd/commit/9c93dd71cdb301595a8a9e364348967a173c010c) that is used to [detect zstd](https://github.com/actions/toolkit/blob/d2b7d85e7cb61467395604aae1b84439c320cb84/packages/cache/src/internal/cacheUtils.ts#L100).
This causes a cache miss on this runner.

cdce8p:
zstd `1.5.4` changed the `--version` output which broke the version parsing and subsequent cache restores across different OS versions (e.g. `ubuntu-20.04` and `ubuntu-22.04` as those use different `zstd` versions). https://github.com/facebook/zstd/commit/9c93dd71cdb301595a8a9e364348967a173c010c

```diff
-*** zstd command line interface 64-bits v1.5.2, by Yann Collet ***
+*** Zstandard CLI (64-bit) v1.5.4, by Yann Collet ***
```

Use `--quiet` option to output machine readable version.
https://www.mankier.com/1/zstd#Options-Operation_Modifiers

Closes #1341
Closes https://github.com/actions/cache/issues/1110
Closes https://github.com/actions/cache/issues/1112

### DIFF
--- a/packages/cache/__tests__/tar.test.ts
+++ b/packages/cache/__tests__/tar.test.ts
@@ -73,7 +73,7 @@ test('zstd extract tar', async () => {
       .concat(IS_MAC ? ['--delay-directory-restore'] : [])
       .concat([
         '--use-compress-program',
-        IS_WINDOWS ? '"zstd -d --long=30"' : 'unzstd --long=30'
+        IS_WINDOWS ? '"zstd -d --long=30"' : '"unzstd --long=30"'
       ])
       .join(' '),
     undefined,
@@ -233,7 +233,7 @@ test('zstd create tar', async () => {
       .concat(IS_MAC ? ['--delay-directory-restore'] : [])
       .concat([
         '--use-compress-program',
-        IS_WINDOWS ? '"zstd -T0 --long=30"' : 'zstdmt --long=30'
+        IS_WINDOWS ? '"zstd -T0 --long=30"' : '"zstdmt --long=30"'
       ])
       .join(' '),
     undefined, // args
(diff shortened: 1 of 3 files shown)


pdotl:
@cdce8p appreciate your PR for fixing this. However, I found the current semver version check is redundant as version is null. Please see [here](https://github.com/actions/toolkit/blob/c2d3089f8337053c39f7c368d6786a52077ea90b/packages/cache/src/internal/cacheUtils.ts#L99-L113). So for all zstd command line strings, we are just calling `ZstdWithoutLong`. For now I've raised this [PR](https://github.com/actions/toolkit/pull/1353) to fix the issue introduced by zstd version change. I'll followup with enabling `zstd --long` in an upcoming release PR. For this reason closing your PR here. Appreciate the effort and also incorporated `--quiet`  taking after your PR.

cdce8p:
Thanks for picking this up @pdotl!

> However, I found the current semver version check is redundant as version is null. Please see [here](https://github.com/actions/toolkit/blob/c2d3089f8337053c39f7c368d6786a52077ea90b/packages/cache/src/internal/cacheUtils.ts#L99-L113). So for all zstd command line strings, we are just calling `ZstdWithoutLong`.

Yeah, noticed that as well as I was testing my fix. Turned out that `zstd --long` didn't even work on the ubuntu runners after the version parsing was fixed. The commands needed to be string escaped `'"unzstd --long=30"'` / `'"zstdmt --long=30"'`.

PR #1351, opened 2023-02-19
Revert "Standardize behaviour of no_proxy environmental variable"

emmyforeignet:
Reverts actions/toolkit#1223


emmyforeignet (commented):
packages/http-client/__tests__/proxy.test.ts

PR #1353, opened 2023-02-20
Hotfix Zstd breaking due to version change in runners

pdotl:
## Summary

Currently zstd check is dependent on the text output of the `zstd --version` to include `zstd command line interface`. This text changed to `*** Zstandard CLI (64-bit) v1.5.4, by Yann Collet ***` breaking the check and `@actions/cache` always falling back to `gzip` compression.

## Implementation

- Set to use `gzip` if the version output string is empty which should be the case if zstd is not installed. 
- Introduced `--quiet` flag to `zstd` so moving on the `zstd --version` should output machine readable version parsable via SemVer instead of text.
- Removed code that checks for version less than 1.3.2 as it was not working. Now we default to using zstd without `--long` as that is what currently happens.

## Next Steps

Will have another PR that fixes use of `zstd` with `--long` as intended. This release should unblock `zstd` usage.

## Issues

Fixes https://github.com/actions/toolkit/issues/1341

### DIFF
--- a/packages/cache/src/internal/cacheUtils.ts
+++ b/packages/cache/src/internal/cacheUtils.ts
@@ -79,7 +79,7 @@ async function getVersion(
   additionalArgs.push('--version')
   core.debug(`Checking ${app} ${additionalArgs.join(' ')}`)
   try {
-    await exec.exec(`${app}`, additionalArgs, {
+    await exec.exec(`${app} `, additionalArgs, {
       ignoreReturnCode: true,
       silent: true,
       listeners: {
--- a/packages/cache/src/internal/cacheUtils.ts
+++ b/packages/cache/src/internal/cacheUtils.ts
@@ -99,12 +99,17 @@ async function getVersion(app: string, args?: string[]): Promise<string> {
 export async function getCompressionMethod(): Promise<CompressionMethod> {
   const versionOutput = await getVersion('zstd', ['--quiet'])
   const version = semver.clean(versionOutput)
-  if (version) {
+
+  if (versionOutput === '') {
+    // zstd is not installed
+    return CompressionMethod.Gzip
+  } else if (!version || semver.lt(version, 'v1.3.2')) {
+    // zstd is installed but using a version earlier than v1.3.2
+    // v1.3.2 is required to use the `--long` options in zstd
     return CompressionMethod.ZstdWithoutLong
   } else {
-    return CompressionMethod.Gzip
+    return CompressionMethod.Zstd
   }
-
 }
 
 export function getCacheFileName(compressionMethod: CompressionMethod): string {
--- a/packages/cache/src/internal/cacheUtils.ts
+++ b/packages/cache/src/internal/cacheUtils.ts
@@ -95,7 +95,9 @@ async function getVersion(app: string): Promise<string> {
 // Use zstandard if possible to maximize cache performance
 export async function getCompressionMethod(): Promise<CompressionMethod> {
   const versionOutput = await getVersion('zstd')
+  core.debug(`versionOutput: ${versionOutput}`)
   const version = semver.clean(versionOutput)
+  core.debug(`version: ${version}`)
 
   if (!versionOutput.toLowerCase().includes('zstd command line interface')) {
     // zstd is not installed
--- a/packages/cache/src/internal/cacheUtils.ts
+++ b/packages/cache/src/internal/cacheUtils.ts
@@ -72,10 +72,10 @@ export async function unlinkFile(filePath: fs.PathLike): Promise<void> {
 }
 
 async function getVersion(app: string, args?: string[]): Promise<string> {
-  core.debug(`Checking ${app} --version`)
   let versionOutput = ''
   typeof args !== 'undefined' ? args : (args = [])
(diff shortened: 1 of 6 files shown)


packages/cache/src/internal/cacheUtils.ts
```diff
@@ -71,11 +71,13 @@ export async function unlinkFile(filePath: fs.PathLike): Promise<void> {
   return util.promisify(fs.unlink)(filePath)
 }
 
-async function getVersion(app: string): Promise<string> {
+async function getVersion(app: string, args?: string[]): Promise<string> {
```
bishal-pdMSFT, line 74: As mentioned in another comment, this need not be an argument. Will simplify the code as well.

bishal-pdMSFT:
Checks are failing. Please fix the unit tests.

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -94,18 +96,14 @@ async function getVersion(app: string): Promise<string> {
 
 // Use zstandard if possible to maximize cache performance
 export async function getCompressionMethod(): Promise<CompressionMethod> {
-  const versionOutput = await getVersion('zstd')
+  const versionOutput = await getVersion('zstd', ['--quiet'])
   const version = semver.clean(versionOutput)
 
-  if (!versionOutput.toLowerCase().includes('zstd command line interface')) {
+  if (versionOutput === '' || version === null) {
```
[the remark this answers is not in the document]
pdotl: `versionOutput` is initialised with `''` but `version` returns null due to `semver.clean` returning null for versionOutput value that is not a valid semver.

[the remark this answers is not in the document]
pdotl: Yes Bishal, I took a call to remove that from PR as it was currently redundant and we were not using `--long` at all. Since folks are having issues using `zstd` currently, I was thinking we can introduce `--long` after testing that it works but release now to unblock `zstd` usage. Although the change to fix `long` is also not much work.

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -94,18 +96,14 @@ async function getVersion(app: string): Promise<string> {
 
 // Use zstandard if possible to maximize cache performance
 export async function getCompressionMethod(): Promise<CompressionMethod> {
-  const versionOutput = await getVersion('zstd')
+  const versionOutput = await getVersion('zstd', ['--quiet'])
```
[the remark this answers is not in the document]
bishal-pdMSFT: Got it

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -71,11 +71,13 @@ export async function unlinkFile(filePath: fs.PathLike): Promise<void> {
   return util.promisify(fs.unlink)(filePath)
 }
 
-async function getVersion(app: string): Promise<string> {
-  core.debug(`Checking ${app} --version`)
+async function getVersion(app: string, args?: string[]): Promise<string> {
```
bishal-pdMSFT, line 74: Does using default value for args work? i.e. `args: string[] = []`. If so then we need not handle the undefined thing in L76

bishal-pdMSFT, line 74: Name args as `additionalArgs` as the primay arg is `--version` which is not being passed.

packages/cache/src/internal/cacheUtils.ts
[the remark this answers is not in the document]
bishal-pdMSFT: Ok.

bishal-pdMSFT (approved):

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -71,11 +71,15 @@ export async function unlinkFile(filePath: fs.PathLike): Promise<void> {
...
+async function getVersion(
+  app: string,
+  additionalArgs: string[] = []
+): Promise<string> {
   let versionOutput = ''
+  additionalArgs.push('--version')
+  core.debug(`Checking ${app} ${additionalArgs.join(' ')}`)
   try {
-    await exec.exec(`${app} --version`, [], {
+    await exec.exec(`${app}`, additionalArgs, {
```
cdce8p, line 82: Thanks again @pdotl!

Just saw this while looking at the PR. `'${app}'` seems a bit redundant. The first argument should be a string and `app` is already one. Maybe you can fix that when you address the `zstd --long` issue. It would be a bit much to open a separate PR for it 😅

Phantsure (commented):
Looks fine to remove --long check entirely. --quiet would supress warnings but not clear how that would help in semver readable text

cdce8p:
> --quiet would supress warnings but not clear how that would help in semver readable text

Found this in the documentation for `zstd`
> V, --version: display version number and exit. Advanced : -vV also displays supported formats. -vvV also displays POSIX support. -q will only display the version number, suitable for machine reading.

packages/cache/src/internal/cacheUtils.ts
pdotl: Sure @cdce8p will take this up in the next PR.

PR #1361, opened 2023-03-01
Bypass proxy on loopback IPs (localhost, 127.*, ::1 etc)

fhammerl:
Requests to localhost should not be proxied (if the proxy is on another device / network, the proxied request targeting 'localhost' will likely be lost)

The [runner](https://github.com/actions/runner/blob/5421fe3f7107f770c904ed4c7e506ae7a5cde2c2/src/Runner.Sdk/RunnerWebProxy.cs#L203) also ignores loopback IPs
[Golang as well](https://go.dev/src/vendor/golang.org/x/net/http/httpproxy/proxy.go?s=4195:4265#:~:text=if%20ip.IsLoopback,%7D)

### DIFF
--- a/packages/http-client/__tests__/proxy.test.ts
+++ b/packages/http-client/__tests__/proxy.test.ts
@@ -223,19 +223,29 @@ describe('proxy', () => {
     expect(_proxyConnects).toEqual(['httpbin.org:443'])
   })
 
-  it('HttpClient does basic https get request when bypass proxy', async () => {
-    process.env['https_proxy'] = _proxyUrl
-    process.env['no_proxy'] = 'httpbin.org'
+it('HttpClient bypasses proxy for loopback addresses (localhost, ::1, 127.*)', async () => {
+  // setup a server listening on localhost:8091
+  var server = http.createServer(function (request, response) {
+    response.writeHead(200);
+    request.pipe(response);
+  });
+  await server.listen(8091)
+  try {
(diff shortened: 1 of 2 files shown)


packages/http-client/__tests__/proxy.test.ts
```diff
@@ -237,6 +237,31 @@ describe('proxy', () => {
...
+    try {
+      process.env['http_proxy'] = _proxyUrl
+      const httpClient = new httpm.HttpClient()
+      let res = await httpClient.get('http://localhost:8091')
+      expect(res.message.statusCode).toBe(200)
+      res = await httpClient.get('http://127.0.0.1:8091')
+      expect(res.message.statusCode).toBe(200)
+
+      // no support for ipv6 for now
+      expect(httpClient.get('http://[::1]:8091')).rejects.toThrow()
```
JoannaaKL, line 256: Nit: I think this assertion is not relevant to this test case (we are testing the behaviour of proxy for loopback address). Maybe it could be moved into separate test?

JoannaaKL (approved):

packages/http-client/__tests__/proxy.test.ts
fhammerl: Currently we don't support ipv6 in `httpClient.get('http://[::1]:8091')`. Once we do (aka `httpClient.get` no longer throws), I want this test to fail, because it means proxy bypass for ipv6 loopback adresses can now be tested and verified.

PR #1363, opened 2023-03-06
Updating blob storage client to 12.13.0 in actions/cache

kotewar:
This PR updates the blob storage client to v12.13.0 which is the latest to accommodate the latest SDK changes in cache library.

Also updated the typescript version because the [builds](https://github.com/actions/toolkit/actions/runs/4343131738/jobs/7584792497) were failing and azure sdk [uses](https://github.com/Azure/azure-sdk-for-js/commit/526926acc6b089b0e199c54a48bab4afc7e052e1) typescript 4.8.0

### DIFF
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -51,6 +51,6 @@
   "devDependencies": {
     "@types/semver": "^6.0.0",
     "@types/uuid": "^3.4.5",
-    "typescript": "^3.8.3"
+    "typescript": "~4.8.0"
   }
 }
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -44,7 +44,7 @@
     "@actions/io": "^1.0.1",
     "@azure/abort-controller": "^1.1.0",
     "@azure/ms-rest-js": "^2.6.0",
-    "@azure/storage-blob": "^12.8.0",
+    "@azure/storage-blob": "^12.13.0",
     "semver": "^6.1.0",
     "uuid": "^3.3.3"
   },
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@actions/cache",
-  "version": "3.1.4",
+  "version": "3.1.5",
   "preview": true,
   "description": "Actions cache lib",
   "keywords": [
--- a/packages/cache/package.json
+++ b/packages/cache/package.json
@@ -51,6 +51,6 @@
   "devDependencies": {
     "@types/semver": "^6.0.0",
     "@types/uuid": "^3.4.5",
-    "typescript": "~4.8.0"
+    "typescript": "^4.8.0"
   }
 }
--- a/packages/cache/src/internal/downloadUtils.ts
+++ b/packages/cache/src/internal/downloadUtils.ts
@@ -242,7 +242,9 @@ export async function downloadCacheStorageSDK(
     // If the file exceeds the buffer maximum length (~1 GB on 32-bit systems and ~2 GB
     // on 64-bit systems), split the download into multiple segments
     // ~2 GB = 2147483647, beyond this, we start getting out of range error. So, capping it accordingly.
-    const maxSegmentSize = Math.min(2147483647, buffer.constants.MAX_LENGTH)
+
+    // Updated segment size to 128MB to complete a segment faster and fail fast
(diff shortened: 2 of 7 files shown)


packages/cache/tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "baseUrl": "./",
     "outDir": "./lib",
     "rootDir": "./src",
-    "lib": ["es6", "dom"]
+    "lib": ["es6", "dom"],
+    "useUnknownInCatchVariables": false
```
kotewar, line 8: Added this because the ts compiler raised issues with the catch block err variable having no type defined. Use of  `Error`, `any` and `unknown` was also not allowed.

packages/cache/package.json
```diff
@@ -44,13 +44,13 @@
...
-    "@azure/storage-blob": "^12.8.0",
+    "@azure/storage-blob": "^12.13.0",
     "semver": "^6.1.0",
     "uuid": "^3.3.3"
   },
   "devDependencies": {
     "@types/semver": "^6.0.0",
     "@types/uuid": "^3.4.5",
-    "typescript": "^3.8.3"
+    "typescript": "~4.8.0"
```
bishal-pdMSFT, line 54: The root level package.json uses a different `typescript` [version](https://github.com/actions/toolkit/blob/f3de1e53d628310756183415a00887f3ac6e15cd/package.json#L30). Wondering how Lerna resolves the version when running build/test commands.

packages/cache/src/internal/downloadUtils.ts
```diff
@@ -242,7 +242,9 @@ export async function downloadCacheStorageSDK(
     // If the file exceeds the buffer maximum length (~1 GB on 32-bit systems and ~2 GB
     // on 64-bit systems), split the download into multiple segments
     // ~2 GB = 2147483647, beyond this, we start getting out of range error. So, capping it accordingly.
-    const maxSegmentSize = Math.min(2147483647, buffer.constants.MAX_LENGTH)
+
+    // Updated segment size to 128MB = 134217728 bytes, to complete a segment faster and fail fast
+    const maxSegmentSize = Math.min(134217728, buffer.constants.MAX_LENGTH)
```
bishal-pdMSFT, line 247: Should it be one less to keep it similar as earlier?

kotewar: I thought of it too, but we had done it for 2GB not to cross the limit. Here 128 is safe. So every segment will be complete 128MB and not one byte less, which I thought shouldn't be a problem?

packages/cache/package.json
kotewar: On first attempt I had received an error saying version mismatch, but running `npm install` and updating the `package-lock.json` file resolved the issue. I think it uses lockfile to resolve the dev dependencies as well.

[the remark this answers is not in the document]
kotewar: I'll update it to caret symbol. Didn't realise there was a difference.

kotewar:
Can someone from @actions/actions-runtime team please review if the typescript update has been done correctly? We have to upgrade it to 4.x to support the newer versions of the @azure/storage-sdk. 🙇🏽‍♂️

packages/cache/src/internal/downloadUtils.ts
bishal-pdMSFT: Sounds good!

bishal-pdMSFT (approved):

packages/cache/src/options.ts
```diff
@@ -101,7 +101,7 @@ export function getDownloadOptions(copy?: DownloadOptions): DownloadOptions {
     useAzureSdk: true,
     downloadConcurrency: 8,
     timeoutInMs: 30000,
-    segmentTimeoutInMs: 3600000,
+    segmentTimeoutInMs: 600000,
```
lukaw3d, line 104: These were not updated to match:
- https://github.com/actions/toolkit/blob/457303960f03375db6f033e214b9f90d79c3fe5c/packages/cache/src/options.ts#L50-L55
- https://github.com/actions/toolkit/blob/457303960f03375db6f033e214b9f90d79c3fe5c/packages/cache/src/internal/downloadUtils.ts#L267

and README was updated in https://github.com/actions/toolkit/pull/1367 but RELEASES wasn't https://github.com/actions/toolkit/blob/457303960f03375db6f033e214b9f90d79c3fe5c/packages/cache/RELEASES.md?plain=1#L160-L162

PR #1373, opened 2023-03-14
Removing childprocess for rmRF

vmjoseph:
- Uses node's rm instead of custom rmRF logic
- Fixes locked file tests
- Removes symlink tests for windows specific cases


packages/io/__tests__/io.test.ts
```diff
@@ -331,11 +332,26 @@ describe('rmRF', () => {
     await fs.appendFile(filePath, 'some data')
     await assertExists(filePath)
 
-    const fd = await fs.open(filePath, 'r')
-    await io.rmRF(testPath)
-
-    await assertNotExists(testPath)
+    // for windows, we need to explicitly set an exlcusive lock
```
cory-miller, line 335:
```suggestion
    // for windows, we need to explicitly set an exclusive lock
```

packages/io/__tests__/io.test.ts
```diff
@@ -331,11 +332,26 @@ describe('rmRF', () => {
...
+    // or we won't get an EBUSY error. this can be fixed after
+    // https://github.com/libuv/libuv/issues/3267 is resolved
```
cory-miller, line 337: Can you move this inside the try statement? It's about the EPERM comment. I don't think Node considers its `fs.open` call not creating a lock on Windows a bug.

packages/io/src/io-util.ts
```diff
@@ -6,16 +6,20 @@ export const {
...
   readlink,
   rename,
+  rm,
   rmdir,
   stat,
   symlink,
   unlink
 } = fs.promises
-
+// export const {open} = 'fs'
```
cory-miller, line 19:
```suggestion
```

packages/io/src/io-util.ts
```diff
@@ -6,16 +6,20 @@ export const {
...
 export const IS_WINDOWS = process.platform === 'win32'
+export const UV_FS_O_EXLOCK = 0x10000000
```
cory-miller, line 21:
```suggestion
// See https://github.com/nodejs/node/blob/d0153aee367422d0858105abec186da4dff0a0c5/deps/uv/include/uv/win.h#L691
export const UV_FS_O_EXLOCK = 0x10000000
```

cory-miller (approved):

packages/io/__tests__/io.test.ts
```diff
@@ -331,11 +332,26 @@ describe('rmRF', () => {
...
+    // exclusive lock windows flag: https://github.com/nodejs/node/blob/c2e4b1fa9ad0b744616c4e4c13a5017772a630c4/deps/uv/src/win/fs.c#L499-L513
+    const fd = await fs.open(
+      filePath,
+      fs.constants.O_RDONLY | ioUtil.UV_FS_O_EXLOCK
+    )
+    if (ioUtil.IS_WINDOWS) {
+      try {
+        // additionally, can't stat an open file on Windows without getting EPERM
+        await io.rmRF(testPath)
```
jww3, line 346: you're expecting an exception to be thrown, right?  if so, we probably want to fail the test on the line immediately following the `io.rmRM(testPath)` call since we expect anything else within the `try` block to be unreachable code.

packages/io/__tests__/io.test.ts
```diff
@@ -331,11 +332,25 @@ describe('rmRF', () => {
...
+    )
+    if (ioUtil.IS_WINDOWS) {
+      try {
+        // for windows, we need to explicitly set an exclusive lock
+        // or we won't get an EBUSY error. this can be fixed after
+        // https://github.com/libuv/libuv/issues/3267 is resolved
+        await io.rmRF(testPath)
+      } catch (err) {
+        expect(err.message).toContain('EBUSY')
+      }
```
cory-miller, line 348:
```suggestion
    // For windows we need to explicitly set an exclusive lock flag, because by default Node will open the file with the 'Delete' FileShare flag.
    // See the exclusive lock windows flag definition:
   // https://github.com/nodejs/node/blob/c2e4b1fa9ad0b744616c4e4c13a5017772a630c4/deps/uv/src/win/fs.c#L499-L513
    const fd = await fs.open(
      filePath,
      fs.constants.O_RDONLY | ioUtil.UV_FS_O_EXLOCK
    )
    if (ioUtil.IS_WINDOWS) {
      // On Windows, we expect an error due to an lstat call implementation in the underlying libuv code.
      // See https://github.com/libuv/libuv/issues/3267 is resolved
      await expect(() => io.rmRF(testPath)).rejects.toThrow('EBUSY')
```

PR #1378, opened 2023-03-20
cache - getCacheVersion - dup paths array

### LINKED ISSUE #1377: @actions/cache not restoring cache, version mismatch?
**Describe the bug**

We use @actions/cache in a popular setup action.  Recently updated the lock file which previously used 3.0.6 without issue.  Caches appear to be saved, but they are not restored with `restoreCache`.

In the first job run, often the cache was saved.  If the next run did not restore it, the run also had the common error:

'Failed to save: Unable to reserve cache with key <key>, another job may be creating this cache. More details: Cache already exists.'

But, inspecting all the keys, the key of the saved cache from the previous job run is the same as the key it cannot find in the 2nd job run.  See https://github.com/MSP-Greg/puma/actions/runs/4452716291/jobs/7820684254#step:3:121

As to versions, first, we updated to current (3.1.2).  That did not work with code that worked with 3.0.6.  Investigated further:
```
                     1st Job Run               2nd Job Run
@actions/cache  Ubuntu  macOS  Windows    Ubuntu  macOS  Windows
      3.0.6       ✅     ✅      ✅        ✅     ✅      ✅
      3.1.0       ✅     ✅      ❌        ✅     ✅      ✅
      3.1.1       ❌     ❌      ❌        ✅     ✅      ✅
      3.1.2       ❌     ❌      ❌        ❌     ❌      ❌
      3.1.4       ❌     ❌      ❌        ❌     ❌      ❌
```

As noted above, version 3.1.2 and later never restores a cache.  3.1.1 did not restore a cache on the first, run but did on the second run.

**To Reproduce**
See https://github.com/MSP-Greg/puma/actions/workflows/actions-cache.yaml.  Each run in Puma was done with a corresponding branch in https://github.com/MSP-Greg/setup-ruby that was built with the given @actions/cache version.

**Expected behavior**
Versions after 3.0.6 restore caches.

**Screenshots**
NA

NA

MSP-Greg:
Previous to commit b2d865f18051, code used `components = paths.concat(...)`, which left `paths` unchanged.

The commit changed to using `components = paths`, then pushing onto `components` later.  This changes the `paths` array.

Closes #1377

### DIFF
--- a/packages/cache/src/internal/cacheHttpClient.ts
+++ b/packages/cache/src/internal/cacheHttpClient.ts
@@ -76,7 +76,8 @@ export function getCacheVersion(
   compressionMethod?: CompressionMethod,
   enableCrossOsArchive = false
 ): string {
-  const components = paths
+  // don't pass changes upstream
+  const components = paths.slice()
 
   // Add compression method to cache version to restore
   // compressed cache as per compression method


MSP-Greg:
Not sure how deep the copy/clone should be.  Maybe a better choice would be the below, or something similar:
```js
JSON.parse(JSON.stringify(arr))
```

WebeWizard:
bump.  I think a lot of people have been patiently waiting for this fix.

deemp:
@kotewar, @rentziass, please, review this PR. 

The fix suggested in https://github.com/actions/toolkit/pull/1378#issuecomment-1478388929 worked for me.

After https://github.com/deemp/cache-nix-too/commit/93041f23b0c50519ac84e1d59e3405660db99be2 cache was restored successfully in the second run https://github.com/deemp/cache-nix-too/actions/runs/5573652888/jobs/10181692942#step:4:19

ali-kafel:
I have run into this issue as well can we merge and release please?

ali-kafel:
@robherley @bethanyj28  Sorry for mentioning you but it would be great if we could get a review on this from the actions team

robherley (approved):

robherley:
👋 The `@actions/cache` package now has this fix within v3.2.3, thanks for the contribution!

PR #1383, opened 2023-03-23
Mention single quote limitation on documentation of exec.exec and exec.getExecOutput

### LINKED ISSUE #1382: [Documentation] Update exec.exec documentation to mention single quote limitation
**Describe the enhancement**

The current documentation of exec.exec does not mention the limitation
regarding single quotes and double quotes. In the current implementation,
only double quotes are supported for splitting arguments.

The documentation should be updated to include
this information to prevent confusion and save developers' time.

Current document on `exec.exec` is as follows.

https://github.com/actions/toolkit/blob/457303960f03375db6f033e214b9f90d79c3fe5c/packages/exec/src/exec.ts#L7-L16

```
 * @param     commandLine        command to execute (can include additional args). Must be correctly escaped.
```

This lacks information on single quotes and double quotes, only the latter of which is supported currently. I would like to append the information on the doc.

**Code Snippet**

```
 * @param     commandLine        command to execute (can include additional args). Must be correctly escaped. Currently, only double quotes are supported for splitting args.
```

**Additional information**

I spent many hours debugging the reason why the following code did not work as expected

```bash
exec.exec(`git config --local some.key 'some value with white space'`)
```

It is only after I read the internal code that I realized the root cause is the single quotes.
I believe the slight change on doc can prevent developers like me from failing into pitfalls.

https://github.com/actions/toolkit/blob/457303960f03375db6f033e214b9f90d79c3fe5c/packages/exec/src/toolrunner.ts#L568

aeroastro:
This resolves #1382.

By describing the limitation of single quote,
I would like to prevent confusion and save developers' time.

https://github.com/actions/toolkit/blob/457303960f03375db6f033e214b9f90d79c3fe5c/packages/exec/src/toolrunner.ts#L568

### DIFF
--- a/packages/exec/src/exec.ts
+++ b/packages/exec/src/exec.ts
@@ -9,7 +9,7 @@ export {ExecOptions, ExecOutput, ExecListeners}
  * Output will be streamed to the live console.
  * Returns promise with return code
  *
- * @param     commandLine        command to execute (can include additional args). Must be correctly escaped.
+ * @param     commandLine        command to execute (can include additional args). Must be correctly escaped. Currently, only double quotes are supported for splitting args.
  * @param     args               optional arguments for tool. Escaping is handled by the lib.
  * @param     options            optional exec options.  See ExecOptions
  * @returns   Promise<number>    exit code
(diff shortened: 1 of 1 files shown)


aeroastro:
@nebuk89 @bdehamer 
Could you please review this Pull Request?

PR #1392, opened 2023-04-08
Provide and use local cache client

Cakasim:


moseoh:
Hi, I was interested in the same feature and came across this PR. I noticed that it was closed—could you share why? Was there a specific reason, or any blockers that prevented it from being merged?
I’d love to understand more, as this functionality would be really useful. Thanks!

Issue #1421, opened 2023-05-10
[SECURITY] Toolkit `setSecret` does not protect against insecure uses


JLLeitschuh:
Original report on H1 to the GitHub security team: https://hackerone.com/reports/1929487

---

## Description: 

From the GitHub actions toolkit documentation:
> **WARNING** The add-mask and setSecret  commands only support single line secrets. To register a multiline secrets you must register each line individually otherwise it will not be masked.
> \- https://github.com/actions/toolkit/blob/main/docs/commands.md?plain=1#LL53-L53C189

However, if you look at the code for the `setSecret` method doesn't protect against this vulnerability in any way.

https://github.com/actions/toolkit/blob/457303960f03375db6f033e214b9f90d79c3fe5c/packages/core/src/core.ts#L95-L101

## Steps To Reproduce:

Run the following code in any action script:
```javascript
core.setSecret('VeryComplexSecretThat\nThatIsMultiline\r\nOfBothTypes')
```
Prints the following, without erroring, when this mask will have, as documented, no effect.
```
::add-mask::VeryComplexSecretThat%0AThatIsMultiline%0D%0AOfBothTypes'
```

## Problems

1. The `setSecret` method doesn't document this behaviour in the API docs, it's an external documentation line only found in the docs. AT MINIMUM this should be documented on the method itself.
2. The actions toolkit library is supposed to be an API to make interacting with GitHub actions easier/safer. By not incorporating any sort of protections on this API, this API leaves users vulnerable to leaking sensitive multi-line secrets to the console. 

Fundamentally, this is a security issue because it fails open instead of failing closed.

## Impact

Multi-line secrets passed to `core.setSecret` do not protect those secrets. But because the lack of protection is silent, instead of an error, the end user will not know they are not protected, although the toolkit provides no indication they are not.

---

Response from GitHub Hacker one Triage Team:

> Hi @jlleitschuh,
>
> Thanks for the submission! This is an intentional design decision and is working as expected and as currently documented. We may make this functionality more strict in the future, but we don't have anything to announce right now. As a result, this is not eligible for reward under the Bug Bounty program.
>
> However, because the Toolkit repository is open source, [we encourage you to open an issue or a PR to help further clarify the documentation for this method](https://github.com/actions/toolkit/blob/main/.github/CONTRIBUTING.md).
>
> Best regards and happy hacking!

---

[...]

PR #1427, opened 2023-05-22
Add support for multi-line secrets in setSecret

rentziass:
As reported in https://github.com/actions/toolkit/issues/1421, the current version of `setSecret` produces unexpected behaviour when provided with a multi-line secret in that the secret isn't masked throughout the workflow logs. This splits any multi-line secret and masks each line as a single secret (documented workaround for this kind of secret).

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -50,7 +50,10 @@ function setSecret(secret: string): void {}
 
 Now, future logs containing BAR will be masked. E.g. running `echo "Hello FOO BAR World"` will now print `Hello FOO **** World`.
 
-**WARNING** The add-mask and setSecret  commands only support single line secrets. To register a multiline secrets you must register each line individually otherwise it will not be masked.
+**WARNING** The add-mask command only supports single line secrets. To register
+a multiline secret you must register each line individually otherwise it will
+not be masked, if you use `@actions/core >= 1.11.0` `setSecret` will do this
+for you.
 
 **WARNING** Do **not** mask short values if you can avoid it, it could render your output unreadable (and future steps' output as well).
 For example, if you mask the letter `l`, running `echo "Hello FOO BAR World"` will now print `He*********o FOO BAR Wor****d`
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -50,10 +50,9 @@ function setSecret(secret: string): void {}
 
 Now, future logs containing BAR will be masked. E.g. running `echo "Hello FOO BAR World"` will now print `Hello FOO **** World`.
 
-**WARNING** The add-mask command only supports single line secrets. To register
-a multiline secret you must register each line individually otherwise it will
-not be masked, if you use `@actions/core >= 1.11.0` `setSecret` will do this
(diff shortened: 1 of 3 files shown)


packages/core/src/core.ts
```diff
@@ -97,7 +97,11 @@ export function exportVariable(name: string, val: any): void {
  * @param secret value of the secret
  */
 export function setSecret(secret: string): void {
-  issueCommand('add-mask', {}, secret)
+  secret.split(/\r?\n|\r/).forEach(part => {
```
jww3, line 100: Why not just `/[\r\n]/` ?

rentziass: On Windows `os.EOL` is `\r\n`, I would argue against `\r` on its own if anything, is that realistically being used in multi line text? 🤔

docs/commands.md
```diff
@@ -50,7 +50,10 @@ function setSecret(secret: string): void {}
 
 Now, future logs containing BAR will be masked. E.g. running `echo "Hello FOO BAR World"` will now print `Hello FOO **** World`.
 
-**WARNING** The add-mask and setSecret  commands only support single line secrets. To register a multiline secrets you must register each line individually otherwise it will not be masked.
+**WARNING** The add-mask command only supports single line secrets. To register
+a multiline secret you must register each line individually otherwise it will
+not be masked, if you use `@actions/core >= 1.11.0` `setSecret` will do this
+for you.
```
jww3, line 56:
```suggestion
**WARNING** The add-mask command only supports single-line secrets. To register
a multi-line secret, the recommended practice is to register each line individually. Otherwise, it will
not be masked. `@actions/core >= 1.11.0` `setSecret` will perform this automatically.
```

packages/core/src/core.ts
jww3: The way I read your RegEx is:

`/\r?\n|\r/`
1. any `\n`
2. any `\r\n`
3. any `\r`

Given (1) and (3), I think (2) is effectively moot.

So, it appears to me that you're RegEx is functionally equivalent to `[\r\n]+`.

rentziass: Fair, and it's a lot simpler. Changing now, thanks!

packages/core/src/core.ts
```diff
@@ -97,7 +97,11 @@ export function exportVariable(name: string, val: any): void {
  * @param secret value of the secret
  */
 export function setSecret(secret: string): void {
-  issueCommand('add-mask', {}, secret)
+  for (const part of secret.split(/[\r\n]/)) {
```
jww3, line 100: minor optimization would be to add a `+` outside the square brackets.  That should reduce the number of empty strings the following conditional has to evaluate (particularly on Windows).

rentziass: Great shout, sorry I missed it in your previous comment!

rentziass:
Closing this in favour of https://github.com/actions/toolkit/pull/1430

PR #1430, opened 2023-05-25
Update docs for setSecret

rentziass:
As a follow up to https://github.com/actions/toolkit/issues/1421 we've conducted testing that proved that `setSecret` does in fact mask multi line secrets (and so does `::add-mask::` as long as a multi-line string is escaped). This updates the docs to reflect that and adds a test to ensure `setSecret` does escape the string it's being passed.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -50,7 +50,7 @@ function setSecret(secret: string): void {}
 
 Now, future logs containing BAR will be masked. E.g. running `echo "Hello FOO BAR World"` will now print `Hello FOO **** World`.
 
-**WARNING** The add-mask and setSecret  commands only support single-line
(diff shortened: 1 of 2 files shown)


docs/commands.md
```diff
@@ -50,7 +50,18 @@ function setSecret(secret: string): void {}
 
 Now, future logs containing BAR will be masked. E.g. running `echo "Hello FOO BAR World"` will now print `Hello FOO **** World`.
 
-**WARNING** The add-mask and setSecret  commands only support single line secrets. To register a multiline secrets you must register each line individually otherwise it will not be masked.
+**WARNING** The add-mask and setSecret  commands only support single-line
```
JoannaaKL, line 53:
```suggestion
**WARNING** The add-mask and setSecret commands only support single-line
```

PR #1479, opened 2023-08-03
[Artifacts] Prepare for v2.0.0 of @actions/artifact

konradpabjan:
The first major PR into `main` for an overhaul of `@actions/artifact`.

We decided to just remove everything and start fresh as it is going to be easier to review and the new APIs and code flow is completly different from the current implementation.

This PR does the following:
- Remove dependencies that will no longer be needed `package.json` and `package.lock.json`
- Bumps up all remaining dependencies to their latest version
- Add a comment to the artifacts CI YAML file saying this is disabled (better than deleting the whole file and adding back later)
- Remove all existing tests
- Remove existing docs and add just a tiny bit of information about the upcoming work + link to v1.1.1
- Remove existing code but keep a basic skeleton structure for the upload logic that soon will be added

Tracking https://github.com/github/c2c-actions-checks/issues/1699


packages/artifact/src/internal/crc64.ts
robherley, line 1: 🫡

robherley (approved):
🎨 📠

bethanyj28 (approved):
🚀

jtbandes:
Hi there, this work looks pretty exciting! 🙌 Just wondering if there is a plan to address https://github.com/actions/download-artifact/issues/3 as part of this effort — i.e. to allow cross-workflow artifact downloads?

PR #1481, opened 2023-08-03
[Artifacts] Add artifact Twirp API definitions

bethanyj28:

As we continue to refactor the artifacts package for v4, we need to add our Twirp API types. To do this, we used [hopin-team/twirp-ts](https://github.com/hopin-team/twirp-ts). The `generated` file has an `index.ts` so accessing these entities can be done using:

```
import { ArtifactServiceClientJSON, CreateArtifactRequest, Timestamp } from 'src/generated'
```

PRs following to add the Twirp client.


packages/artifact/src/generated/google/protobuf/timestamp.ts
```diff
@@ -0,0 +1,277 @@
...
+        if (!target)
+            target = this.create();
+        target.seconds = PbLong.from(ms / 1000).toBigInt();
+        target.nanos = 0;
+        if (matches[7])
+            target.nanos = (parseInt("1" + matches[7] + "0".repeat(9 - matches[7].length)) - 1000000000);
+        return target;
+    }
+    create(value?: PartialMessage<Timestamp>): Timestamp {
+        const message = { seconds: 0n, nanos: 0 };
```
robherley, line 233: We may have to wait until we figure out what to do with the big int literals: https://github.com/actions/toolkit/actions/runs/5755118666/job/15601882272?pr=1481

Unsure if twirp-ts has compile options, but maybe we can chat with actions assignment team to see if we can bump the compile target.

Or the manual hacky approach 😅 

```suggestion
        const message = { seconds: BigInt(0), nanos: 0 };
```

PR #1482, opened 2023-08-03
[Artifacts] Name validation + zip specification creation

konradpabjan:
Continuation of work started in https://github.com/actions/toolkit/pull/1479

Adding name validation + zip specification creation that happens at the start of the artifact upload. Also adding tests for each of these functions. Most of this code is similar to the existing one in `@actions/artifact` that was deleted earlier, this just has a few tweaks.

With v2.0.0 we no longer upload files 1 at a time so potentially we may be able to remove the name + path restrictions as zipping might take care of everything across multiple file systems 🤔 We can test this out in the future once things work end to end but it's not as big of a problem as last time


robherley (approved):
:shipit:

PR #1484, opened 2023-08-04
Add option for concurrent cache downloads with timeout

chkimes:
This PR introduces a new default option for download from the Actions Cache. The new option uses the Actions HTTP Client with concurrent downloads from Blob storage, similar to how the Azure SDK works, except this includes much more granular control over timing out downloads during the initial response and while consuming the full response stream. This should once and for all address the issues seen in https://github.com/actions/cache/issues/810.

I have performed extensive testing on the new downloader and comparing to the existing downloader. Here's a week's worth of 99th percentile data showing that the old method is significantly more impacted by Azure Blob issues:

image

Due to a few more connection caching optimizations I made, we can also see by putting percentiles on the x-axis that total execution time is _always_ lower in similar test scenarios:

image

### DIFF
--- a/packages/cache/src/internal/downloadUtils.ts
+++ b/packages/cache/src/internal/downloadUtils.ts
@@ -204,6 +204,153 @@ export async function downloadCacheHttpClient(
 }
 
 /**
+ * Download the cache using the Actions toolkit http-client concurrently
+ *
+ * @param archiveLocation the URL for the cache
+ * @param archivePath the local path where the cache is saved
+ */
+export async function downloadCacheHttpClientConcurrent(
+  archiveLocation: string,
+  archivePath: fs.PathLike,
+  options: DownloadOptions
+): Promise<void> {
+  const archiveDescriptor = await fs.promises.open(archivePath, 'w')
+  const httpClient = new HttpClient('actions/cache', undefined, {
+    socketTimeout: options.timeoutInMs,
+    keepAlive: true
+  })
+  try {
+    const res = await retryHttpClientResponse(
+      'downloadCacheMetadata',
+      async () => await httpClient.request('HEAD', archiveLocation, null, {})
+    )
+
+    const lengthHeader = res.message.headers['content-length']
+    if (lengthHeader === undefined || lengthHeader === null) {
+      throw new Error('Content-Length not found on blob response')
+    }
+
+    const length = parseInt(lengthHeader)
+    if (Number.isNaN(length)) {
+      throw new Error(`Could not interpret Content-Length: ${length}`)
+    }
+
+    const downloads: {
+      offset: number
+      promiseGetter: () => Promise<DownloadSegment>
+    }[] = []
+    const blockSize = 4 * 1024 * 1024
+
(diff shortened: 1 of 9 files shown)


packages/cache/src/internal/downloadUtils.ts
```diff
@@ -203,6 +203,153 @@ export async function downloadCacheHttpClient(
...
+async function downloadSegmentRetry(
+  httpClient: HttpClient,
+  archiveLocation: string,
+  offset: number,
+  count: number
+): Promise<DownloadSegment> {
+  const retries = 5
+  let failures = 0
+
+  while (true) {
```
vmjoseph, line 308: `eslint` has a rule enabled by default that we'd like to keep for not allowing constant conditions, but a 
```js
 while (true){...}
``` 
loop should be fine. Can you add `// eslint-disable-next-line no-constant-condition` in the line above your while loop to silence the linting error?

packages/cache/src/internal/downloadUtils.ts
```diff
@@ -203,6 +203,153 @@ export async function downloadCacheHttpClient(
...
+    const progressFn = progress.onProgress()
+
+    const activeDownloads: {[offset: number]: Promise<DownloadSegment>} = []
+    let nextDownload:
+      | {offset: number; promiseGetter: () => Promise<DownloadSegment>}
+      | undefined
+
+    const waitAndWrite: () => Promise<void> = async () => {
+      const segment = await Promise.race(Object.values(activeDownloads))
+      await archiveDescriptor.write(segment.buffer, 0, segment.count, segment.offset);
```
vmjoseph, line 274: Looks like a few linting errors :) Please run `npm run lint-fix` from root to fix the linting issues.

vmjoseph (changes requested):

packages/cache/src/internal/downloadUtils.ts
chkimes: I opted for specifically disabling the check in loops, per https://github.com/eslint/eslint/issues/5477: 

https://github.com/actions/toolkit/pull/1484/files#diff-6884918dc8291219be508e05e28965b958c734def85324f3b53858ea4702090fR25

cwille97:
@mattiaerre @yacaovsnc This would likely help alleviate a significant amount of failures related to that known Azure Blob issue we discussed.

cwille97 (approved):

kgrzebien (approved):
LGTM

vmjoseph (approved):

adityamaru:
Hey @chkimes, this is a great patch! We recently forked the actions cache to try and getting it working against our minio instance and initially had hand-rolled our own download logic, but then decided what you have here is much more robust. We have seen one instance where a cache restore hangs after making some progress using the method in this patch. Considering each segment has a timeout of 30s we're kinda perplexed how a hang is even possible. Is there any chance you can share how you tested this patch? We're unsure what the best way to setup a testing hook for this method is and wanted to see if you had some ideas.

PR #1486, opened 2023-08-08
Add twirp client

bethanyj28:
Tracking: https://github.com/github/c2c-actions-checks/issues/1699

This PR builds off of https://github.com/actions/toolkit/pull/1481 to initialize the generated twirp client. It uses the `@actions/http-client` package to serve the HTTP request. Additionally, I chose to use the JSON client as it was easier to seralize the data and grok the code. I added this functionality to the `src/shared` folder as it will eventually be leveraged by the download code.

PR #1487, opened 2023-08-09
Utilize client to create and finalize artifact

bethanyj28:
Building on https://github.com/actions/toolkit/pull/1486, this leverages the client to create and finalize artifacts within the `uploadArtifact` function. Additionally, this adds logic to fetch the IDs needed to submit a create and finalize request.


packages/artifact/src/internal/shared/artifact-twirp-client.ts
```diff
@@ -56,16 +56,14 @@ class ArtifactHttpClient implements Rpc {
...
-
     try {
       const response = await this.retryableRequest(async () =>
         this.httpClient.post(url, JSON.stringify(data), headers)
       )
       const body = await response.readBody()
       return JSON.parse(body)
     } catch (error) {
-      throw new Error(error.message)
+      throw new Error(`Failed to ${method}: ${error.message}`)
```
bethanyj28, line 66: This will end up looking like `Failed to CreateArtifact: <error message>` or `Failed to FinalizeArtifact: <error message>`

packages/artifact/src/internal/client.ts
```diff
@@ -39,6 +40,13 @@ export class Client implements ArtifactClient {
     rootDirectory: string,
     options?: UploadOptions | undefined
   ): Promise<UploadResponse> {
-    return uploadArtifact(name, files, rootDirectory, options)
+    try {
+      return uploadArtifact(name, files, rootDirectory, options)
+    } catch (error) {
+      warning(`Failed to upload artifact: ${error}`)
```
konradpabjan, line 46: I would try to improve this message. We could do something similar to what we do if there is a `startup_failure` 

image

I would mention that errors are sometimes temporary and I would add a link to https://www.githubstatus.com/ that users can go to. This can help reduce that amount of issues or support tickets that users may write in. We can also steer users towards rerunning with `debug logs` enabled for more output.

We have this function that detects if we're running on GHES. If we're not on GHES then we shouldn't provide a link to githubstatus

https://github.com/actions/toolkit/blob/92695f58daf7b460a9ae2a4682d8e65a5dea7bc8/packages/cache/src/internal/cacheUtils.ts#L134

packages/artifact/src/internal/client.ts
```diff
@@ -39,6 +40,13 @@ export class Client implements ArtifactClient {
     rootDirectory: string,
     options?: UploadOptions | undefined
   ): Promise<UploadResponse> {
-    return uploadArtifact(name, files, rootDirectory, options)
+    try {
```
konradpabjan, line 43: SideNote, you could use the same `isGHES()` function early on here and immediately fail the upload We don't want anyone on GHES using this package yet and it will be a while until we add support for GHES.

packages/artifact/src/internal/upload/upload-artifact.ts
```diff
@@ -28,13 +31,74 @@ export async function uploadArtifact(
...
     success: true,
     size: 0,
-    id: 0
+    id: parseInt(finalizeArtifactResp.artifactId) // TODO - will this be a problem due to the id being a bigint?
   }
 
   return uploadResponse
 }
+
+function getExpiration(retentionDays?: number): Timestamp | undefined {
```
konradpabjan, line 95: The expiration logic is a bit more complex.

We have to factor in any org/repo level settings that may be configured as that is the maximum allowed value. That is available via this env variable https://github.com/actions/toolkit/blob/91d3933eb52b351f437151400a88ba7d57442a9b/packages/artifact/src/internal/config-variables.ts#L65-L67

Here is this current bit of logic we have
https://github.com/actions/toolkit/blob/91d3933eb52b351f437151400a88ba7d57442a9b/packages/artifact/src/internal/upload-http-client.ts#L71C13-L75

https://github.com/actions/toolkit/blob/91d3933eb52b351f437151400a88ba7d57442a9b/packages/artifact/src/internal/utils.ts#L276-L295

bethanyj28: Ah, I was having trouble finding that logic! Thank you!

konradpabjan (approved):
🚀

PR #1488, opened 2023-08-09
[Artifacts] zip creation + blob storage upload functionality

konradpabjan:

This builds up on the work that was done in https://github.com/actions/toolkit/pull/1479,  https://github.com/actions/toolkit/pull/1481, https://github.com/actions/toolkit/pull/1486  and https://github.com/actions/toolkit/pull/1487

This PR adds zip creation + upload functionality to the upcoming v2.0.0 version of `@actions/artifact`. This in turn will power new v4 versions of `upload-artifact` and `download-artifact`.

Uploads should be significantly faster compared to the v1-v3 versions of upload-artifact since there is no proxying now and we can directly upload to blob storage from the runner. This should also make uploading less error-prone as the entire flow has been simplified.

We're also zipping without creating any temporary files on disk, this should prevent any issues with low disk space that we have seen happen for a subset of v1-v3 users of upload-artifact

### DIFF
--- a/packages/artifact/src/internal/shared/config.ts
+++ b/packages/artifact/src/internal/shared/config.ts
@@ -1,3 +1,9 @@
+// Used for controlling the highWaterMark value of the zip that is being streamed
+// The same value is used as the chunk size that is use during upload to blob storage
+export function getUploadChunkSize(): number {
+  return 8 * 1024 * 1024 // 8 MB Chunks
+}
+
 export function getRuntimeToken(): string {
   const token = process.env['ACTIONS_RUNTIME_TOKEN']
   if (!token) {
--- a/packages/artifact/src/internal/upload/upload-artifact.ts
+++ b/packages/artifact/src/internal/upload/upload-artifact.ts
@@ -56,7 +56,7 @@ export async function uploadArtifact(
   const createArtifactReq: CreateArtifactRequest = {
     workflowRunBackendId: backendIds.workflowRunBackendId,
     workflowJobRunBackendId: backendIds.workflowJobRunBackendId,
-    name,
+    name: name,
     version: 4
   }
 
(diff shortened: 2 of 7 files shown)


packages/artifact/src/internal/upload/upload-artifact.ts
```diff
@@ -28,11 +30,14 @@ export async function uploadArtifact(
     }
   }
 
-  // TODO - Implement upload functionality
+  const zipUploadStream = await createZipUploadStream(zipSpecification)
+
+  const sasURL = '' // TODO replace with actual URL
+  const uploadResult = await uploadZipToBlobStorage(sasURL, zipUploadStream)
```
bethanyj28, line 36: We should check whether the upload succeeded here, otherwise it might cause an error while trying to access `uploadResult.uploadSize` if an error occurs:
https://github.com/actions/toolkit/blob/1573d74a9f68f9b53cbac68264702ae9b9c9afb0/packages/artifact/src/internal/upload/blob-upload.ts#L51-L56

packages/artifact/src/internal/upload/zip.ts
```diff
@@ -0,0 +1,95 @@
...
+  core.info(error)
+
+  throw new Error('An error has occurred during zip creation for the artifact')
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const zipWarningCallback = (error: any): void => {
+  if (error.code === 'ENOENT') {
+    core.warning('ENOENT warning during artifact zip creation.')
+    core.info(error)
```
bethanyj28, line 80: Also curious you're reasoning behind putting these in info logs?

packages/artifact/src/internal/shared/config.ts
```diff
@@ -1,3 +1,9 @@
+// Used for controlling the highWaterMark value of the zip that is being streamed
+// The same value is used as the chunk size that is use during upload to blob storage
+export function getUploadChunkSize(): number {
+  return 8 * 1024 * 1024 // 8 MB Chunks
```
yacaovsnc, line 4: We should make sure we use [append blocks ](https://learn.microsoft.com/en-us/azure/storage/blobs/scalability-targets#scale-targets-for-blob-storage)for artifacts, as block blobs are limited to 4GiB.  

Also the chunk size supported by Azure is 4MiB.

packages/artifact/src/internal/upload/zip.ts
konradpabjan: The full errors tend to have more information. Someone will inevitably want to know "what path caused the ENOENT error, and the full error message should have that information". Doesn't hurt to include it"

bethanyj28: Sorry! I meant as opposed to a debug log - definitely agree it should be included!

bethanyj28 (approved):

packages/artifact/src/internal/upload/zip.ts
konradpabjan: Debug logs don't shown up by default. Users would have to re-run with the flag which could be a bit annoying. A warning like this should have the full information displayed by default

konradpabjan:
Postman-echo started redirecting `http` to `https` so now a bunch of unrelated tests are failing in `http-client`. Will merge in for now and take a look later as this is unrelated

davhdavh:
Will this change also include deduplication before upload? It is quite silly to have to upload 400mb when the previous build already uploaded 398mb that is exactly the same

PR #1494, opened 2023-08-15
[Artifacts] Save md5 hash for each artifact upload

konradpabjan:

This builds up on the work that was done in https://github.com/actions/toolkit/pull/1479, https://github.com/actions/toolkit/pull/1481, https://github.com/actions/toolkit/pull/1486 and https://github.com/actions/toolkit/pull/1487 and https://github.com/actions/toolkit/pull/1488

We want to save a hash of the artifact upload for integrity. Unfortunately the response from `uploadStream` does not have an MD5 hash value. There is a crc64 value in the raw headers but it doesn't appear to be officially documented in the azure blob NPM package so we're not going to try to use it as a result.

We can't also call `blobClient.properties` to get a hash after the upload finished since the token only has permissions to create and write, not read.

I'm using a solution that I found here. We pipe the zip stream to a crypto stream on the side and after the upload finishes we can read the hash and save if with the finalize call. When we get to download artifact we can then return this value and do integrity checks and a host of other things.

Here is an example run with the code running E2E after doing `npm link` and setting everything up with a fork that I've been using for testing: https://github.com/bbq-beets/testing-artifacts-v4/actions/runs/5869253502/job/15913849975

image

I verified afterwards if the hash of the zip is correct by downloading the artifact and manually inspecting it and things work 👍

### DIFF
--- a/packages/artifact/src/internal/upload/blob-upload.ts
+++ b/packages/artifact/src/internal/upload/blob-upload.ts
@@ -72,7 +72,7 @@ export async function uploadZipToBlobStorage(
     core.info(`MD5 hash of uploaded artifact zip is ${md5Hash}`)
 
   } catch (error) {
-    core.info(`Failed to upload artifact zip to blob storage, error: ${error}`)
+    core.warning(`Failed to upload artifact zip to blob storage, error: ${error}`)
     return {
       isSuccess: false
     }
--- a/packages/artifact/src/internal/upload/blob-upload.ts
+++ b/packages/artifact/src/internal/upload/blob-upload.ts
@@ -3,6 +3,8 @@ import {TransferProgressEvent} from '@azure/core-http'
 import {ZipUploadStream} from './zip'
 import {getUploadChunkSize} from '../shared/config'
 import * as core from '@actions/core'
+import * as crypto from 'crypto'
+import * as stream from 'stream'
 
 export interface BlobUploadResponse {
   /**
@@ -14,6 +16,11 @@ export interface BlobUploadResponse {
    * The total reported upload size in bytes. Empty if the upload failed
    */
   uploadSize?: number
+
+  /**
+   * The MD5 hash of the uploaded file. Empty if the upload failed
+   */
+  md5Hash?: string
 }
 
 export async function uploadZipToBlobStorage(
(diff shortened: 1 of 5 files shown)


konradpabjan:
Failing tests are unrelated, just like on my last PR. See https://github.com/actions/toolkit/pull/1488#issuecomment-1673790699

packages/artifact/src/internal/upload/upload-artifact.ts
```diff
@@ -77,30 +83,50 @@ export async function uploadArtifact(
...
     workflowRunBackendId: backendIds.workflowRunBackendId,
     workflowJobRunBackendId: backendIds.workflowJobRunBackendId,
     name: name,
     size: uploadResult.uploadSize!.toString()
-  })
+  }
+
+  if (uploadResult.md5Hash) {
+    finalizeArtifactReq.hash = StringValue.create({
+      value: `md5: ${uploadResult.md5Hash!}`
```
robherley, line 106: very nit: remove the space
```suggestion
      value: `md5:${uploadResult.md5Hash!}`
```

packages/artifact/src/internal/upload/upload-artifact.ts
```diff
@@ -77,30 +83,50 @@ export async function uploadArtifact(
...
 
+  const artifactId = parseInt(finalizeArtifactResp.artifactId)
+  core.info(
+    `Artifact ${name}.zip successfully finalized. Artifact ID ${artifactId}`
+  )
+
   return {
     success: true,
     size: uploadResult.uploadSize,
-    id: parseInt(finalizeArtifactResp.artifactId) // TODO - will this be a problem due to the id being a bigint?
```
robherley, line 104: re: this comment, wonder if we should wrap it in a `BigInt(finalizeArtifactResp.artifactId)`? @bethanyj28 i think changed the proto compiler options though to use strings

packages/artifact/src/internal/upload/blob-upload.ts
```diff
@@ -41,13 +48,29 @@ export async function uploadZipToBlobStorage(
...
       bufferSize,
       maxBuffers,
       options
     )
+
+    core.info('Finished uploading artifact content to blob storage!')
+
+    hashStream.end()
+    md5Hash = hashStream.read() as string
+    core.info(`MD5 hash of uploaded artifact zip is ${md5Hash}`)
```
bethanyj28, line 72: Out of curiosity, will printing the hash get a little noisy? Maybe we could make this a debug log?

packages/artifact/src/internal/upload/upload-artifact.ts
```diff
@@ -39,7 +43,9 @@ export async function uploadArtifact(
   // get the IDs needed for the artifact creation
   const backendIds = getBackendIdsFromToken()
   if (!backendIds.workflowRunBackendId || !backendIds.workflowJobRunBackendId) {
-    core.warning(`Failed to get backend ids`)
+    core.warning(
+      `Failed to get the necessary backend ids which are necessary to create the artifact`
```
bethanyj28, line 47:
```suggestion
      `Failed to get the necessary backend ids which are required to create the artifact`
```
Nit on wording

packages/artifact/src/internal/upload/upload-artifact.ts
konradpabjan: Done!

bethanyj28 (approved):
Functionality looks good! Just had some logging nits!

packages/artifact/src/internal/upload/blob-upload.ts
konradpabjan: I'd prefer to at last initially keep this in the logs. We're not going to immediately expose the hash via our REST APIs for artifacts unless we do more work so there is no real way for users to get this information unless we log it. When we update expose this potentially in our APIs then we could make this debug

robherley (approved):

AnnAngela:
@konradpabjan Sorry for commenting here, but I found out that this pr introduced a new dependency "crypto" in <https://github.com/actions/toolkit/pull/1494/files#diff-3a2287638a98ab8c93525fe63e6e7e35a2e3be09bd2e4db80173c68ed83c72f8R47>, which is [marked as deprecated and hold nothing for use](https://www.npmjs.com/package/crypto?activeTab=code). Maybe this dependency can be removed?

PR #1495, opened 2023-08-17
[Artifacts] @actions/artifact list artifact functionality + download interface setup

konradpabjan:
Part of https://github.com/github/c2c-actions-checks/issues/1562

## API Contracts

Simplifying much of the existing logic by having a single shared `interfaces.ts` file that defines what each function accepts and returns.

Adding the skeleton structure for `downloadArtifact` and `getArtifact`

## List Artifacts

Adding a new function called `listArtifacts` that paginates through all artifacts for a workflow run and lists at most 1000 artifacts.

Adding a new verify job that runs after all 3 separate uploads finish

image

The verify job calls the `listArtifacts` method call and we make sure that all 3 of the artifacts that were recently uploaded are present. See the output of this recent run: https://github.com/actions/toolkit/actions/runs/5894413718?pr=1495

image


packages/artifact/src/internal/find/retry-options.ts
```diff
@@ -0,0 +1,44 @@
+import * as core from '@actions/core'
+import {OctokitOptions} from '@octokit/core/dist-types/types'
+import {RequestRequestOptions} from '@octokit/types'
+
+export type RetryOptions = {
+  doNotRetry?: number[]
+  enabled?: boolean
+}
+
+export function getRetryOptions(
```
konradpabjan, line 10: Copied this from here: https://github.com/actions/github-script/blob/main/src/retry-options.ts

This warrants installing `@octokit/core`, `@octokit/plugin-request-log` and `@octokit/plugin-retry` which is why you see those new additions to `package.json`

robherley (approved):
👍

packages/artifact/src/internal/shared/interfaces.ts
```diff
@@ -0,0 +1,123 @@
...
+ *                                                                           *
+ *                           DownloadArtifact                                *
+ *                                                                           *
+ *****************************************************************************/
+export interface DownloadArtifactResponse {
+  /**
+   * If the artifact download was successful
+   */
+  success: boolean
+}
```
robherley, line 83: we may want to have the download path like: https://github.com/actions/toolkit/blob/91d3933eb52b351f437151400a88ba7d57442a9b/packages/artifact/src/internal/download-response.ts#L1-L11

but we'll wait until implementation

bethanyj28 (approved):
🎨

packages/artifact/src/internal/shared/interfaces.ts
konradpabjan: Yeah that would be useful, but it can be added later when download gets fully implemented

alec-drw:
Hello, is there any clarification as to why this is not supported on GHES?

PR #1503, opened 2023-08-22
Get a single artifact by name and download to `GITHUB_WORKSPACE`

bethanyj28:
Building off of @robherley's PR, this adds functionality to `getArtifact` and defaults to `GITHUB_WORKSPACE` if a `path` is not provided. Additionally, there are some linting updates.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,4 +3,3 @@ packages/*/node_modules/
 packages/*/lib/
 packages/glob/__tests__/_temp
 packages/*/src/generated/*/
-packages/artifact/src/internal/shared/user-agent.ts
--- a/.eslintignore
+++ b/.eslintignore
@@ -3,3 +3,4 @@ packages/*/node_modules/
 packages/*/lib/
 packages/glob/__tests__/_temp
 packages/*/src/generated/*/
+packages/artifact/src/internal/shared/user-agent.ts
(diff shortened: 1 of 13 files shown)


packages/artifact/src/internal/find/list-artifacts.ts
bethanyj28, line 1: Changes in this file were due to prettier/linting

packages/artifact/src/internal/upload/blob-upload.ts
bethanyj28, line 1: Changes in this file were all for linting.

.eslintignore
```diff
@@ -3,3 +3,4 @@ packages/*/node_modules/
 packages/*/lib/
 packages/glob/__tests__/_temp
 packages/*/src/generated/*/
+packages/artifact/src/internal/shared/user-agent.ts
```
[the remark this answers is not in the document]
bethanyj28: Add ignore on the require line.

packages/artifact/src/internal/shared/user-agent.ts
```diff
@@ -1,4 +1,5 @@
-var packageJson = require('../../../package.json')
+// eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
```
bethanyj28, line 1: Attempted to use import for user-agent.ts, but it seems like require is our best bet if we don't want to change a bunch of config: https://stackoverflow.com/questions/50822310/how-to-import-package-json-in-typescript

robherley (approved):

ulrichb:
@bethanyj28

As now `getArtifact()` simply uses `GET /repos/{owner}/{repo}/actions/runs/{run_id}/artifacts{?name}`, i.e. simply the REST API, does this mean that this API can also be used _during the workflow run_ when later using `actions/upload-artifact@v4` (which is currently not possible with v3 ...)?


I'm asking because otherwise I wouldn't understand how artifact sharing between jobs would work with `actions/download-artifact@v4` ...

PR #1518, opened 2023-09-03
ci: Run tests on Node.js v16 as well as v20

legobeat:
- Run build&test job on v16 to check for compatibility breakage
- Read nodejs versions for other jobs from `.nvmrc`
- Add `.nvmrc` with version set to `v20`

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -17,6 +17,7 @@ jobs:
     strategy:
       matrix:
         runs-on: [ubuntu-latest, macos-latest, windows-latest]
+        node-version: [16.x, 20.x]
       fail-fast: false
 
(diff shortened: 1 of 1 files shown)


Kryslynn93 (approved):

eggyhead:
:wave: Hi! Thanks for the contribution. I like the idea of introducing an `.nvmrc` file, but since Node16 is [deprecated](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/) I don’t see any value in this addition, especially since the CI checks are [failing](https://github.com/actions/toolkit/actions/runs/8781652028/job/24162393483?pr=1518). Could you update to include only the `.nvmrc` changes or close this PR?

PR #1522, opened 2023-09-05
Update 403 error message for GHES

bethanyj28:
The current error message for 403s doesn't make sense for GHES customers and doesn't provide much help. This aims to update the error to be more helpful.


konradpabjan (approved):
🚀

PR #1547, opened 2023-10-05
Add function to return proxy agent dispatcher for compatibility with latest `octokit` packages

takost:
Add function to return proxy agent dispatcher for compatibility with latest `octokit` packages.
[Octokit docs](https://github.com/octokit/octokit.js#proxy-servers-nodejs-only)

### DIFF
--- a/packages/http-client/__tests__/proxy.test.ts
+++ b/packages/http-client/__tests__/proxy.test.ts
@@ -3,6 +3,7 @@
 import * as http from 'http'
 import * as httpm from '../lib/'
 import * as pm from '../lib/proxy'
+import { ProxyAgent } from "undici";
 // eslint-disable-next-line @typescript-eslint/no-var-requires, @typescript-eslint/no-require-imports
 const proxy = require('proxy')
 
@@ -13,7 +14,7 @@ const _proxyUrl = 'http://127.0.0.1:8080'
 describe('proxy', () => {
   beforeAll(async () => {
     // Start proxy server
-    _proxyServer = proxy()
+    _proxyServer = proxy.createProxy()
     await new Promise<void>(resolve => {
       const port = Number(_proxyUrl.split(':')[2])
       _proxyServer.listen(port, () => resolve())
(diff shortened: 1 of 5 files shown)


packages/http-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/http-client",
-  "version": "2.1.1",
+  "version": "3.0.0",
```
pje, line 3: Why does this need to be a major version bump? Are there any breaking changes here?

packages/http-client/src/index.ts
```diff
@@ -701,6 +714,44 @@ export class HttpClient {
...
+
+    // if agent is already assigned use that agent.
+    if (proxyAgent) {
+      return proxyAgent
+    }
+
+    const usingSsl = parsedUrl.protocol === 'https:'
+
+    // This is `useProxy` again, but we need to check `proxyURl` directly for TypeScripts's flow analysis.
+    if (proxyUrl && proxyUrl.hostname) {
```
pje, line 732: Any reason not to just use optional chaining here?

```suggestion
    if (proxyUrl?.hostname) {
```

pje: Ah, actually `proxyUrl` is defined as a **non-nullable** `URL`, so you don't even need to check it, it's guaranteed by the type system to be there. 

```suggestion
    if (proxyUrl.hostname) {
```

pje (approved):
I was unfamiliar with undici, so I had to do some research :)

Overall I'm happy to ✅ — my only question is really on the necessity of bumping the major version of the package, but I'll defer to your  judgment there!

packages/http-client/src/index.ts
takost: Resolved

packages/http-client/package.json
```diff
@@ -39,10 +39,13 @@
...
+    "@types/node": "20.7.1",
     "@types/tunnel": "0.0.3",
-    "proxy": "^1.0.1"
+    "proxy": "^2.1.1",
+    "@types/proxy": "^1.0.1"
   },
   "dependencies": {
-    "tunnel": "^0.0.6"
+    "tunnel": "^0.0.6",
+    "undici": "^5.25.4"
```
Jason3S, line 49: @takost,

When updating today, I noticed that size of my action bloated by 15k lines because `undici` was included as a dependency.

packages/http-client/__tests__/proxy.test.ts
```diff
@@ -3,6 +3,7 @@
 import * as http from 'http'
 import * as httpm from '../lib/'
 import * as pm from '../lib/proxy'
+import {ProxyAgent} from 'undici'
```
mgrybyk, line 6: this line causes the "lightweight" client size to increase 3x times by 1Mb of code.

packages/http-client/package.json
grybykm: the same for me. I've raised a ticket for the issue https://github.com/actions/toolkit/issues/1621

Issue #1542, opened 2023-09-27
Support the latest Octokit in `@actions/github`


joshmgross:
The latest Octokit only supports Node 18 and higher - https://github.com/orgs/octokit/discussions/44

Actions currently supports Node 16 and 20 - https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

We should update `@actions/github` to the latest version of Octokit and drop support for Node 16 so that actions can transition to Node 20. This will be a breaking change.

joshmgross:
This should be available with v6 from https://github.com/actions/toolkit/pull/1553

https://github.com/actions/toolkit/blob/main/packages/github/RELEASES.md#600

Thanks @takost!

PR #1553, opened 2023-10-09
Update dependencies in github package

takost:
Update octokit dependencies in github package.
According to octokit [docs](https://github.com/octokit/octokit.js#proxy-servers-nodejs-only) to use proxy we need to start using custom fetch function. This functionality was implemented in new `@actions/http-client`. 
As usage of fetch function is new logic, and octokit update has breaking changes, I bumped major version of github package. After this update github package will be able to use only with node>=18

I tested this change with `checkout` action: https://github.com/actions/checkout/pull/1504

### DIFF
--- a/packages/github/package.json
+++ b/packages/github/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@actions/github",
-  "version": "5.1.1",
+  "version": "6.0.0",
   "description": "Actions github lib",
   "keywords": [
     "github",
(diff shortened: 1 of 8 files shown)


packages/github/package.json
```diff
@@ -38,12 +38,12 @@
     "url": "https://github.com/actions/toolkit/issues"
   },
   "dependencies": {
-    "@actions/http-client": "^2.0.1",
-    "@octokit/core": "^3.6.0",
-    "@octokit/plugin-paginate-rest": "^2.17.0",
-    "@octokit/plugin-rest-endpoint-methods": "^5.13.0"
+    "@actions/http-client": "^2.2.0",
+    "@octokit/core": "^4.2.4",
```
joshmgross, line 42: Can we update to v5 and drop support for Node 16?

https://github.com/octokit/core.js/releases/tag/v5.0.0

https://github.com/actions/toolkit/issues/1542

takost: I u[dated to latest octokit

Issue #1555, opened 2023-10-10
Upgrading from `@actions/github` V5 -> V6 caused TypeErrors within `getOcktokit()` 


Zidious:
### What happened?

Version 5 of `@actions/github`, I was able to do this within my tests: 

```ts
import { getOctokit } from '@actions/github'
import type { Endpoints } from '@octokit/types'
import sinon from 'sinon'

const octokit = getOctokit('token')

const stubLabelList = () => {
	return sinon.stub(octokit.rest.issues, 'listLabelsForRepo').resolves({
		data: [MOCK_LIST_LABELS],
		status: 200
		} as Endpoints["GET /repos/{owner}/{repo}/labels"]["response"])
	}
}

const github = {
	getOctokit: () => octokit
}

// pass github to other test functions etc 

```

As of version 6, I've had to do something like: 

```ts
import { getOctokit } from '@actions/github'
import type { Endpoints } from '@octokit/types'
import sinon from 'sinon'

const octokit = getOctokit('token')

const github = {
	getOctokit: () => {
		...octokit, 
		rest: { 
			issues: {
				listLabelsForRepo: () => {
					return {
						data: [MOCK_LIST_LABELS],
						status: 200
					} as Endpoints["GET /repos/{owner}/{repo}/labels"]["response"])
				}
			}
		} 
	}
}
```

I noticed this was a change in way the endpoints are generated and is now done via a proxy as of https://github.com/octokit/plugin-rest-endpoint-methods.js/pull/622, could this of introduced unwanted side effects or is this the intended behaviour? 

### Versions

@actions/github: v6
NodeJS: 18

### Relevant log output

```shell
TypeError: Cannot destructure property ‘decorations’ of ‘endpointMethodsMap.get(...).get(...)’ as it is undefined.
```

PR #1558, opened 2023-10-12
fix(packages/github): upstream `@octokit/plugin-rest-endpoint-methods` fix

Zidious:
Closes: https://github.com/actions/toolkit/issues/1555


Zidious:
Any update on this?

PR #1563, opened 2023-10-16
Use sha256 instead of md5 for artifact v4 integrity hash

robherley:
Recent MSFT guidance is to not use MD5, even for integrity. This updates the upload response to use SHA256 instead.

Related:
- https://github.com/github/actions-results-team/issues/1907

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -185,12 +185,12 @@
       }
     },
     "node_modules/@babel/generator": {
-      "version": "7.22.10",
-      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.22.10.tgz",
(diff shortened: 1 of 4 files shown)


konradpabjan (approved):

yacaovsnc (approved):

Shurikenttt:
- name: Tar files
    run: tar -cvf my_files.tar /path/to/my/directory

  - name: Upload Artifact
    uses: actions/upload-artifact@v3
    with:
      name: my-artifact
      path: my_files.tar

Gitanito333:
packages/artifact/src/internal/upload/blob-upload.ts

PR #1572, opened 2023-10-26
Make sure RequestOptions.keepAlive is applied properly on node20 runtime

teatimeguest:
Starting with Node.js v19, http(s) `globalAgent` now uses `keep-alive` by default. The current `HttpClient` implementation uses `globalAgent` when `RequestOptions.keepAlive` is set to `false` (default), which is fine for `node16` runtime but incorrect for `node20`.

This PR fixes `HttpClient` to use a custom agent regardless of `RequestOptions.keepAlive` so that the option works correctly on both runtimes.

- [Node.js v19.9.0 documentation | `http.globalAgent`](https://nodejs.org/docs/latest-v19.x/api/http.html#httpglobalagent)
- https://github.com/nodejs/node/pull/43522

### DIFF
--- a/packages/http-client/__tests__/keepalive.test.ts
+++ b/packages/http-client/__tests__/keepalive.test.ts
@@ -11,6 +11,12 @@ describe('basics', () => {
     _http.dispose()
   })
 
+  it.each([true, false])('creates Agent with keepAlive %s', keepAlive => {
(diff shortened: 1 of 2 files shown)


eregon:
It would be great to merge this as it would fix actions using the cache package taking an extra 2 minutes if they don't explicitly `process.exit()`: https://github.com/ruby/setup-ruby/issues/543#issuecomment-1793608370

PR #1584, opened 2023-11-20
Increase Artifact v4 upload speed

robherley:
Some additional tweaks to help upload speeds in v4:
- Adjust concurrency based on number of CPUs (following [azcopy's formula](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-optimize))
- Sets the `highWaterMark` for archiver to use the same other streams.
- Ability to tweak compression levels (zlib) to fit use case.

We'll expose a new input called `compression-level` on artifact upload.

For a random 4GB binary file with (compression-level: 0), this brought the upload time down from 3m+ to ~48s.

ℹ️ This also fixes a small bug with URL construction for the twirp client.

### DIFF
--- a/packages/artifact/src/internal/upload/zip.ts
+++ b/packages/artifact/src/internal/upload/zip.ts
@@ -5,6 +5,8 @@ import {createReadStream} from 'fs'
 import {UploadZipSpecification} from './upload-zip-specification'
 import {getUploadChunkSize} from '../shared/config'
 
+export const DEFAULT_COMPRESSION_LEVEL = 6
+
 // Custom stream transformer so we can set the highWaterMark property
 // See https://github.com/nodejs/node/issues/8855
 export class ZipUploadStream extends stream.Transform {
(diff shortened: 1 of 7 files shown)


yacaovsnc (approved):

TingluoHuang (approved):

packages/artifact/src/internal/upload/zip.ts
```diff
@@ -5,6 +5,8 @@ import {createReadStream} from 'fs'
 import {UploadZipSpecification} from './upload-zip-specification'
 import {getUploadChunkSize} from '../shared/config'
 
+export const DEFAULT_COMPRESSION_LEVEL = 6
```
konradpabjan, line 8: Nit, potentially you can put the default value into `packages/artifact/src/internal/shared/config.ts` but no biggie

konradpabjan (approved):

packages/artifact/src/internal/upload/zip.ts
robherley: Since it's constant and only used for the zip method in this file, I'm just going to leave it here for now 👌

PR #1587, opened 2023-11-27
fix #1579: getCacheVersion mutates arguments

### LINKED ISSUE #1579: @actions/cache mutates the `paths` argument.
**Describe the bug**
https://github.com/actions/toolkit/blob/main/packages/cache/src/internal/cacheHttpClient.ts#L83
This line means that the changes to the array change the array the user passes.
This breaks code such as
```js
import * as cache from "@actions/cache";

const paths = ["some/paths"];
const key = "a-key";

// This calls `getCacheVersion` internally and mutates `paths`
const cacheKey = await cache.restoreCache(paths, key);
if (cacheKey) {
  core.info(`Cache restored from key: ${cacheKey}`);
  return;
}

// At this point the variable `paths = ["some/paths", "zstd-without-long", "1.0"]`


const savedCacheKey = await cache.saveCache(paths, key);
// This means when `saveCache` calls `getCacheVersion` it was actually calculated off of the final array value: `["some/paths", "zstd-without-long", "1.0", "zstd-without-long", "1.0"]`
}
```

This means that we run into a version mismatch when restoring the cache in another step later on.

**Expected behavior**
Inputs to functions don't get mutated and break intended functionality of the library.

**Screenshots**
image
This is following the hashing to create versions, with the inputs from our code, and you can see that `components2` with the compression and version salt in twice matches the version of the created caches.

Smeb:
Ran into the issue described in https://github.com/actions/toolkit/issues/1579, this fixes it.

### DIFF
--- a/packages/cache/__tests__/cacheHttpClient.test.ts
+++ b/packages/cache/__tests__/cacheHttpClient.test.ts
@@ -5,6 +5,12 @@ import {DownloadOptions, getDownloadOptions} from '../src/options'
 
 jest.mock('../src/internal/downloadUtils')
 
+test('getCacheVersion does not mutate arguments', async () => {
+  const paths = ['node_modules']
+  getCacheVersion(paths, undefined, true)
+  expect(paths).toEqual(['node_modules'])
+})
+
 test('getCacheVersion with one path returns version', async () => {
   const paths = ['node_modules']
   const result = getCacheVersion(paths, undefined, true)
--- a/packages/cache/src/internal/cacheHttpClient.ts
+++ b/packages/cache/src/internal/cacheHttpClient.ts
@@ -80,7 +80,7 @@ export function getCacheVersion(
   compressionMethod?: CompressionMethod,
   enableCrossOsArchive = false
 ): string {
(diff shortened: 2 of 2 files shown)


Smeb:
It would be great to get this addressed as we use this internally. From my side linting and checks are passing.

BrandonLWhite (approved):

ramblingenzyme (approved):

bethanyj28 (approved):
Thanks for the contribution! LGTM

bethanyj28:
Actually seeing this was fixed in https://github.com/actions/toolkit/pull/1378 - so going to close. Thank you!

Smeb:
Ok - what about including the test? I see the fix in #1378 does not add any tests? I can modify this patch in that case.

bethanyj28:
@Smeb - keeping the test would be fantastic! That would be a big help 🙏🏻

ramblingenzyme:
@bethanyj28 can we get https://github.com/actions/toolkit/issues/1579 closed off as fixed too?

PR #1593, opened 2023-12-05
Consistent error behavior for Artifact methods

robherley:
Partially addresses:
- https://github.com/github/actions-results-team/issues/1805

For our artifact operations, we weren't being consistent on how we were surfacing errors. Sometimes we'd resolve with a response object of `{success:false}` or we'd reject via an error throw.

This PR changes the behavior so that the errors consistently throw/reject for promises to "bubble up" and don't require a bunch of `if(success) { }` logic. Any common errors are exposed that support additional debugging information.

Also, the workflow tests used for this package were updated to reflect these changes.

### DIFF
--- a/.github/workflows/artifact-tests.yml
+++ b/.github/workflows/artifact-tests.yml
@@ -24,10 +24,10 @@ jobs:
 
     steps:
     - name: Checkout
-      uses: actions/checkout@v3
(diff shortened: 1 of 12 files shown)


bethanyj28 (approved):
LGTM!

PR #1596, opened 2023-12-06
Cleanup artifact handlers hanging node process

robherley:
Fixes a bug where an error response causes the actions to hang for ~2 min.

Using [`why-is-node-running`](https://www.npmjs.com/package/why-is-node-running) we were able to track down the `ZLIB`, `TCPWRAP`, and `TLSWRAP` handlers clogging things up.

tl;dr

1. ZIP stream was being created way too early, we encounter an error & throw (reject) and the stream just hangs around.
2. In our retry http wrapper, we weren't reading the body on the request if it's a non-200 status code. For some reason (more investigation needed) this keeps things going until the keep-alive is met for the request? (or some other 2 minute timeout)

### DIFF
--- a/packages/artifact/src/internal/shared/artifact-twirp-client.ts
+++ b/packages/artifact/src/internal/shared/artifact-twirp-client.ts
@@ -62,7 +62,7 @@ class ArtifactHttpClient implements Rpc {
       const {body} = await this.retryableRequest(async () =>
         this.httpClient.post(url, JSON.stringify(data), headers)
       )
-      
+
       return JSON.parse(body)
     } catch (error) {
       throw new Error(`Failed to ${method}: ${error.message}`)
@@ -71,7 +71,7 @@ class ArtifactHttpClient implements Rpc {
 
   async retryableRequest(
     operation: () => Promise<HttpClientResponse>
-  ): Promise<{response: HttpClientResponse, body: string}> {
+  ): Promise<{response: HttpClientResponse; body: string}> {
     let attempt = 0
     let errorMessage = ''
     while (attempt < this.maxAttempts) {
--- a/packages/artifact/src/internal/shared/artifact-twirp-client.ts
+++ b/packages/artifact/src/internal/shared/artifact-twirp-client.ts
@@ -80,11 +80,10 @@ class ArtifactHttpClient implements Rpc {
       try {
         const response = await operation()
         const statusCode = response.message.statusCode
-        const raw = await response.readBody()
-        const body = JSON.parse(raw)
+        const body = await response.readBody()
         debug(`[Response] - ${response.message.statusCode}`)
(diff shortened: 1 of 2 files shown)


packages/artifact/src/internal/shared/artifact-twirp-client.ts
```diff
@@ -80,11 +80,13 @@ class ArtifactHttpClient implements Rpc {
       try {
         const response = await operation()
         const statusCode = response.message.statusCode
-        debug(`[Response] ${response.message.statusCode}`)
-        debug(JSON.stringify(response.message.headers, null, 2))
+        const body = await response.readBody()
+        debug(`[Response] - ${response.message.statusCode}`)
+        debug(`Headers: ${JSON.stringify(response.message.headers, null, 2)}`)
+        debug(`Body: ${body}`)
```
mattiaerre, line 86: I believe we may need to remove these lines.

packages/artifact/src/internal/shared/artifact-twirp-client.ts
```diff
@@ -80,11 +80,13 @@ class ArtifactHttpClient implements Rpc {
...
 
         if (this.isSuccessStatusCode(statusCode)) {
-          return response
+          return {response, body}
```
mattiaerre, line 89: This is changing the return type; are there any tests failing?

packages/artifact/src/internal/shared/artifact-twirp-client.ts
```diff
@@ -80,11 +80,13 @@ class ArtifactHttpClient implements Rpc {
       try {
         const response = await operation()
         const statusCode = response.message.statusCode
-        debug(`[Response] ${response.message.statusCode}`)
-        debug(JSON.stringify(response.message.headers, null, 2))
+        const body = await response.readBody()
```
mattiaerre, line 83: noice! that's it 👍

mattiaerre (approved):

packages/artifact/src/internal/shared/artifact-twirp-client.ts
robherley: I actually added them to assist in debugging customer workflows. If they run into an issue, they can turn on debug logs and this can potentially help figure out what's wrong

packages/artifact/src/internal/upload/upload-artifact.ts
```diff
@@ -73,6 +68,11 @@ export async function uploadArtifact(
     )
   }
 
+  const zipUploadStream = await createZipUploadStream(
+    zipSpecification,
+    options?.compressionLevel
+  )
+
```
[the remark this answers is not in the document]
robherley: Yea, we were initializing this wayyy above in the method before we were even consuming the stream

PR #1600, opened 2023-12-08
Improve error messages (duplicate artifacts; too many artifacts)

vmjoseph:
Adds a few catches on vague errors:
- 409s that are tied to Conflict errors (aka artifact cannot be uploaded due to duplicate name error) now displays a more user friendly message
- 400s that are tied to InvalidArgument errors (aka too many artifacts for 1 job) now displays a more user friendly message

### DIFF
--- a/packages/artifact/__tests__/artifact-http-client.test.ts
+++ b/packages/artifact/__tests__/artifact-http-client.test.ts
@@ -190,4 +190,51 @@ describe('artifact-http-client', () => {
     expect(mockHttpClient).toHaveBeenCalledTimes(1)
     expect(mockPost).toHaveBeenCalledTimes(1)
   })
+
+  it('should fail with a descriptive error', async () => {
+    // 409 duplicate error
+    const mockPost = jest.fn(() => {
+      const msgFailed = new http.IncomingMessage(new net.Socket())
+      msgFailed.statusCode = 409
+      msgFailed.statusMessage =
+        'Conflict: an artifact with this name already exists on the workflow run'
+      return {
+        message: msgFailed,
+        readBody: async () => {
+          return Promise.resolve(`{"ok": false}`)
+        }
+      }
+    })
+
+    const mockHttpClient = (
+      HttpClient as unknown as jest.Mock
+    ).mockImplementation(() => {
+      return {
+        post: mockPost
+      }
+    })
+    const client = internalArtifactTwirpClient({
+      maxAttempts: 5,
+      retryIntervalMs: 1,
+      retryMultiplier: 1.5
+    })
+    await expect(async () => {
+      await client.CreateArtifact({
+        workflowRunBackendId: '1234',
+        workflowJobRunBackendId: '5678',
+        name: 'artifact',
+        version: 4
+      })
+      await client.CreateArtifact({
+        workflowRunBackendId: '1234',
+        workflowJobRunBackendId: '5678',
+        name: 'artifact',
+        version: 4
+      })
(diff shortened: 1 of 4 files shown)


packages/artifact/src/internal/shared/artifact-twirp-client.ts
```diff
@@ -76,6 +76,9 @@ class ArtifactHttpClient implements Rpc {
     let errorMessage = ''
     while (attempt < this.maxAttempts) {
       let isRetryable = false
+      let isDuplicateArtifactName = false
+      let isTooManyArtifacts = false
```
bethanyj28, line 80: I wonder if these variables are needed - or if in the error block you could do something like:

```
if (this.isAlreadyExistsStatusCode(statusCode)) {
  throw new Error.....
}
```

I also thought the service today has an interesting approach to making a single function that maps error codes, which might consolidate some code: https://github.com/actions/toolkit/blob/e8f21993484c802107e8c4c3839e86b0b6e8954b/packages/artifact/src/internal/upload-http-client.ts#L93-L94

bethanyj28 (commented):
Also an additional test here might be nice <3 https://github.com/actions/toolkit/blob/main/packages/artifact/__tests__/artifact-http-client.test.ts

Wily96 (approved):
https://github.com/actions/toolkit/blob/vmjoseph%2Fdup-artifact-names/packages%2Fartifact%2Fsrc%2Finternal%2Fshared%2Fartifact-twirp-client.ts

packages/artifact/src/internal/shared/artifact-twirp-client.ts
```diff
@@ -84,11 +85,11 @@ class ArtifactHttpClient implements Rpc {
         debug(`[Response] - ${response.message.statusCode}`)
         debug(`Headers: ${JSON.stringify(response.message.headers, null, 2)}`)
         debug(`Body: ${body}`)
-
         if (this.isSuccessStatusCode(statusCode)) {
           return {response, body}
         }
-
+        // only parse the JSON if status is non-successful
+        errorResponse = JSON.parse(body).msg
```
robherley, line 92: meganit: let's reuse the existing `errorMessage`

move this down past line 94 and have something like:

```js
const responseMessage = JSON.parse(body).msg
if (responseMessage) {
  errorMessage = `${errorMessage}: ${responseMessage}`
}
```

packages/artifact/__tests__/artifact-http-client.test.ts
```diff
@@ -190,4 +190,51 @@ describe('artifact-http-client', () => {
...
     expect(mockPost).toHaveBeenCalledTimes(1)
   })
+
+  it('should fail with a descriptive error', async () => {
+    // 409 duplicate error
+    const mockPost = jest.fn(() => {
+      const msgFailed = new http.IncomingMessage(new net.Socket())
+      msgFailed.statusCode = 409
+      msgFailed.statusMessage =
+        'Conflict: an artifact with this name already exists on the workflow run'
```
robherley, line 200: _technically_ this `statusMessage` will be something like `OK` or `Not Found` from the headers

we'll never be changing this on the server response side

it would make more sense for this test to return a body with a `{"msg": "an artifact with this name already exists on the workflow run"}` to test the new logic we're adding 👍

vmjoseph: Gotcha! It should be `Conflict` to match the appropriate 409 error

robherley (approved):
👏

PR #1601, opened 2023-12-08
Replace unzipper with unzip-stream

jtamsut:
There is a [bug](https://github.com/ZJONSSON/node-unzipper/issues/286) in the `unzipper` library that causes files to become corrupted when streamed. This stems from [changes](https://github.com/nodejs/node/commit/654b7474e38e9a6272b73586023848109cfe358c) from Node's handling of streams. We are replacing `unzipper` with `unzip-stream` to as its a drop-in replacement with an identical API that doesn't suffer from the same issue.

Closes https://github.com/github/actions-results-team/issues/2096

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
         "@types/jest": "^29.5.4",
         "@types/node": "^20.5.7",
         "@types/signale": "^1.4.1",
+        "@types/unzip-stream": "^0.3.4",
         "concurrently": "^6.1.0",
         "eslint": "^8.0.1",
         "eslint-config-prettier": "^8.9.0",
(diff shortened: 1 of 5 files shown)


package-lock.json
```diff
@@ -9,6 +9,7 @@
         "@types/jest": "^29.5.4",
         "@types/node": "^20.5.7",
         "@types/signale": "^1.4.1",
+        "@types/unzip-stream": "^0.3.4",
```
robherley, line 12: this shouldn't be installed here, it should be within `packages/artifact`

packages/artifact/src/internal/download/download-artifact.ts
robherley, line 1: FYI the `unzip-stream` package doesn't have the `#promise` method so we'll need something like this at the end of `streamExtract`:

```ts
  return new Promise((resolve, reject) => {
    response.message
      .pipe(unzip.Extract({path: directory}))
      .on('close', resolve)
      .on('error', reject)
  })
```

robherley:
Closed in favor of: https://github.com/actions/toolkit/pull/1602

PR #1604, opened 2023-12-12
Support http proxy for artifact upload

robherley:
Closes: https://github.com/github/actions-results-team/issues/2094

Since the `BlobClient` from the Azure SDK does not use our `@actions/http-client`, it does not implicitly support http proxied network requests.

This PR adds support to automatically set the proxy options on the blob client if they are set in the environment.

### DIFF
--- a/packages/artifact/__tests__/upload-artifact.test.ts
+++ b/packages/artifact/__tests__/upload-artifact.test.ts
@@ -352,3 +352,46 @@ describe('upload-artifact', () => {
     expect(uploadResp).rejects.toThrow()
   })
 })
+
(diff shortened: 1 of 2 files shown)


robherley:
JK, the azure sdk for js already supports the same env vars we use: https://github.com/Azure/azure-sdk-for-js/blob/cf2e157f9af7551c6960af85e8cb4b45a4e60edb/sdk/core/ts-http-runtime/src/policies/proxyPolicy.ts#L111

I tested this with on a self hosted w/o these changes: 
image

16987814:
you are the responsability or this problems karem espinoza my kids are not for sale bitch

PR #1605, opened 2023-12-12
Better error message for artifact usage limits

robherley:
If customers encounter storage quota limits, they'll hit an error when trying to create artifacts.

This'll make the error more useful for the customers:
> Failed to CreateArtifact: Artifact storage quota has been hit. Unable to upload any new artifacts. Usage is recalculated every 6-12 hours.
> More info on storage limits: https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#calculating-minute-and-storage-spending

Let me know if there's any other context we'd like to add to the message 👍

### DIFF
--- a/packages/artifact/__tests__/artifact-http-client.test.ts
+++ b/packages/artifact/__tests__/artifact-http-client.test.ts
@@ -4,6 +4,7 @@ import {HttpClient} from '@actions/http-client'
 import * as config from '../src/internal/shared/config'
 import {internalArtifactTwirpClient} from '../src/internal/shared/artifact-twirp-client'
 import {noopLogs} from './common'
+import {NetworkError, UsageError} from '../src/internal/shared/errors'
 
 jest.mock('@actions/http-client')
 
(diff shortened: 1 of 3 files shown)


packages/artifact/__tests__/artifact-http-client.test.ts
```diff
@@ -257,9 +258,42 @@ describe('artifact-http-client', () => {
         name: 'artifact',
         version: 4
       })
-    }).rejects.toThrowError(
-      'Failed to CreateArtifact: Unable to make request: ENOTFOUND\nIf you are using self-hosted runners, please make sure your runner has access to all GitHub endpoints: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#communication-between-self-hosted-runners-and-github'
-    )
+    }).rejects.toThrowError(new NetworkError('ENOTFOUND').message)
```
robherley, line 261: `toThrowError` does a substring match, so this is a bit more cleaner IMO

packages/artifact/src/internal/shared/artifact-twirp-client.ts
```diff
@@ -64,42 +64,57 @@ class ArtifactHttpClient implements Rpc {
...
+      return body
     } catch (error) {
       throw new Error(`Failed to ${method}: ${error.message}`)
     }
   }
 
   async retryableRequest(
     operation: () => Promise<HttpClientResponse>
-  ): Promise<{response: HttpClientResponse; body: string}> {
+  ): Promise<{response: HttpClientResponse; body: object}> {
```
robherley, line 75: Just made this return an object since we're always going to be JSON parsing the response

bethanyj28 (approved):
🎉

PR #1594, opened 2023-12-06
@actions/artifact doc updates

robherley:
This starts the initial documentation for `@action/artifact` v2.

I'll be following up in another PR with autogenerated markdown docs for the functions, interfaces, etc.

Also, this change alters the export behavior of the default client.

Instead of:

```js
import * as artifact from '@actions/artifact'
const artifactClient = artifact.create()

artifact.getArtifact( /* ... */ )
```

Now there is a default export with the client already initialized:

```js
import artifact from '@actions/artifact'

artifact.getArtifact( /* ... */ )
```

PR #1607, opened 2023-12-13
Update artifact workflow tests

robherley:
These CI tests are used in addition to our jest unit-test to see how the node module behaves in an actions environment.
I've updated them slightly to include download tests too.

Follow up to this comment:
- https://github.com/actions/toolkit/pull/1594#discussion_r1417601789


konradpabjan (approved):
🚀

Issue #1606, opened 2023-12-13
The default export of @actions/artifact@2.0.0 is incorrect


AnnAngela:
**Describe the bug**

After upgrading to @actions/artifact@2.0.0, I found out that the default export is incorrect, see the reproduce code below:

**To Reproduce**

The reproduce code: https://github.com/AnnAngela/AnnAngela/blob/58d95c9ca456e2aa18943eb7e4f25d7a40406173/test.mjs

The output: https://github.com/AnnAngela/AnnAngela/actions/runs/7190162702/job/19582850361#step:5:1

**Expected behavior**
The below code should work fine:

```js
import artifact from "@actions/artifact";

await artifact.uploadArtifact("somefile", [somePath], rootPath); // => `TypeError: artifact.uploadArtifact is not a function` in @actions/artifact@2.0.0
```

zyd14:
I'm seeing issues with this too, preventing our builds from completing

robherley:
👋 I apologize for the confusion!

I think this might be something wrong with the configuration of how our TypeScript config is transpiling the modules. The default export works when using ts. I'll take a look to see how this can be resolved.

In the meantime, to avoid any possible breaking import changes, you can import and instanitate the default client (which is what the default export is):

```js
import { DefaultArtifactClient } from "@actions/artifact";
const artifactClient = new DefaultArtifactClient();
```

And you should be good to go 👍

robherley:
I've updated the readme:
- https://github.com/actions/toolkit/issues/1606

After talking with the team, we're not looking to turn this into a module yet. So we'll just avoid using default exports for now to avoid confusion between all the different module kinds 👍

AnnAngela:
@robherley Huge thanks for the clarification in the documentation (11a2dd3117d07bec18494a428098ea25aaead547), just a little suggestion: can you beautify the code in these two line:

https://github.com/actions/toolkit/blob/68f22927e727a60caff909aaaec1ab7267b39f75/packages/artifact/README.md?plain=1#L66

https://github.com/actions/toolkit/blob/68f22927e727a60caff909aaaec1ab7267b39f75/packages/artifact/README.md?plain=1#L69

robherley:
Our prettier config does not use bracket spacing unfortunately 😅 
- https://github.com/actions/toolkit/blob/68f22927e727a60caff909aaaec1ab7267b39f75/.prettierrc.json#L8

Since everything is clarified now, I'll close out the issue. Thanks for your feedback!

PR #1608, opened 2023-12-14
Update artifact module quick start

robherley:
The instructions for the default export don't work for es modules (`.mjs`) files, so lets instruct users to just import the default client and instantiate instead.

Moving to `"type": "module"` is not something we want to commit to yet for `@actions/artifact`

Addresses:
- https://github.com/actions/toolkit/issues/1606

### DIFF
--- a/packages/artifact/README.md
+++ b/packages/artifact/README.md
@@ -63,10 +63,16 @@ Import the module:
 
 ```js
 // ES6 module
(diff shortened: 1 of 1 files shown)


bethanyj28 (approved):

mattiaerre (approved):
👍

PR #1609, opened 2023-12-17
Include fs stats in zip archive

rotu:
The main driver of this is that artifact permissions are lost on upload:

https://github.com/actions/upload-artifact/issues/37
https://github.com/actions/upload-artifact/issues/38

But this will also keep file creation and modification timestamps to assist troubleshooting.

### DIFF
--- a/packages/artifact/__tests__/upload-artifact.test.ts
+++ b/packages/artifact/__tests__/upload-artifact.test.ts
@@ -8,6 +8,7 @@ import * as blobUpload from '../src/internal/upload/blob-upload'
 import {uploadArtifact} from '../src/internal/upload/upload-artifact'
 import {noopLogs} from './common'
 import {FilesNotFoundError} from '../src/internal/shared/errors'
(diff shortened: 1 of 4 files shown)


rmunn:
Looks like the `mode` parameter is also required in order to preserve Unix file permissions. The zip.append method that actions/toolkit uses is a tiny wrapper around the zip-stream package:

https://github.com/archiverjs/node-archiver/blob/0e0e9ceab65c35e8131701fd2b29ba3c28397385/lib/plugins/zip.js#L55

The `entry` method in zip-stream looks for the `mode` property on the incoming data:

https://github.com/archiverjs/node-zip-stream/blob/d2669793807f869da07b38e3e53a088e470702e2/index.js#L157

But nothing in zip-stream handles the `stats` property.

rmunn:
I've created https://github.com/actions/toolkit/pull/1723, which is essentially this PR plus passing the `mode` property into `zip.append`.

PR #1613, opened 2023-12-18
Update `http.client` to retry transient network hang ups 

SrRyan:
This pull request includes changes to the `packages/artifact/src/internal/download/download-artifact.ts` file to enhance the robustness of the artifact download process. The changes include the introduction of a retry mechanism, a timeout feature for the blob storage chunk response, and an explicit process exit after successful download. 

Here are the key changes:

* <a href="diffhunk://#diff-f584271373db702b060321f361b0c6fa27119ba146a37eae78a7e19184167cb5R42-R59">`packages/artifact/src/internal/download/download-artifact.ts`</a>: Added a retry mechanism in the `streamExtract` function. The code will now attempt to download the artifact up to 5 times before throwing an error. This change is designed to handle temporary network issues or other transient problems that may occur during the download process.

* <a href="diffhunk://#diff-f584271373db702b060321f361b0c6fa27119ba146a37eae78a7e19184167cb5R69-R97">`packages/artifact/src/internal/download/download-artifact.ts`</a>: Introduced a timeout feature in the `streamExtract` function. If the blob storage chunk does not respond within 30 seconds, the response stream is closed. This change helps to prevent the process from hanging indefinitely when waiting for a response.

* <a href="diffhunk://#diff-f584271373db702b060321f361b0c6fa27119ba146a37eae78a7e19184167cb5R195">`packages/artifact/src/internal/download/download-artifact.ts`</a>: Added an explicit process exit in the `downloadArtifactInternal` function after a successful artifact download. This ensures that the process terminates correctly after completing the download.

* <a href="diffhunk://#diff-f584271373db702b060321f361b0c6fa27119ba146a37eae78a7e19184167cb5R20">`packages/artifact/src/internal/download/download-artifact.ts`</a>: Imported the `clear` function from the `console` module. The purpose of this import is not clear from the diff and may require further clarification.


packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -47,11 +72,37 @@ async function streamExtract(url: string, directory: string): Promise<void> {
...
+      )
+    }
+    const timer = setTimeout(timerFn, timeout)
+
     response.message
+      .on('data', () => {
+        timer.refresh()
+      })
+      .on('error', (error: Error) => {
+        core.debug(
```
vmjoseph, line 90: Only show error message in debug mode - preventing extra noise

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -38,6 +38,31 @@ async function exists(path: string): Promise<boolean> {
 }
 
 async function streamExtract(url: string, directory: string): Promise<void> {
+  let retryCount = 0
+  while (retryCount < 5) {
+    try {
+      await streamExtractExternal(url, directory)
+      core.info(`Artifact downloaded successfully after ${retryCount} retries.`)
```
bethanyj28, line 45: Should probably remove this.

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -38,6 +38,31 @@ async function exists(path: string): Promise<boolean> {
...
+  }
+
+  throw new Error(`Artifact download failed after ${retryCount} retries.`)
+}
+
+export async function streamExtractExternal(
+  url: string,
+  directory: string
+): Promise<void> {
+  core.info(`Stream extract internal started`)
```
bethanyj28, line 64: Should probably remove this.

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -38,6 +38,31 @@ async function exists(path: string): Promise<boolean> {
...
+  while (retryCount < 5) {
+    try {
+      await streamExtractExternal(url, directory)
+      core.info(`Artifact downloaded successfully after ${retryCount} retries.`)
+      return
+    } catch (error) {
+      retryCount++
+      core.warning(
+        `Failed to download artifact after ${retryCount} retries due to ${error.message}. Retrying in 5 seconds...`
+      )
```
bethanyj28, line 51: Replace with core.debug

jest.config.js
```diff
@@ -8,4 +8,4 @@ module.exports = {
     '^.+\\.ts$': 'ts-jest'
   },
   verbose: true
-}
\ No newline at end of file
+}
```
vmjoseph, line 11: adding extra line for linter

konradpabjan (approved):

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -38,20 +38,65 @@ async function exists(path: string): Promise<boolean> {
...
   const client = new httpClient.HttpClient(getUserAgentString())
   const response = await client.get(url)
-
   if (response.message.statusCode !== 200) {
     throw new Error(
       `Unexpected HTTP response from blob storage: ${response.message.statusCode} ${response.message.statusMessage}`
     )
   }
 
+  const timeout = 30 * 1000 // 30 seconds
```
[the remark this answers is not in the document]
SrRyan: It's similar to how the SDK works if I recall but also we found in testing it was a good middle ground to validate the socket hadn't closed on us.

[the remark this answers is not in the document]
konradpabjan: If it's similar to how the SDK works then a link with a short explanation would be good for anyone randomly looking through the code for better understanding

konradpabjan (approved):

PR #1617, opened 2024-01-05
[WIP] tool-cache: in `tc.cacheDir()` and `tc.cacheFile()`, allow the user to opt-in to preserving timestamps when copying

DilumAluthge:
Replaces #1504 (closes #1504)

The value of `preserveTimestamp` defaults to `false`, so by default we preserve the existing behavior.

### DIFF
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -413,7 +413,7 @@ async function extractZipNix(file: string, dest: string): Promise<void> {
  * @param tool               tool name
  * @param version            version of the tool.  semver format
  * @param arch               architecture of the tool.  Optional.  Defaults to machine architecture
- # @param preserveTimestamps whether to preserve timestamps when copying.  Optional.  Defaults to true
+ * @param preserveTimestamps whether to preserve timestamps when copying.  Optional.  Defaults to true
  */
 export async function cacheDir(
   sourceDir: string,
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -483,7 +483,7 @@ export async function cacheFile(
   // anti-virus software having an open handle on a file.
   const destPath: string = path.join(destFolder, targetFile)
   core.debug(`destination file ${destPath}`)
-  await io.cp(sourceFile, destPath, {preserveTimestamps: preserveTimestamps})
+  await io.cp(sourceFile, destPath, {preserveTimestamps})
 
   // write .complete
   _completeToolPath(tool, version, arch)
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -409,19 +409,22 @@ async function extractZipNix(file: string, dest: string): Promise<void> {
 /**
  * Caches a directory and installs it into the tool cacheDir
  *
- * @param sourceDir    the directory to cache into tools
- * @param tool          tool name
- * @param version       version of the tool.  semver format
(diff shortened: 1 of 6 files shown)


packages/tool-cache/src/tool-cache.ts
```diff
@@ -435,7 +438,7 @@ export async function cacheDir(
   // due to anti-virus software having an open handle on a file.
   for (const itemName of fs.readdirSync(sourceDir)) {
     const s = path.join(sourceDir, itemName)
-    await io.cp(s, destPath, {recursive: true})
+    await io.cp(s, destPath, {recursive: true, preserveTimestamps: preserveTimestamps})
```
omus, line 441: Probably want the same treatment for `cacheFile` for consistency

DilumAluthge: Good call. Done - https://github.com/actions/toolkit/pull/1617/commits/5702324552e377d96f5e746f38b33c62e7f25045

packages/tool-cache/src/tool-cache.ts
```diff
@@ -477,7 +483,7 @@ export async function cacheFile(
   // anti-virus software having an open handle on a file.
   const destPath: string = path.join(destFolder, targetFile)
   core.debug(`destination file ${destPath}`)
-  await io.cp(sourceFile, destPath)
+  await io.cp(sourceFile, destPath, {preserveTimestamps: preserveTimestamps})
```
omus, line 486: Having `{x: x}` is equivalent to `{x}`:
```suggestion
  await io.cp(sourceFile, destPath, {preserveTimestamps})
```

omus (commented):
Looks good otherwise to me

packages/tool-cache/src/tool-cache.ts
DilumAluthge: TIL.

packages/tool-cache/src/tool-cache.ts
```diff
@@ -409,19 +409,22 @@ async function extractZipNix(file: string, dest: string): Promise<void> {
...
  *
- * @param sourceDir    the directory to cache into tools
- * @param tool          tool name
- * @param version       version of the tool.  semver format
- * @param arch          architecture of the tool.  Optional.  Defaults to machine architecture
+ * @param sourceDir          the directory to cache into tools
+ * @param tool               tool name
+ * @param version            version of the tool.  semver format
+ * @param arch               architecture of the tool.  Optional.  Defaults to machine architecture
+ # @param preserveTimestamps whether to preserve timestamps when copying.  Optional.  Defaults to true
```
IanButterworth, line 416:
```suggestion
 * @param preserveTimestamps whether to preserve timestamps when copying.  Optional.  Defaults to true
```

packages/tool-cache/src/tool-cache.ts
```diff
@@ -448,21 +451,24 @@ export async function cacheDir(
...
- * @param targetFile    the name of the file name in the tools directory
- * @param tool          tool name
- * @param version       version of the tool.  semver format
- * @param arch          architecture of the tool.  Optional.  Defaults to machine architecture
+ * @param sourceFile         the file to cache into tools.  Typically a result of downloadTool which is a guid.
+ * @param targetFile         the name of the file name in the tools directory
+ * @param tool               tool name
+ * @param version            version of the tool.  semver format
+ * @param arch               architecture of the tool.  Optional.  Defaults to machine architecture
+ # @param preserveTimestamps whether to preserve timestamps when copying.  Optional.  Defaults to true
```
IanButterworth, line 459:
```suggestion
 * @param preserveTimestamps whether to preserve timestamps when copying.  Optional.  Defaults to true
```

Issue #380, opened 2020-03-12
Delete Artifacts using @actions/artifact


konradpabjan:
It is currently possible to delete an artifact that is associated with a workflow run using the GitHub API: https://developer.github.com/v3/actions/artifacts/#delete-an-artifact

This API is limited though because an artifact is not available until after a run is complete

The `@actions/artifact` package should add a public `deleteArtifact()` method that will allow users to delete an artifact during a run

This will make it easier to support certain types types of scenarios such as retention

bradennapier:
Agreed - additionally the artifacts using the lib are scoped to the workflow only so there are some other challenges .  I wanted to make sure linting occurred when a forked PR was created and ended up needing to do some magic with the rest api directly to get it all working.

https://github.com/bradennapier/eslint-plus-action/blob/master/src/artifacts.ts

This type of application should be natively supported without question - especially since it solves the security problem, but it is also an example of how you can use the rest api for it.   I have some extra pieces in there which may not be necessary but I haven't had a change to test without yet (I am deleting then remaking each time to make sure as I wasn't positive if it would overwrite if different workflows were creating them).

robinxb:
+1 for deploying some artifacts when using other deployment. Now I have to sleep for few seconds to make this work.

sefasenturk95:
When is this expected to be implemented?

muzimuzhi:
> This API is limited though because an artifact is not available until after a run is complete

This limitation is removed in [`@actions/artifact@v2`](https://github.com/actions/toolkit/blob/5430c5d84832076372990c7c27f900878ff66dc9/packages/artifact/README.md):

> Improvements
> [...]
> 2\. Once uploaded, an Artifact ID is returned and Artifacts are immediately available in the UI and [REST API](https://docs.github.com/en/rest/actions/artifacts). Previously, you would have to wait for the run to be completed before an ID was available or any APIs could be utilized.
> 3\. Artifacts can now be downloaded and deleted from the UI before the entire workflow run finishes.

PR #1626, opened 2024-01-17
Add methods to delete artifacts

robherley:
Having official client methods to delete artifacts will make some use cases much simpler.
For instance, this could allow the ability to overwrite an artifact eventually in actions/upload-artifact.

e.g.

```yaml
- uses: actions/upload-artifact@v4
  with:
    name: my-artifact
    path: path/to/artifact/
    overwrite: true # something like this
```

Where overwrite would delete an artifact first, allowing for the upload of another.

I'll update the documentation in a separate PR.

Closes: https://github.com/actions/toolkit/issues/380
Partially supports: https://github.com/actions/upload-artifact/issues/471

### DIFF
--- a/packages/artifact/README.md
+++ b/packages/artifact/README.md
@@ -12,6 +12,7 @@ This is the core library that powers the [`@actions/upload-artifact`](https://gi
   - [Quick Start](#quick-start)
   - [Examples](#examples)
     - [Upload and Download](#upload-and-download)
+    - [Delete an Artifact](#delete-an-artifact)
     - [Downloading from other workflow runs or repos](#downloading-from-other-workflow-runs-or-repos)
     - [Speeding up large uploads](#speeding-up-large-uploads)
   - [Additional Resources](#additional-resources)
@@ -104,6 +105,19 @@ const {downloadPath} = await artifact.downloadArtifact(id, {
 })
 
 console.log(`Downloaded artifact ${id} to: ${downloadPath}`)
+```
+
+### Delete an Artifact
+
+To delete an artifact, all you need is the name. Also supports options to delete from other repos/runs given a token with proper permissions is supplied.
+
+```js
+const {id} = await artifact.deleteArtifact(
+  // name of the artifact
+  'my-artifact'
+)
+
+console.log(`Deleted Artifact ID: ${id}`)
 ```
 
 ### Downloading from other workflow runs or repos
--- a/packages/artifact/README.md
+++ b/packages/artifact/README.md
@@ -109,7 +109,7 @@ console.log(`Downloaded artifact ${id} to: ${downloadPath}`)
 
 ### Delete an Artifact
 
-To delete an artifact, all you need is the name. Also supports options to delete from other repos/runs given a token with proper permissions is supplied.
+To delete an artifact, all you need is the name.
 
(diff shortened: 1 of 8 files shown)


packages/artifact/README.md
```diff
@@ -106,6 +107,19 @@ const {downloadPath} = await artifact.downloadArtifact(id, {
 console.log(`Downloaded artifact ${id} to: ${downloadPath}`)
 ```
 
+### Delete an Artifact
+
+To delete an artifact, all you need is the name. Also supports options to delete from other repos/runs given a token with proper permissions is supplied.
```
konradpabjan, line 112: Could we add a separate example here for deleting an artifact from another workflow run with the `findBy` options? Just so it's super obvious for anyone going through the documentation

Permissions might be a bit vague here. `GITHUB_TOKEN` won't work but you rather need a PAT with the repository write permission I think

packages/artifact/src/internal/client.ts
```diff
@@ -225,4 +244,41 @@ If the error persists, please check whether Actions and API requests are operati
...
+  async deleteArtifact(
+    artifactName: string,
+    options?: FindOptions
+  ): Promise<DeleteArtifactResponse> {
+    try {
+      if (isGhes()) {
+        throw new GHESNotSupportedError()
+      }
+
+      if (options?.findBy) {
```
konradpabjan, line 257: I'm imagining an edge cases where someone provides `options?.findBy` but the `repositoryOwner`, `repositoryName` and `workflowRunID` all match the current run so it should technically go and try to use the internal delete method. Do you think we need this bit of logic here?

packages/artifact/README.md
robherley: Yea that's a good idea

packages/artifact/src/internal/client.ts
robherley: At least for the folks using typescript, you _must_ provide a token when using `findBy` because of it's typing: https://github.com/actions/toolkit/blob/1b5a6e26f42769414fe2302e6a6e8b5ac7bf9600/packages/artifact/src/internal/shared/interfaces.ts#L128-L135

I have this same `if (options?.findBy)` check in other methods, honestly I think it's good enough. If we were to change it, I would also want to do the same for the others as well.

packages/artifact/README.md
robherley: Added in https://github.com/actions/toolkit/pull/1626/commits/1852eb2115645114d47975ca9eef48d4e56b17b0

vmjoseph (approved):

packages/artifact/src/internal/client.ts
```diff
@@ -225,4 +244,41 @@ If the error persists, please check whether Actions and API requests are operati
...
+          workflowRunId,
+          repositoryOwner,
+          repositoryName,
+          token
+        )
+      }
+
+      return deleteArtifactInternal(artifactName)
+    } catch (error) {
+      warning(
```
campersau, line 273: Since there is no `await` for `deleteArtifactPublic` / `deleteArtifactInternal` is this warning actually printed when something goes wrong asynchronously? To me it looks this only catches errors which are thrown synchronously.

PR #1652, opened 2024-02-01
Update workflows to use v4 actions

konradpabjan:
v3 actions run on node16 which will be deprecated. We should move to the latest and greatest


thboop (approved):
lgtm

CosmicJesterX (commented):
i gotta run. good luck witht he workflows

PR #1666, opened 2024-02-22
Use `unzip.Parse` over `unzip.Extract`

bethanyj28:
Updating our unzipping logic to use `unzip.Parse` and handle the `entry` event for greater control of the unzip process.


SrRyan (approved):

konradpabjan (approved):

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -89,10 +92,40 @@ export async function streamExtractExternal(
         clearTimeout(timer)
         reject(error)
       })
-      .pipe(unzip.Extract({path: directory}))
-      .on('close', () => {
+      .pipe(unzip.Parse())
+      .on('entry', (entry: unzip.Entry) => {
+        const entryPath = path
+          .normalize(entry.path)
+          .replace(/^(\.\.(\/|\\|$))+/, '')
```
ericsciple, line 99: What if entryPath is empty after replace is called?

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -89,10 +92,40 @@ export async function streamExtractExternal(
...
+        const fullPath = path.join(directory, entryPath)
+        core.debug(`Extracting artifact entry: ${fullPath}`)
+        if (entry.type === 'Directory') {
+          promises.push(resolveOrCreateDirectory(fullPath).then(() => {}))
+          entry.autodrain()
+        } else {
+          promises.push(
+            resolveOrCreateDirectory(path.dirname(fullPath)).then(() => {})
```
ericsciple, line 107: Curious why would we create a directory using `fullPath` when it's not a directory?

packages/artifact/src/internal/download/download-artifact.ts
ericsciple: For example, the regexp pattern appears to be designed to handle cases like:

- `..`
- `../`
- `..\`

which will all result in an empty string after the replace call

packages/artifact/src/internal/download/download-artifact.ts
bethanyj28: `path.dirname(fullPath)` only takes the directory to the file - not the file itself - and creates it if it doesn't exist

ericsciple: Ah! Gotcha!

packages/artifact/src/internal/download/download-artifact.ts
bethanyj28: `path.join` ignores zero-length paths: https://nodejs.org/api/path.html#pathjoinpaths

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -89,10 +92,40 @@ export async function streamExtractExternal(
         clearTimeout(timer)
         reject(error)
       })
-      .pipe(unzip.Extract({path: directory}))
-      .on('close', () => {
+      .pipe(unzip.Parse())
+      .on('entry', (entry: unzip.Entry) => {
```
ericsciple, line 96: Wondering whether we need any new tests for the new logic or whether we feel confident that we have coverage based on existing tests?

bethanyj28: Added a test, however the current tests were very helpful for validating this change so I'm pretty confident in the existing ones!

packages/artifact/src/internal/download/download-artifact.ts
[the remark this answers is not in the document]
bethanyj28: I added a set containing checked directories to prevent reaching out to `exists()` so much!

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -89,10 +98,47 @@ export async function streamExtractExternal(
...
+        const fullPath = path.normalize(path.join(directory, entry.path))
```
JarLob, line 103:
```suggestion
        const fullPath = path.normalize(path.join(directory, entry.path))
        if (!directory.endsWith(path.sep))
          directory += path.sep
```

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -89,8 +99,51 @@ export async function streamExtractExternal(
...
         reject(error)
       })
-      .pipe(unzip.Extract({path: directory}))
-      .on('close', () => {
+      .pipe(unzip.Parse())
+      .pipe(
+        new stream.Transform({
+          objectMode: true,
+          transform: async (entry, _, callback) => {
+            const fullPath = path.normalize(path.join(directory, entry.path))
```
TingluoHuang, line 107: `const fullPath = path.resolve(directory, entry.path)`

will `path.resolve()` be an better option, assume `entry.path` should be a relative path?

bethanyj28: That makes sense to me! I think `download-artifact` ensures that the path is absolute, but it doesn't look like toolkit does any validation.

bethanyj28: Ah, doing that throws off the check below for the directory path 😞

ericsciple (approved):

packages/artifact/src/internal/download/download-artifact.ts
bethanyj28: Would you mind if I changed it back? On second thought, it might be better to keep this generic and not assume path for anyone utilizing this package. For download-artifact, we have the absolute path: https://github.com/actions/download-artifact/blob/348754975ef0295bfa2c111cba996120cfdf8a5d/src/download-artifact.ts#L38

TingluoHuang: i thought `directory` is absolute path, looks like it can be relative...

bethanyj28: Directory is absolute when it comes from `download-artifact`, but isn't guaranteed absolute from other sources.

TingluoHuang (approved):

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -89,8 +99,51 @@ export async function streamExtractExternal(
...
+            if (!directory.endsWith(path.sep)) {
```
JarLob, line 108:
```suggestion
            directory = path.resolve(directory)
            if (!directory.endsWith(path.sep)) {
```

JarLob: It is important for the comparison below. Otherwise it will fail with, for example, "/tmp/dest/file".startsWith("../dest/")

bethanyj28: We decided not to do the path.resolve above since this library could take relative paths!

bethanyj28: That was definitely failing when we switched to resolve temporarily.

JarLob: Ah, OK, `fullPath.startsWith(directory)` should work correctly then.

gdamiaN538 (commented):
[gh]

gdamiaN538:
Read

PR #1670, opened 2024-02-26
Ensure callback is only called once

bethanyj28:
We have gotten reports of errors using the latest download-artifact@main:
```
##[debug]Failed to download artifact after 5 retries due to Callback called multiple times. Retrying in 5 seconds...
Error: Unable to download artifact(s): Unable to download and extract artifact: Artifact download failed after 5 retries.
##[debug]response.message: Artifact download failed: Blob storage chunk did not respond in 30000ms
##[debug]response.message: Artifact download failed: Blob storage chunk did not respond in 30000ms
##[debug]response.message: Artifact download failed: Blob storage chunk did not respond in 30000ms
##[debug]response.message: Artifact download failed: Blob storage chunk did not respond in 30000ms
##[debug]response.message: Artifact download failed: Blob storage chunk did not respond in 30000ms
##[debug]Node Action run completed with exit code 1
``` 
Given `Callback called multiple times` it seems like it's attempting to create the directory and calling `callback` twice. It should only be called once when the file is extracted.

Part of https://github.com/actions/download-artifact/issues/302

### DIFF
--- a/packages/artifact/src/internal/download/download-artifact.ts
+++ b/packages/artifact/src/internal/download/download-artifact.ts
@@ -128,10 +128,7 @@ export async function streamExtractExternal(
               if (!createdDirectories.has(path.dirname(fullPath))) {
                 createdDirectories.add(path.dirname(fullPath))
                 await resolveOrCreateDirectory(path.dirname(fullPath)).then(
-                  () => {
-                    entry.autodrain()
-                    callback()
-                  }
+                  () => {}
                 )
               }
 
--- a/packages/artifact/src/internal/download/download-artifact.ts
+++ b/packages/artifact/src/internal/download/download-artifact.ts
@@ -127,9 +127,7 @@ export async function streamExtractExternal(
             } else {
               if (!createdDirectories.has(path.dirname(fullPath))) {
                 createdDirectories.add(path.dirname(fullPath))
-                await resolveOrCreateDirectory(path.dirname(fullPath)).then(
-                  () => {}
-                )
+                await resolveOrCreateDirectory(path.dirname(fullPath))
               }
 
               const writeStream = createWriteStream(fullPath)


rentziass (approved):

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -128,10 +128,7 @@ export async function streamExtractExternal(
               if (!createdDirectories.has(path.dirname(fullPath))) {
                 createdDirectories.add(path.dirname(fullPath))
                 await resolveOrCreateDirectory(path.dirname(fullPath)).then(
-                  () => {
-                    entry.autodrain()
```
robherley, line 132: is the `entry.autodrain()` no longer necessary?

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -128,10 +128,7 @@ export async function streamExtractExternal(
...
-                    callback()
-                  }
+                  () => {}
```
robherley, line 131: you can drop the `.then( ... )` altogether

packages/artifact/src/internal/download/download-artifact.ts
bethanyj28: Yep! Since below we're using `entry.pipe(...)`

packages/artifact/src/internal/download/download-artifact.ts
bethanyj28: Ah, you're right! Got my wires crossed from when I was using a promises array.

bethanyj28: https://github.com/actions/toolkit/pull/1670/commits/dcc55dfd0485545474048b888c6167aa9206d92c

robherley (approved):

PR #1677, opened 2024-02-29
Flip releases update order

bethanyj28:
Flipping `RELEASES.md` order for cache so it's easier to tell what changed via Dependabot. Additionally adding a release note to artifacts.


robherley (approved):
👏

PR #1690, opened 2024-03-15
Replacing `archiver` package with direct use of `zip-stream` for fixing macOS upload issues

vmjoseph:
This PR replaces the previous `archiver` package in upload-artifacts for the direct use of the `zip-stream` so that we can better hook into the finalize portion of the lifecycle. This should resolve issues on macOS where upload streams are incorrectly closed before the upload is finished due to a pre-existing issue with `archiver`.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -43,7 +43,7 @@ Note that before a PR will be accepted, you must ensure:
 1. In a new branch, create a new Lerna package:
 
 ```console
-$ npm run create-package new-package
+$ npm run new-package [name]
 ```
 
 This will ask you some questions about the new package. Start with `0.0.0` as the first version (look generally at some of the other packages for how the package.json is structured).
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "eslint-plugin-prettier": "^5.0.0",
         "flow-bin": "^0.115.0",
         "jest": "^29.6.4",
-        "lerna": "^7.1.4",
+        "lerna": "^6.4.1",
         "nx": "16.6.0",
         "prettier": "^3.0.0",
         "ts-jest": "^29.1.1",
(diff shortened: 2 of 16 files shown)


Nidacoke (approved):

packages/artifact/src/internal/upload/upload-artifact.ts
```diff
@@ -67,34 +67,38 @@ export async function uploadArtifact(
...
+  // Create the zipupload stream for use in blob upload
   const zipUploadStream = await createZipUploadStream(
     zipSpecification,
     options?.compressionLevel
-  )
+  ).catch(err => {
+    throw new InvalidResponseError(
+      `createZipUploadStream: response from backend was not ok: ${err}`
+    )
+  })
```
robherley, line 78: these awaited async functions should bubble up thrown errors, we shouldn't need this if we have an outer try/catch

packages/artifact/src/internal/upload/zip.ts
```diff
@@ -30,31 +31,57 @@ export async function createZipUploadStream(
...
-  const bufferSize = getUploadChunkSize()
-  const zipUploadStream = new ZipUploadStream(bufferSize)
+  async.eachSeries(uploadSpecification, addFileToZip, (error: unknown) => {
+    if (error) {
+      core.error('Failed to add a file to the zip:')
+      core.info(error.toString()) // Convert error to string
+      return
+    }
+    zip.finalize() // Finalize the archive once all files have been added
+  })
```
robherley, line 84: instead of introducing a new dependency on `async`, can we just await on a group of these promises?

e.g. 

```js
await Promise.all(uploadSpecification.map(addFileToZip))
```

or something like that? `eachSeries` [does one at a time](https://github.com/caolan/async/blob/92ea2182de27d3e6a02d79562ba562e3c6d59b88/lib/eachSeries.js#L5), does it have to do that for a specific reason? and do we know if this will impact speed?

packages/artifact/__tests__/upload-artifact.test.ts
```diff
@@ -7,11 +7,21 @@ import {Timestamp, ArtifactServiceClientJSON} from '../src/generated'
...
 
+  constructor(message?: string, code?: string) {
+    super(message) // Pass the message to the Error constructor
+    this.code = code || ''
+  }
+}
 describe('upload-artifact', () => {
   beforeEach(() => {
-    noopLogs()
+    // noopLogs()
```
bethanyj28, line 24: Should this be uncommented?

packages/artifact/src/internal/upload/upload-artifact.ts
```diff
@@ -67,34 +67,38 @@ export async function uploadArtifact(
...
   }
-
+  // Create the zipupload stream for use in blob upload
   const zipUploadStream = await createZipUploadStream(
     zipSpecification,
     options?.compressionLevel
-  )
+  ).catch(err => {
+    throw new InvalidResponseError(
+      `createZipUploadStream: response from backend was not ok: ${err}`
```
bethanyj28, line 76: This error message might be a bit misleading since errors can come from that function without reaching out to a backend.

PR #1693, opened 2024-03-22
(@actions/attest) build provenance statement from OIDC claims

bdehamer:
Updates the `attestProvenance` function in the `@actions/attest` package so that the provenance statement is populated from data read from a verified GHA OIDC token.

Previously, the provenance data was read directly from the environment. With the new approach, an OIDC token is retrieved from the GHA issuer and the provenance statement is built from the claims present in that token. This more closely mirrors the approach that is used by Sigstore certificate authority (Fulcio) when it populates the claims in the signing certificate.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5781,9 +5781,9 @@
       }
     },
     "node_modules/follow-redirects": {
-      "version": "1.15.4",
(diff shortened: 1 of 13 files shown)


ejahnGithub (approved):
nice work!

luketomlinson (approved):

PR #947, opened 2021-11-25
Ability to overwrite base url in actions/cache

sindrig:
As ACTIONS_CACHE_URL cannot be overwritten in actions/cache,
provide the ability to overwrite it using GITHUB_ACTIONS_CACHE_URL.

See https://github.com/actions/cache/pull/679 for more context

Issue #679, opened 2020-12-22
DOC: Is return necessary after setFailed?


pllim:
Hello. If I decide to use `core.setFailed(...)` but there are code after that runs if it did not fail, do I need to also add a `return` or does `setFailed` automatically terminates the Action? It is not clear from the README.

Example logic:

```typescript
import * as core from "@actions/core";

async function run() {
    if (condition1) {
        core.setFailed("You should not have done that");
        // Do I need return; here?
    }
   do_stuff_on_success();
}
```

thboop:
Hey @pllim thanks for the information! `setFailed` does not automatically terminate the process, you need to do that manually. You can do that via a return, or following an example like our [typescript actions](https://github.com/actions/typescript-action/blob/main/src/main.ts)

I think the docs [here](https://github.com/actions/toolkit/blob/main/packages/core/src/core.ts#L143) are good now, but if you have an idea for how to improve them, please submit a pr!

PR #1695, opened 2024-03-26
feat: allow passing custom `ACTIONS_CACHE_URL`

### LINKED ISSUE #1051: Add support for non-GitHub-hosted caching for self-hosted runners
**Describe the enhancement**

Right now, Action caching isn't very useful when using self-hosted GitHub Actions runners: I have runners deployed on my own hardware but the cached data is being sent to and from GitHub cache servers. The time and cost of making those transfers across the Internet and back makes caching almost entirely pointless for self-hosted runners.

Instead I'd like the ability to use the GitHub Actions caching API to cache data on my own servers nearer my runners.

What's missing:
* there's no downloadable cache server implementation that we can use to run a cache server somewhere else
* there's no documentation for how to _write_ such a caching server even if GitHub doesn't provide one
* it doesn't look like there's any ability to control the action/toolkit cache package to make it work with a non-GitHub server. I can see there's an `ACTIONS_CACHE_URL` that is provided by the runner environment and points to a GitHub server, but there's no easy way to override that. Even if there were, it looks like the cache system authenticates using `ACTIONS_RUNTIME_TOKEN`, which also lacks documentation or the ability to override.  It's also used for more than just caching, so if I did somehow override it to point to a custom caching server, it would break something else in the actions runner.

**Additional information**
Add any other context about the feature here.

LouisHaftmann:
Closes #1051 

Currently it is not possible to overwrite the `ACTIONS_CACHE_URL` which is used for communicating with the cache api. The `ACTIONS_CACHE_URL` is always overwritten by the runner itself ([runner code snippet](https://github.com/actions/runner/blob/77e0bfbb8a8fde1f01fc1cf1ed2d7f0e81a0a407/src/Runner.Worker/Handlers/NodeScriptActionHandler.cs#L59)).

Being able to overwrite this environment variable is crucial for using the `actions/cache` action with self-hosted runners, as it allows for a self-hosted caching solution.

To set a custom `ACTIONS_CACHE_URL`, you just need to define the `CUSTOM_ACTIONS_CACHE_URL` (open for naming suggestions) env var, which takes precedence over the default cache url env var. 

Self-hosting an actions cache server is a much requested feature and this PR would make that way easier

Other issues/discussions:
- https://github.com/orgs/community/discussions/18549
- https://github.com/github/docs/issues/2271
- https://github.com/actions/runner/pull/3411

### DIFF
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -50,7 +50,7 @@ function checkKey(key: string): void {
  */
 
 export function isFeatureAvailable(): boolean {
-  return !!process.env['ACTIONS_CACHE_URL']
+  return !!process.env['CUSTOM_ACTIONS_CACHE_URL'] || !!process.env['ACTIONS_CACHE_URL']
 }
 
 /**
--- a/packages/cache/src/internal/cacheHttpClient.ts
+++ b/packages/cache/src/internal/cacheHttpClient.ts
@@ -40,7 +40,7 @@ import {
 const versionSalt = '1.0'
 
 function getCacheApiUrl(resource: string): string {
-  const baseUrl: string = process.env['ACTIONS_CACHE_URL'] || ''
+  const baseUrl: string = process.env['CUSTOM_ACTIONS_CACHE_URL'] || process.env['ACTIONS_CACHE_URL'] || ''
   if (!baseUrl) {
     throw new Error('Cache Service Url not found, unable to restore cache.')
   }


edigaryev:
@robherley, @bethanyj28 would you mind taking a look at this?

LouisHaftmann:
Any feedback on this PR? Is there something I should implement differently or something missing? I will gladly do anything to get this PR merged.

LouisHaftmann:
Here are some more PRs mentioning the same issue:
#947
#679

Would be great to have this feature! I don't see any real problem or security risk with this. Am I missing something? Would there be a *safer* way to allow for overwriting the cache server url?

Please let me know 🙏 @robherley @bethanyj28

fingerart:
@robherley @bethanyj28

breezewish:
Would be great to have this option! It is hard to overwrite the cache url in self-hosted runners currently.

fkorotkov:
In the meantime we created a [simple fork](https://github.com/cirruslabs/cache) that patches compiled JavaScript in `dist` folder of `actions/cache` to allow overriding `ACTIONS_CACHE_URL`.

See [here](https://github.com/cirruslabs/cache?tab=readme-ov-file#running-with-your-own-implementation-of-github-actions-cache-api) how to run it with your own implementation of the Caching API.

Tycale:
[Patching the binary](https://gha-cache-server.falcondev.io/getting-started#docker) also works, but it’s frustrating not to be able to do it correctly. It's such a small patch..

LouisHaftmann:
I'm beginning to feel like only Github staff contributions are welcome in this repo 🥲

Tycale:
I tried contacting a staff member via email to review this PR. Fingers crossed! 🤞🏻

enescakir:
Hi @Link- , Enabling the overwriting the `ACTIONS_CACHE_URL` variable would simplify things for the community. What do you think about this PR?

Link-:
> Hi @Link- , Enabling the overwriting the `ACTIONS_CACHE_URL` variable would simplify things for the community. What do you think about this PR?

I do not believe we have plans to support this path at the moment. We're introducing major changes and optimisations to the cache service which will be landing soon.

cc: @lkfortuna

tetienne:
Hi @Link- Can we have visibility on your plan, anyplace where can see progress?

PR #1699, opened 2024-04-02
feat(cache): support passing extra tar args

deemp:
In `cache-nix-action`, I exclude some files when restoring a cache ([link](https://github.com/nix-community/cache-nix-action/blob/4284df42cb6d492efb1a32aeb325a95a7b1a2260/src/utils/action.ts#L100-L112)).

Excluding requires access to the `tar` command arguments.

Currently, I have a fork of `actions/toolkit`.

However, I believe it'd be nice to have such functionality in the main `actions/toolkit`.


deemp:
@robherley, @bethanyj28, please review

jgoux:
As a user of `cache-nix-action` using an alternative cache, having this merged in the official cache would be super useful for us as well.

mikehearn:
+1 this would allow a targeted workaround for issue #1753

deemp:
@robherley @Link- please review

deemp:
TODO 

some arguments of `create`, `list`, and `extract` are positional. Hence, extra arguments should be inserted into right places when invoking these commands. E.g., before '--files-from' in 'create'

https://github.com/actions/toolkit/blob/8780507298eb66f677ace2cadf09e1c2f5559661/packages/cache/src/internal/tar.ts#L86

PR #1704, opened 2024-04-05
Add option to overwrite files on cache extraction

siburny:
When using GNU tar, if a file already exists, from a previous run or another cached action, it results in a slew of warning messages.

This change adds the use of `--overwrite` when using GNU tar so that any existing file will just be overwritten.

Similar issue was fixed in the `@actions/tool-cache` package, but not in `@actions/cache`: https://github.com/actions/toolkit/pull/807/files#diff-725a5ef102ee3762738a5500c3d7c5b6afef874a1cb67c55085ad5e4615b1a6f

### DIFF
--- a/packages/cache/src/internal/tar.ts
+++ b/packages/cache/src/internal/tar.ts
@@ -111,6 +111,9 @@ async function getTarArgs(
 
   // Platform specific args
   if (tarPath.type === ArchiveToolType.GNU) {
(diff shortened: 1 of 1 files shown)


ytimocin (approved):

siburny:
I deleted the my forked repo without realizing it will break the PR. I recreated the report and PR: https://github.com/actions/toolkit/pull/1891

PR #1712, opened 2024-04-09
Upgrading `upload-artifact` and `download-artifact` archiver package

vmjoseph:
macOS uploads look like they're stalling due to unclosed node processes. We're going to upgrade the `archiver` pckage used by both `upload-artifact` and `download-artifact` to see if that alleviates the issue. If not, our next step will be to replace the `archiver` package all together.


packages/artifact/src/internal/upload/blob-upload.ts
```diff
@@ -37,8 +39,19 @@ export async function uploadZipToBlobStorage(
   const uploadCallback = (progress: TransferProgressEvent): void => {
     core.info(`Uploaded bytes ${progress.loadedBytes}`)
     uploadByteCount = progress.loadedBytes
+    progressTimeout(timeoutDuration)
```
robherley, line 42: should update `lastProgressTime` after this

packages/artifact/src/internal/upload/blob-upload.ts
```diff
@@ -37,8 +39,19 @@ export async function uploadZipToBlobStorage(
...
   }
 
+  // Timeout if the upload stalls
+  const progressTimeout = (timeout: number): NodeJS.Timeout =>
+    setTimeout(() => {
```
robherley, line 47: let's set this to a variable scoped outside of this like `timer` and use that to clear the interval

packages/artifact/src/internal/upload/blob-upload.ts
```diff
@@ -37,8 +56,12 @@ export async function uploadZipToBlobStorage(
   const uploadCallback = (progress: TransferProgressEvent): void => {
     core.info(`Uploaded bytes ${progress.loadedBytes}`)
     uploadByteCount = progress.loadedBytes
+    chunkTimer(timeoutDuration)
+    lastProgressTime = Date.now()
   }
 
+  // // Timeout if the upload stalls
+  // const progressTimeout
```
robherley, line 64: needs a lil comment cleanup 😄

packages/http-client/package-lock.json
```diff
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@actions/http-client",
-      "version": "2.2.0",
+      "version": "2.2.1",
```
robherley, line 9: 🤔 was this just missed when http-client was updated to 2.2.1?

packages/artifact/src/internal/upload/blob-upload.ts
```diff
@@ -80,6 +109,8 @@ export async function uploadZipToBlobStorage(
     )
   }
 
+  // clear the progress timeout when upload completes
+  clearTimeout(chunkTimer(timeoutDuration))
```
robherley, line 113: we probably don't need this anymore bc of the `finally` above

also this should use `timeoutId` not the chunk timer func

packages/http-client/package-lock.json
vmjoseph: 👍🏾 it looks like it

packages/artifact/__tests__/upload-artifact.test.ts
```diff
@@ -351,4 +354,119 @@ describe('upload-artifact', () => {
...
+    // eslint-disable-next-line @typescript-eslint/await-thenable
+    await fs.writeFile(
+      path.join(dirPath, 'file3.txt'),
+      'test file content',
+      err => {
+        if (err) {
+          throw err
+        }
+      }
+    )
```
robherley, line 393: you can import the promise-wrapped version of fs's writeFile. these that you are awaiting aren't really doing what you are expecting, they are not returning a promise which is why you are tripping eslint

```js
import {writeFile} from 'fs/promises'
```

robherley: you can also use `writeFileSync` from the normal fs package since this is only a test

packages/artifact/package.json
```diff
@@ -60,6 +60,7 @@
     "@types/unzip-stream": "^0.3.4",
     "typedoc": "^0.25.4",
     "typedoc-plugin-markdown": "^3.17.1",
+    "mock-fs": "^5.2.0",
```
robherley, line 63: is this being used anywhere? we can probably take it out

robherley (approved):

luketomlinson (approved):

PR #1724, opened 2024-04-23
Use latest `unzip-stream` and `unzip.Extract`

bethanyj28:
This updates the `unzip-stream` dependency and goes back to using `unzip.Extract` over `unzip.Parse`.

### DIFF
--- a/packages/artifact/src/internal/download/download-artifact.ts
+++ b/packages/artifact/src/internal/download/download-artifact.ts
@@ -1,7 +1,4 @@
 import fs from 'fs/promises'
-import * as stream from 'stream'
-import {createWriteStream} from 'fs'
-import * as path from 'path'
 import * as github from '@actions/github'
 import * as core from '@actions/core'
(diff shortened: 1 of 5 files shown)


yacaovsnc (approved):

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -99,46 +89,8 @@ export async function streamExtractExternal(
...
-              const writeStream = createWriteStream(fullPath)
-              writeStream.on('finish', callback)
-              writeStream.on('error', reject)
-              entry.pipe(writeStream)
-            }
-          }
-        })
-      )
-      .on('finish', async () => {
+      .pipe(unzip.Extract({path: directory}))
```
yacaovsnc, line 92: Glad we don't have to roll our own implementation anymore

bethanyj28: Agreed! Simplifies this a ton.

PR #1725, opened 2024-04-24
(@actions/attest) retry request on failure to save attestation

bdehamer:
Updates the attestations API client to be more resilient to intermittent errors.

Enables the `@octokit/plugin-retry` plugin for the Octokit client and configures a (default) retry count of 3. Any attempt to persist the generated attestation which results in a 5XX error will be retried up to 3 times (with [exponential backoff](https://github.com/octokit/plugin-retry.js/blob/v6.0.1/src/error-request.ts#L13)).

### DIFF
--- a/packages/attest/__tests__/store.test.ts
+++ b/packages/attest/__tests__/store.test.ts
@@ -52,7 +52,39 @@ describe('writeAttestation', () => {
     })
 
     it('throws an error', async () => {
-      await expect(writeAttestation(attestation, token)).rejects.toThrow(/oops/)
+      await expect(
+        writeAttestation(attestation, token, {retry: 0})
+      ).rejects.toThrow(/oops/)
+    })
+  })
+
+  describe('when the api call fails but succeeds on retry', () => {
+    beforeEach(() => {
+      const pool = mockAgent.get('https://api.github.com')
+
+      pool
+        .intercept({
+          path: '/repos/foo/bar/attestations',
+          method: 'POST',
+          headers: {authorization: `token ${token}`},
+          body: JSON.stringify({bundle: attestation})
+        })
+        .reply(500, 'oops')
+        .times(1)
+
+      pool
+        .intercept({
+          path: '/repos/foo/bar/attestations',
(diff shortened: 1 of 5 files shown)


packages/attest/package.json
```diff
@@ -46,9 +46,15 @@
...
+    "@octokit/plugin-retry": "^6.0.1",
     "@sigstore/bundle": "^2.3.0",
     "@sigstore/sign": "^2.3.0",
     "jsonwebtoken": "^9.0.2",
     "jwks-rsa": "^3.1.0"
+  },
+  "overrides": {
+    "@octokit/plugin-retry": {
+      "@octokit/core": "^5.2.0"
+    }
```
bdehamer, line 58: The `@octokit/plugin-retry` package has a very loose peer dependency on `@octokit/core`, however we need to pin it to the same version which is used by `@actions/github` to avoid issues with type inconsistencies.

packages/attest/__tests__/store.test.ts
```diff
@@ -52,7 +52,43 @@ describe('writeAttestation', () => {
...
+          body: JSON.stringify({bundle: attestation})
+        })
+        .reply(201, {id: '123'})
+        .times(1)
+    })
+
+    it('persists the attestation', async () => {
+      await expect(
+        writeAttestation(attestation, token, {
+          retry: 2
```
trevrosen, line 89: I think `5` might make more sense here?

packages/attest/src/store.ts
```diff
@@ -1,7 +1,12 @@
 import * as github from '@actions/github'
+import {retry} from '@octokit/plugin-retry'
 
 const CREATE_ATTESTATION_REQUEST = 'POST /repos/{owner}/{repo}/attestations'
+const DEFAULT_RETRY_COUNT = 3
```
trevrosen, line 5: Mentioned above, but also for posterity: let's make this `5`

bdehamer: 👍

phillmv (approved):

PR #1735, opened 2024-05-16
Read the server url from the environment variable.

kommendorkapten:
Instead of having the urls hardcoded, read them from the environment. I opted to read from the environment variable instead of the github context because it would be easier to test.

### DIFF
--- a/packages/attest/src/endpoints.ts
+++ b/packages/attest/src/endpoints.ts
@@ -42,15 +42,14 @@ export const signingEndpoints = (sigstore?: SigstoreInstance): Endpoints => {
 }
 
 function buildGitHubEndpoints(): Endpoints {
-  const serverURL = process.env.GITHUB_SERVER_URL ?? `https://github.com`
-  let url = serverURL.replace('https://', '')
+  const serverURL = process.env.GITHUB_SERVER_URL || 'https://github.com'
+  let host = new URL(serverURL).hostname
 
-  if (url === 'github.com') {
-    url = 'githubapp.com'
+  if (host === 'github.com') {
+    host = 'githubapp.com'
   }
-  const endpoints: Endpoints = {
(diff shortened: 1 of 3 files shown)


packages/attest/src/endpoints.ts
```diff
@@ -45,6 +37,20 @@ export const signingEndpoints = (sigstore?: SigstoreInstance): Endpoints => {
...
       return SIGSTORE_PUBLIC_GOOD
     case GITHUB_ID:
-      return SIGSTORE_GITHUB
+      return buildGitHubEndpoints()
+  }
+}
+
+function buildGitHubEndpoints(): Endpoints {
+  const serverURL = process.env.GITHUB_SERVER_URL ?? `https://github.com`
+  let url = serverURL.replace('https://', '')
```
bdehamer, line 46: Using the URL parsing here feels a bit safer and will address weird things like a trailing `/` appearing in the URL
```suggestion
  let host = new URL(serverURL).hostname
```

packages/attest/src/endpoints.ts
```diff
@@ -45,6 +37,20 @@ export const signingEndpoints = (sigstore?: SigstoreInstance): Endpoints => {
...
+
+  if (url === 'github.com') {
+    url = 'githubapp.com'
+  }
+  const endpoints: Endpoints = {
+    fulcioURL: `https://fulcio.${url}`,
+    tsaServerURL: `https://timestamp.${url}`
   }
+  return endpoints
```
bdehamer, line 55: Should be able to return a literal here:
```suggestion
  return {
    fulcioURL: `https://fulcio.${url}`,
    tsaServerURL: `https://timestamp.${url}`
  }
```

bdehamer (commented):
I few minor idiomatic JavaScript things here.

bdehamer (approved):
👍

PR #1723, opened 2024-04-23
Include fs stats and mode in zip archive

rmunn:
This should allow Unix file permissions to be included in the zip file created by upload-artifact, fixing https://github.com/actions/upload-artifact/issues/38 and probably https://github.com/actions/upload-artifact/issues/37 as well.

Fixes #1722.

This is basically https://github.com/actions/toolkit/pull/1609, rebased against the current state of the `main` branch. But #1609 did not include the `mode` property in its call to `zip.append`, and [my research suggests](https://github.com/actions/toolkit/pull/1609#issuecomment-2071697751) that `mode` is needed to actually include the permissions in the zip file.

PR #1771, opened 2024-07-23
Prevent "too many open files" in artifact upload

rmunn:
Fixes https://github.com/actions/upload-artifact/issues/485.

[The `.file` method](https://www.archiverjs.com/docs/archiver/#file) uses "a [lazystream](https://github.com/jpommerening/node-lazystream) wrapper to prevent issues with open file limits", which is precisely the scenario we're running into with https://github.com/actions/upload-artifact/issues/485.

This was originally part of https://github.com/actions/toolkit/pull/1723, but I pulled the bugfix out into its own PR so it can be reviewed separately from the bugfix that https://github.com/actions/toolkit/pull/1723 was for (preserving Unix file permissions in uploaded .zip files). This one is more urgent than the Unix file permissions, which can be worked around by uploading a .tar file.

### DIFF
--- a/packages/artifact/src/internal/upload/zip.ts
+++ b/packages/artifact/src/internal/upload/zip.ts
@@ -1,7 +1,6 @@
 import * as stream from 'stream'
 import * as archiver from 'archiver'
 import * as core from '@actions/core'
-import {createReadStream} from 'fs'
 import {UploadZipSpecification} from './upload-zip-specification'
 import {getUploadChunkSize} from '../shared/config'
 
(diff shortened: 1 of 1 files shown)


robherley (approved):

rmunn:
Once #1774 is merged, I'll rebase on top of `develop` so that the tests can pass.

rmunn:
Just rebased on top of `main`; let's see what the unit tests do now.

rmunn:
@robherley - Looks like a team member needs to re-approve running the workflows again.

joshmgross:
Looks like this can be removed:
```ts
import {createReadStream} from 'fs'
```

rmunn:
> Looks like this can be removed:
> 
> ```ts
> import {createReadStream} from 'fs'
> ```

Done in commit 7c61054.

rmunn:
@robherley - Your approving review is now considered stale because I rebased and pushed one new commit. Want to give this another look? I should be done making changes.

robherley (approved):

PR #1774, opened 2024-07-24
Fix artifact upload chunk timeout logic + update tests

robherley:
While looking at [CI output](https://github.com/actions/toolkit/actions/runs/10057662696/job/27819363234?pr=1771#step:7:324) for:
- https://github.com/actions/toolkit/pull/1771

We had an issue/race where we were trying to do stuff _after_ the test completed:
> ReferenceError: You are trying to `import` a file after the Jest environment has been torn down. From packages/artifact/__tests__/upload-artifact.test.ts.

Turns out we had a few errors in our tests and I also discovered we were not handling our timeout logic correctly:

The problematic area of the timeout logic was that it was throwing an error within the `setTimeout` block, which is outside the promise area and went unhandled. To remedy this (and avoid other leaky async code) I swapped it with a `setInterval` wrapped in a Promise that would check the last chunk upload. Then, I gave this function and the blob upload an [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) so we can quit everything when we want to. Then, the main await is a [`Promise.race`](https://developer.mozilla.org/en-us/docs/web/javascript/reference/global_objects/promise/race) where we always abort after the first promise resolves.

As for the tests, we were incorrectly mocking over the logic we were testing and throwing immediately. I also took the liberty of updating the rest of the tests since there was a lot of duplication that made each test harder to reason about.

### DIFF
--- a/packages/artifact/src/internal/shared/config.ts
+++ b/packages/artifact/src/internal/shared/config.ts
@@ -59,5 +59,5 @@ export function getConcurrency(): number {
 }
 
 export function getUploadChunkTimeout(): number {
-  return 30_000
+  return 30_000 // 30 seconds
 }
--- a/packages/artifact/src/internal/shared/config.ts
+++ b/packages/artifact/src/internal/shared/config.ts
@@ -57,3 +57,7 @@ export function getConcurrency(): number {
   const concurrency = 16 * numCPUs
   return concurrency > 300 ? 300 : concurrency
 }
+
+export function getUploadChunkTimeout(): number {
+  return 30_000
+}
--- a/packages/artifact/__tests__/upload-artifact.test.ts
+++ b/packages/artifact/__tests__/upload-artifact.test.ts
@@ -1,260 +1,137 @@
 import * as uploadZipSpecification from '../src/internal/upload/upload-zip-specification'
 import * as zip from '../src/internal/upload/zip'
 import * as util from '../src/internal/shared/util'
-import * as retention from '../src/internal/upload/retention'
 import * as config from '../src/internal/shared/config'
-import {Timestamp, ArtifactServiceClientJSON} from '../src/generated'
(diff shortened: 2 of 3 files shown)


packages/artifact/src/internal/shared/config.ts
```diff
@@ -57,3 +57,7 @@ export function getConcurrency(): number {
   const concurrency = 16 * numCPUs
   return concurrency > 300 ? 300 : concurrency
 }
+
+export function getUploadChunkTimeout(): number {
+  return 30_000
```
konradpabjan, line 62: 300000?

konradpabjan: If it's supposed to be from the earlier `const timeoutDuration = 300000 // 30 seconds`, could you add a similar comment as to what this value is supposed to be. Or `getUploadChunkTimeout` could be renamed to something like `getUploadChunkTimeoutInMilliseconds`

robherley: The original chunk timeout was incorrectly set to 300000 ms, which is 300 seconds:
https://github.com/actions/toolkit/blob/1db73622df8a649a01cb12b8a217c903419009fa/packages/artifact/src/internal/upload/blob-upload.ts#L50

It was intended to be 30 seconds from the comment, which is why I used the numeric separator (`_`) to make it a bit more obvious

konradpabjan: Ah, nice catch! Could you then just add the same comment with `// 30 seconds` so anyone stumbling by can better understand it

robherley: done! https://github.com/actions/toolkit/pull/1774/commits/3e34f6d19cd8feeb813dd7bfb0de5d75aaa43565

konradpabjan (approved):

jww3 (approved):

PR #1775, opened 2024-07-24
feat(core): add getYamlInput function

qoomon:
Use Case Example

```yaml
 - uses: foo/bar@v1
    with:
        permissions: |
             foo: admin
             bar: write
```


qoomon:
@rindeal I think you got a point :-D

PR #1776, opened 2024-07-26
fix proxy support for jwks retrieval

bdehamer:
As part of the provenance generation logic, an OIDC token is retrieved from the GHA issuer in order to populate the claims in the attestation. The process of verifying the OIDC token requires that the JSON Web Key Set (JWKS) for the GHA OIDC issuer is retrieved. The library (`jwks-rsa`) we were using to retrieve the key set was not respecting the proxy settings of the actions runner. If a user was using a self-hosted runner behind a proxy the request to retrieve the JWKS would fail.

This change re-implements the JWKS look-up to use the `@actions/http-client` library for the retrieval -- this library has [proxy support](https://github.com/actions/toolkit/tree/main/packages/http-client#features) already included.

Per: https://github.com/actions/attest-build-provenance/issues/156

### DIFF
--- a/packages/attest/RELEASES.md
+++ b/packages/attest/RELEASES.md
@@ -1,5 +1,9 @@
 # @actions/attest Releases
 
+### 1.3.1
+
+- Fix bug with proxy support when retrieving JWKS for OIDC issuer
+
 ### 1.3.0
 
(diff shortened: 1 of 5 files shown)


ejahnGithub (approved):
lgtm!

PR #1790, opened 2024-08-14
support for headers param in attest functions

bdehamer:
Updates the `attest` and `attestProvenance` functions to accept a new, optional `headers` parameter.

If supplied, any headers will be passed-along to the GH attestations API in addition to the required auth header.

### DIFF
--- a/packages/attest/src/attest.ts
+++ b/packages/attest/src/attest.ts
@@ -28,6 +28,8 @@ export type AttestOptions = {
   // Sigstore instance to use for signing. Must be one of "public-good" or
   // "github".
   sigstore?: SigstoreInstance
+  // HTTP headers to include in request to attestations API.
+  headers?: {[header: string]: string | number | undefined}
   // Whether to skip writing the attestation to the GH attestations API.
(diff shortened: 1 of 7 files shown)


packages/attest/src/attest.ts
```diff
@@ -28,6 +28,8 @@ export type AttestOptions = {
   // Sigstore instance to use for signing. Must be one of "public-good" or
   // "github".
   sigstore?: SigstoreInstance
+  // HTTP headers to include in request to attestations API.
+  headers?: {[header: string]: string | number | undefined}
```
kommendorkapten, line 32: Should we have a block list to make sure the client does not set a header that may interfere with any standard headers that are sent?

bdehamer: I looked into this and everything we'd maybe want to block (`user-agent`, `accept`, `content-type`, `authorization`) is already overridden in the `@octokit/request` library. I think we're good to pass-along whatever and let octokit sort it out.

kommendorkapten (approved):

PR #1791, opened 2024-08-15
Add glob option to ignore hidden files

joshmgross:
This PR adds an option to the glob package to exclude hidden files. This includes:
1. Any directory starting with `.`
2. Any file starting with `.`
3. Any files within a directory starting with `.`

This currently *does not* check for the hidden attribute that can be set on Windows files. Based on the discussion in https://github.com/nodejs/node/issues/38809, this isn't natively supported in Node at the moment.
I'm open to working around that limitation, but the primary focus here is for files and directories that use the `.` naming convention.

### DIFF
--- a/packages/glob/__tests__/internal-globber.test.ts
+++ b/packages/glob/__tests__/internal-globber.test.ts
@@ -751,9 +751,7 @@ describe('globber', () => {
     )
 
     const itemPaths = await glob(root, {excludeHiddenFiles: true})
-    expect(itemPaths).toEqual([
-      root
-    ])
+    expect(itemPaths).toEqual([root])
   })
 
   it('returns normalized paths', async () => {
--- a/packages/glob/__tests__/internal-globber.test.ts
+++ b/packages/glob/__tests__/internal-globber.test.ts
@@ -752,7 +752,7 @@ describe('globber', () => {
 
     const itemPaths = await glob(root, {excludeHiddenFiles: true})
     expect(itemPaths).toEqual([
-      root,
+      root
     ])
   })
 
--- a/packages/glob/__tests__/internal-globber.test.ts
+++ b/packages/glob/__tests__/internal-globber.test.ts
@@ -741,7 +741,7 @@ describe('globber', () => {
     //   <root>/.file
     //   <root>/.folder
     //   <root>/.folder/file
-    const root = path.join(getTestTemp(), 'returns-hidden-files')
+    const root = path.join(getTestTemp(), 'ignores-hidden-files')
     await createHiddenDirectory(path.join(root, '.emptyFolder'))
     await createHiddenDirectory(path.join(root, '.folder'))
     await createHiddenFile(path.join(root, '.file'), 'test .file content')
@@ -751,7 +751,9 @@ describe('globber', () => {
     )
 
     const itemPaths = await glob(root, {excludeHiddenFiles: true})
-    expect(itemPaths).toEqual([])
+    expect(itemPaths).toEqual([
+      root,
+    ])
   })
 
   it('returns normalized paths', async () => {
--- a/packages/glob/__tests__/internal-globber.test.ts
+++ b/packages/glob/__tests__/internal-globber.test.ts
@@ -708,7 +708,7 @@ describe('globber', () => {
     expect(itemPaths).toEqual([])
   })
 
-  it('returns hidden files', async () => {
+  it('returns hidden files by default', async () => {
     // Create the following layout:
     //   <root>
     //   <root>/.emptyFolder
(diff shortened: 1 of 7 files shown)


packages/glob/src/internal-glob-options.ts
```diff
@@ -36,4 +36,11 @@ export interface GlobOptions {
    * @default true
    */
   omitBrokenSymbolicLinks?: boolean
+
+  /**
+   * Indicates whether to include hidden files (files and directories starting with a `.`).
+   * 
+   * @default false
```
reemasardana, line 43: Don't we want the default behavior to be to exclude?

packages/glob/src/internal-globber.ts
```diff
@@ -128,6 +128,11 @@ export class DefaultGlobber implements Globber {
         continue
       }
 
+      // Hidden file or directory?
```
reemasardana, line 131: What upload-artifact version(s) is this code path exercised in?

SrRyan: Glob is utilized here https://github.com/actions/upload-artifact/blob/main/src/shared/search.ts

packages/glob/src/internal-glob-options.ts
joshmgross: We want the default to be exclude for the action that consumes this library, I don't think we need or should make a breaking change in this library.

packages/glob/src/internal-glob-options-helper.ts
```diff
@@ -9,7 +9,8 @@ export function getOptions(copy?: GlobOptions): GlobOptions {
     followSymbolicLinks: true,
     implicitDescendants: true,
     matchDirectories: true,
-    omitBrokenSymbolicLinks: true
+    omitBrokenSymbolicLinks: true,
+    excludeHiddenFiles: false,
```
SrRyan, line 13: Just to be clear the idea is leave this to default to have hidden but in the `upload-artifact` default that to having this set to `true` unless specifically set in the command?

packages/glob/src/internal-glob-options.ts
reemasardana: Makes sense for changes to the glob package to be backward compatible but just curious, what other actions use this package? Are there any actions other than upload-artifact where we would want to pass true?

reemasardana: nit: the comment should read "Indicates whether to exclude hidden files.."

packages/glob/src/internal-glob-options-helper.ts
joshmgross: Correct

packages/glob/src/internal-glob-options.ts
joshmgross: > Are there any actions other than upload-artifact where we would want to pass true?

https://github.com/actions/toolkit/network/dependents?owner=actions&dependent_type=REPOSITORY&owner=actions

I see `actions/deploy-pages`, `actions/upload-artifact`, and `actions/download-artifact`.

packages/glob/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/glob",
-  "version": "0.4.0",
```
thboop, line 3: Can you update the releases file if you are going to bump in this pr as well
https://github.com/actions/toolkit/blob/main/packages/glob/RELEASES.md

thboop: No other changes outside of this, some test fixes, and package bumps for security reasons
https://github.com/actions/toolkit/commits/main/packages/glob

thboop (approved):
LGTM

packages/glob/__tests__/internal-globber.test.ts
```diff
@@ -734,6 +734,26 @@ describe('globber', () => {
...
+    //   <root>
+    //   <root>/.emptyFolder
+    //   <root>/.file
+    //   <root>/.folder
+    //   <root>/.folder/file
+    const root = path.join(getTestTemp(), 'ignores-hidden-files')
+    await createHiddenDirectory(path.join(root, '.emptyFolder'))
+    await createHiddenDirectory(path.join(root, '.folder'))
+    await createHiddenFile(path.join(root, '.file'), 'test .file content')
+    await fs.writeFile(
```
ramblingenzyme, line 748: I'm a little confused about the behaviour here.

I'm not sure if the behaviour for the following globs matches what's intended
* `ignores-hidden-files/.folder` - returns nothing
* `ignores-hidden-files/.folder/*` - returns `file`
* `ignores-hidden-files/.folder/file` - returns `file`

PR #1793, opened 2024-08-15
Include the package name in the Publish Workflow run

joshmgross:
This will give each run a name to help devs understand at a glance which package was published when browsing https://github.com/actions/toolkit/actions/workflows/releases.yml instead of just seeing "Publish NPM" over and over.


thboop (approved):
lgtm

Issue #1795, opened 2024-08-16
@actions/attest should default OIDC issuer based on GitHub tenant


bdehamer:
When invoking `attestProvenance` there is an optional `issuer` parameter to defaults to the URL for OIDC issuer in the primary GitHub tenant ("https://token.actions.githubusercontent.com"). When running in a different tenant, the caller must derive and pass-in the correct issuer value so that the retrieved OIDC token can be properly validated.

It would be more convenient to the caller, if the default issuer was derived automatically from the current execution context.

PR #1796, opened 2024-08-16
derive default OIDC issuer from current tenant

bdehamer:
Closes: https://github.com/actions/toolkit/issues/1795

Updates the attestation logic so that the default OIDC issuer is derived from the current GH tenant context. 

This change should mean that most callers can omit the `issuer` parameter when calling `buildSLSAProvenancePredicate`/`attestProvenance` and rely on the library to select the correct OIDC issuer.


pwelch (approved):
LGTM

PR #1561, opened 2023-10-16
fix: move `undici` to `devDependencies`.

Jason3S:
fixes #1560

PR #1800, opened 2024-08-17
fix: lazy load ProxyAgent to avoid bundling all the undici

### LINKED ISSUE #1560: `@actions/http-client` - unnecessary dependency upon `undici`
When updating today, I noticed that size of my action bloated by 15k lines because `undici` was included as a dependency.

`undici` should have been in `devDependencies`, but was accidentally included in `dependencies`.

_Originally posted by @Jason3S in https://github.com/actions/toolkit/pull/1547#discussion_r1360354392_

aminya:
This makes the loading of undici conditional by lazy loading it inside the actual function that needs the ProxyAgent. 

undici is a huge library, and here it is being imported unconditionally, which results in 500KB bloat in the bundles of http-client. By requiring only what is needed and when it is needed, this issue is resolved.

Fixes #1560 
Closes #1561

### DIFF
--- a/packages/http-client/src/index.ts
+++ b/packages/http-client/src/index.ts
@@ -6,7 +6,7 @@ import * as ifm from './interfaces'
 import * as net from 'net'
 import * as pm from './proxy'
 import * as tunnel from 'tunnel'
-import {ProxyAgent} from 'undici'
+import type ProxyAgent from 'undici/types/proxy-agent'
 
 export enum HttpCodes {
   OK = 200,
@@ -722,6 +722,12 @@ export class HttpClient {
     }
 
     const usingSsl = parsedUrl.protocol === 'https:'
+
+    // Lazy load ProxyAgent to avoid bundling all the undici
+    const ProxyAgent =
+      // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+      require('undici/lib/proxy-agent') as typeof import('undici/types/proxy-agent').default
+
     proxyAgent = new ProxyAgent({
       uri: proxyUrl.href,
       pipelining: !this._keepAlive ? 0 : 1,


packages/http-client/src/index.ts
```diff
@@ -722,6 +722,12 @@ export class HttpClient {
     }
 
     const usingSsl = parsedUrl.protocol === 'https:'
+
+    // Lazy load ProxyAgent to avoid bundling all the undici
+    const ProxyAgent =
+      // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
+      require('undici/lib/proxy-agent') as typeof import('undici/types/proxy-agent').default
```
aminya, line 729: This can be further improved by using dynamic imports. However, I didn't include it in this pull request, because using dynamic imports needs making `getAgentDispatcher` async, which changes the return type of a public method. Based on my research, this is only used in the github package, so it should be fine, but if we want to go with that, we need to have a major version bump for http-client. 
https://github.com/aminya/toolkit/commit/b716d6f9619c1b31f415ccd849abe1374f20a51c

packages/http-client/src/index.ts
```diff
@@ -6,7 +6,7 @@ import * as ifm from './interfaces'
 import * as net from 'net'
 import * as pm from './proxy'
 import * as tunnel from 'tunnel'
-import {ProxyAgent} from 'undici'
+import type ProxyAgent from 'undici/types/proxy-agent'
```
benmccann, line 9: is this line doing anything? I expect it should be shadowed below and so could be removed

aminya: It's the return type of the method

benmccann: oh yeah. I see it used outside the diff now. nevermind :smile:  sorry for the distraction

lwmacct:
This PR would let us remove a downstream workaround. Adding a concrete downstream measurement from a Rollup-packaged GitHub Action.

Dependency chain in the action:
- `@actions/cache@6.1.0`
- `@actions/core@3.0.1`
- `@actions/tool-cache@4.0.0`
- all resolve to `@actions/http-client@4.0.1`

The action never calls `HttpClient#getAgentDispatcher()`, but the top-level `import {ProxyAgent} from 'undici'` in `@actions/http-client` makes Rollup include the full `undici` graph.

Measured on 2026-07-07:

```text
with a Rollup stub for undici.ProxyAgent:
  dist total:        755,357 bytes
  dist tar+gzip:     198,613 bytes
  chunks/undici.js:  absent
  build time:        10.4s

with real undici bundled:
  dist total:        1,160,472 bytes
  dist tar+gzip:     328,653 bytes
  chunks/undici.js:  404,378 bytes raw / 129,724 bytes gzip
  build time:        15s

increase from the static undici import:
  +405,115 bytes raw dist
  +130,040 bytes tar+gzip dist
```

Current downstream workaround: a Rollup virtual module that exports a `ProxyAgent` class which throws if constructed. That preserves the normal `@actions/http-client` Node `http` / `https` agent path and the existing `tunnel` proxy path, while failing explicitly if the dispatcher path is ever used.

A better upstream fix would be to avoid importing `undici` at module top level and only load `ProxyAgent` inside the dispatcher path, or expose dispatcher support through a separate entrypoint.

Issue #1614, opened 2023-12-20
`@actions/artifact` dependency `crypto` is deprecated


muzimuzhi:
**Describe the bug**

Installing `@actions/artifact` gives
```
npm WARN deprecated crypto@1.0.1: This package is no longer supported. It's now a built-in
Node module. If you've depended on crypto, you should switch to the one that's built-in.
```

See also https://www.npmjs.com/package/crypto.

**To Reproduce**

Run `npm install` in `./packages/artifact`.

pauliesnug:
I can reproduce this as well. We need to switch all `crypto.createHash('sha256')` calls to use `node:crypto`

pngwn:
`crypto` can't be resolved when bundling either. This makes the package unusable currently.

PR #1815, opened 2024-08-30
Use native `crypto` package from node

mydea:
Closes https://github.com/actions/toolkit/issues/1614

As far as I can tell, there is no need to actually have this dependency anymore, it should just use the native module instead. This trips up certain bundlers, probably because the package is not really correct - it refers to a `index.js` entrypoint that does not exist (I suspect node does some internal stuff for this??) - see https://www.npmjs.com/package/crypto?activeTab=code

### DIFF
--- a/packages/artifact/package-lock.json
+++ b/packages/artifact/package-lock.json
@@ -19,7 +19,6 @@
         "@octokit/request-error": "^5.0.0",
         "@protobuf-ts/plugin": "^2.2.3-alpha.1",
         "archiver": "^7.0.1",
-        "crypto": "^1.0.1",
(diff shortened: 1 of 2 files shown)


JoannaaKL:
Thank you for this change. Could you also please update the import in https://github.com/actions/toolkit/blob/2e1998fc4227c81d7ab8e491ef7126cd6ad3c17e/packages/artifact/src/internal/upload/blob-upload.ts#L10 from `import * as crypto from 'crypto'` to `import { createHash } from 'node:crypto'`? :)

JoannaaKL (approved):

Issue #925, opened 2021-10-22
Avoid using deprecated version of UUID


aeisenberg:
All versions of the UUID npm package < 7 have [been deprecated](https://github.com/uuidjs/uuid/issues/572) due to a weakness in `Math.random()`. There are two packages in this repository that use the deprecated version.

- https://github.com/actions/toolkit/blob/main/packages/tool-cache/package.json#L44
- https://github.com/actions/toolkit/blob/main/packages/cache/package.json#L48

Any downstream package that depends on these packages will include the following line in their `package-lock.json`:

```
      "deprecated": "Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.",
```

Can you upgrade node UUID to the latest released version? Happy to submit a PR to help you out.

stevenwdv:
@maintainers?

MikeMcC399:
`@actions/cache@3.2.2` (latest) is still using a deprecated version `uuid@3.4.0 `

```text
$ npm ls uuid
cache-test@1.0.0 D:\github2\cache-test
└─┬ @actions/cache@3.2.2
  ├─┬ @actions/core@1.10.1
  │ └── uuid@8.3.2
  ├─┬ @azure/ms-rest-js@2.7.0
  │ └── uuid@8.3.2
  ├─┬ @azure/storage-blob@12.17.0
  │ └─┬ @azure/core-http@3.0.4
  │   └── uuid@8.3.2
  └── uuid@3.4.0
```

- PR #1571 was submitted and is waiting for review and merge

MikeMcC399:
`@actions/cache@3.2.3` also uses the deprecated version `uuid@3.4.0 `

Could this issue please be addressed?

MikeMcC399:
`@actions/cache@3.2.4` has been released, however this issue still remains unresolved.

jakebailey:
FWIW the dep isn't even needed; all uses of the library in the repo are just generating UUIDv4s, and Node has had `crypto.randomUUID()` as a global since Node 16.7 (and even earlier on the crypto module): https://nodejs.org/docs/latest/api/webcrypto.html#cryptorandomuuid

MikeMcC399:
Do the packages `tool-cache` and `cache` in this repo have a designated maintainer and is there any hope that this issue can be resolved?

joshmgross:
[...]

PR #1824, opened 2024-09-04
Remove dependency on `uuid` package

joshmgross:
- Supports https://github.com/actions/toolkit/issues/925

A common pattern in the toolkit is to create temporary file or directory within the existing temp directory using a random UUID.

Versions of the `uuid` package below `v7` are deprecated, so we shouldn't depend on them. Additionally, Node has a built-in [`crypto.randomUUID()`](https://nodejs.org/docs/latest-v20.x/api/crypto.html#cryptorandomuuidoptions) method that can be used to generate UUIDs without the need for an external package. This function was introduced in versions of Node 14 and 15, so we should be safe to use it for actions that depend on both Node 16 and 20 (which are our only supported versions at this time).

I also went ahead and updated `@actions/core`, even though it was using a non-deprecated version of the `uuid` package.

Since these use cases is purely for temporary files, I don't think we strictly need a cryptographically secure UUID but that's an added bonus of using the built-in Node function.

Packages *shouldn't* be depending on the exact format of this temp file/directory, but even if they were this should be a compatible as we're still generating a version 4 UUID

> Generates a random [RFC 4122](https://www.rfc-editor.org/rfc/rfc4122.txt) version 4 UUID. The UUID is generated using a cryptographic pseudorandom number generator.

There are a couple other packages within the toolkit that have an indirect dependency on `uuid` through `@actions/core`, those can be updated as well once we release this new version of `@actions/core`.

### DIFF
--- a/packages/cache/package-lock.json
+++ b/packages/cache/package-lock.json
@@ -17,12 +17,10 @@
         "@azure/abort-controller": "^1.1.0",
         "@azure/ms-rest-js": "^2.6.0",
         "@azure/storage-blob": "^12.13.0",
-        "semver": "^6.3.1",
-        "uuid": "^3.3.3"
+        "semver": "^6.3.1"
       },
       "devDependencies": {
         "@types/semver": "^6.0.0",
-        "@types/uuid": "^3.4.5",
         "typescript": "^5.2.2"
       }
     },
@@ -296,12 +294,6 @@
         "@types/node": "*"
       }
     },
-    "node_modules/@types/uuid": {
-      "version": "3.4.10",
-      "resolved": "https://registry.npmjs.org/@types/uuid/-/uuid-3.4.10.tgz",
-      "integrity": "sha512-BgeaZuElf7DEYZhWYDTc/XcLZXdVgFkVSTa13BqKvbnmUrxr3TJFKofUxCtDO9UQOdhnV+HPOESdHiHKZOJV1A==",
-      "dev": true
-    },
     "node_modules/abort-controller": {
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/abort-controller/-/abort-controller-3.0.0.tgz",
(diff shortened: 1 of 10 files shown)


MikeMcC399:
@joshmgross 

Are you going to be able to progress this PR soon? It looks like you are just waiting for your team to review.

robherley (approved):

luketomlinson (approved):

gfteix:
With the crypto global package is being used, if the client is using a node version < 19 (when crypto became a global package in node.js) the action that uses the toolkit will fail with a `crypt is not defined error`.

MarioUhrikTakeda:
This seems to have caused https://github.com/actions/toolkit/issues/1841
FYI @joshmgross

MikeMcC399:
@joshmgross 

Just to be sure, are you (or somebody else) going to release a new version of `@actions/cache` now?
image

joshmgross:
@MikeMcC399 yes I'm planning to upgrade the packages that depend on `@actions/core`.

It's worth noting that `@actions/cache` will still depend on `uuid` through `@azure/core-http` though:
```
❯ npm why uuid
uuid@8.3.2
node_modules/@azure/core-http/node_modules/uuid
  uuid@"^8.3.0" from @azure/core-http@3.0.2
  node_modules/@azure/core-http
    @azure/core-http@"^3.0.0" from @azure/storage-blob@12.15.0
    node_modules/@azure/storage-blob
      @azure/storage-blob@"^12.13.0" from the root project

uuid@8.3.2
node_modules/@azure/ms-rest-js/node_modules/uuid
  uuid@"^8.3.2" from @azure/ms-rest-js@2.7.0
  node_modules/@azure/ms-rest-js
    @azure/ms-rest-js@"^2.6.0" from the root project
```

MikeMcC399:
@joshmgross 

>  yes I'm planning to upgrade the packages that depend on `@actions/core`.
> 
> It's worth noting that `@actions/cache` will still depend on `uuid` through `@azure/core-http` though:

Thanks for the confirmation! Looking forward to new releases which no longer depend on a **deprecated version** of `uuid`.

jsoref:
When do you intend to release a version of `@actions/tool-cache` newer than `"version": "2.0.1"` to include this fix?

joshmgross:
@jsoref see https://github.com/actions/toolkit/pull/1872, which is waiting on approval

PR #1825, opened 2024-09-09
Update artifact FAQ with details about security advisory

thyeggman:
Some users have been seeing dependabot update `actions/download-artifact` to a version which has breaking changes due to a [security advisory](https://github.com/actions/download-artifact/security/advisories/GHSA-cxww-7g56-2vh6). The error they see in the job output points to this FAQ, which I'm updating to include details about the failure.


thyeggman:
The security advisory originally had the wrong range of affected versions (<4.1.7).  The actual affected range is 4.0.0-4.1.6.  Originally there was a small number of failures due to dependabot trying to update old versions that were actually unaffected, which is why I wanted to add the note to the FAQ.  However, now since the affected versions are correctly listed as v4, there will be no incompatibility, so the there should be no more need for this message 🙂

PR #1826, opened 2024-09-11
exec: fix outdated link in README

rindeal:
The link was a permalink, so it pointed to a very outdated version of the file.


joshmgross (approved):
Thanks!

PR #1830, opened 2024-09-20
Fix regression, auto readlink on symlinks again

robherley:
With the changes introduced in:
- https://github.com/actions/toolkit/pull/1771

Which was meant to resolve:
- https://github.com/actions/upload-artifact/issues/485

We introduced a new bug:
- https://github.com/actions/upload-artifact/issues/590

We had a slight regression with symlinks. The typical behavior is that they are auto-resolved and the contents of the zip are added at the link's location. With the lazy-stream wrapper being used, it did not resolve the symlinks.

This PR switches our `stat` to an `lstat`, and we can check for a symbolic link, then resolve it, when we add the file to the zip archive.

### DIFF
--- a/packages/artifact/src/internal/upload/upload-zip-specification.ts
+++ b/packages/artifact/src/internal/upload/upload-zip-specification.ts
@@ -13,6 +13,12 @@ export interface UploadZipSpecification {
    * The destination path in a zip for a file
    */
   destinationPath: string
+
+  /**
+   * Information about the file
+   * https://nodejs.org/api/fs.html#class-fsstats
+   */
+  stats: fs.Stats
 }
 
 /**
@@ -75,10 +81,11 @@ export function getUploadZipSpecification(
           - file3.txt
   */
   for (let file of filesToZip) {
-    if (!fs.existsSync(file)) {
+    const stats = fs.lstatSync(file, {throwIfNoEntry: false})
+    if (!stats) {
       throw new Error(`File ${file} does not exist`)
     }
-    if (!fs.statSync(file).isDirectory()) {
+    if (!stats.isDirectory()) {
       // Normalize and resolve, this allows for either absolute or relative paths to be used
       file = normalize(file)
       file = resolve(file)
@@ -94,7 +101,8 @@ export function getUploadZipSpecification(
 
       specification.push({
         sourcePath: file,
-        destinationPath: uploadPath
+        destinationPath: uploadPath,
+        stats
       })
     } else {
       // Empty directory
@@ -103,7 +111,8 @@ export function getUploadZipSpecification(
 
       specification.push({
         sourcePath: null,
-        destinationPath: directoryPath
+        destinationPath: directoryPath,
+        stats
       })
     }
   }
--- a/packages/artifact/RELEASES.md
+++ b/packages/artifact/RELEASES.md
@@ -1,5 +1,9 @@
 # @actions/artifact Releases
 
+### 2.1.10
+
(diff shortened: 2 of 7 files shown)


packages/artifact/src/internal/upload/upload-zip-specification.ts
```diff
@@ -75,10 +81,11 @@ export function getUploadZipSpecification(
           - file3.txt
   */
   for (let file of filesToZip) {
-    if (!fs.existsSync(file)) {
+    const stats = fs.lstatSync(file, {throwIfNoEntry: false})
```
robherley, line 84: mildly important part: we switched from `statSync` to [`lstatSync`](https://nodejs.org/api/fs.html#fs_fs_lstatsync_path_options)

> similar to `stat` unless `path` refers to a symbolic link, in which case the link itself is stat-ed, not the file that it refers to

we'll use these stats to check [`isSymbolicLink`](https://nodejs.org/api/fs.html#fs_stats_issymboliclink) when we add the file to the zip, and resolve the target's contents.

konradpabjan (approved):

packages/artifact/src/internal/upload/upload-zip-specification.ts
```diff
@@ -75,10 +81,11 @@ export function getUploadZipSpecification(
...
+    if (!stats) {
       throw new Error(`File ${file} does not exist`)
     }
-    if (!fs.statSync(file).isDirectory()) {
+    if (!stats.isDirectory()) {
```
joshmgross, line 88: Can you have a symbolic link to a directory?
Do we need a normal stat to check this?

packages/artifact/src/internal/upload/upload-zip-specification.ts
```diff
@@ -13,6 +13,12 @@ export interface UploadZipSpecification {
    * The destination path in a zip for a file
    */
   destinationPath: string
+
+  /**
+   * Information about the file
+   * https://nodejs.org/api/fs.html#class-fsstats
+   */
+  stats: fs.Stats
```
joshmgross, line 21: Since this is in the `src/internal` directory, does that mean this interface won't be used by consumers of this package?

Would it make sense to export the minimal information we need rather than the entire `fs.Stats` interface?

joshmgross (approved):

packages/artifact/src/internal/upload/upload-zip-specification.ts
robherley: The reason why I added the `fs.Stats` interface is if we eventually want to add file permissions, see this PR: 
- https://github.com/actions/toolkit/pull/1723

The stats and mode can eventually be passed to the archiver library. I didn't want to do that _yet_ to avoid additional changes.

packages/artifact/src/internal/upload/upload-zip-specification.ts
robherley: > Can you have a symbolic link to a directory?

Yes you can!

> Do we need a normal stat to check this?

I don't think so. As long as we know it's a directory that's all that matters. We add the symlink's path as the `destinationPath` below will a `sourcePath` of `null`, so when we create the directory in the zip, it's just a normal directory.

Directory symlinks don't need to be explicitly followed since [our globber will do that for us](https://github.com/actions/upload-artifact/blob/3eadd8b791cabf7cd572f194da82158c24125bd8/src/shared/search.ts#L16) when traversing the files. So as long as we know that it's a directory (symlink or not), that's all that matters.

PR #1832, opened 2024-09-25
catching error message from github that's in Json format

Lavaerius:
Description:

Updates fail condition for getManifestFromRepo function to include instances when github responds with JSON body for API rate limiting, that is incorrectly handled.
Related issue:

https://github.com/actions/setup-python/issues/903

### DIFF
--- a/packages/tool-cache/src/tool-cache.ts
+++ b/packages/tool-cache/src/tool-cache.ts
@@ -627,8 +627,11 @@ export async function getManifestFromRepo(
       core.debug('Invalid json')
     }
   }
-
-  return releases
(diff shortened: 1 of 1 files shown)


packages/tool-cache/src/tool-cache.ts
```diff
@@ -627,8 +627,11 @@ export async function getManifestFromRepo(
       core.debug('Invalid json')
     }
   }
-
-  return releases
+  if (!releases.hasOwnProperty('documentation_url')) {
```
luketomlinson, line 630: 👋 Perhaps we could check for the rate-limiting response code (429) and return an error instead of relying on the presence of the `documentation_url` property?

PR #1835, opened 2024-09-30
`@actions/cache`: using concurrent download whenever server supported

dungdm93:
The concurrent download cache is currently limited to Azure Blob Storage, requiring users to enable it by setting `DownloadOptions.concurrentBlobDownloads`.

However, since `S3`, `GCS`, and various file servers also allow partial downloads, it would be beneficial for the cache client to automatically detect server support and utilize concurrent downloads when available.


Jura45 (approved):

joshmgross:
@dungdm93 could you file an issue detailing your use case?

Outside of GitHub Enterprise Server, we don't support other storage providers besides Azure Blob Storage.

Issue #1841, opened 2024-10-02
@actions/core version 1.11.0 has error ReferenceError: crypto is not defined


jessecollier:
**Describe the bug**
We have a github action that runs in the following environment:

```
Current runner version: '2.319.1'
Operating System
  Ubuntu
  [2](https://github.com/dynamic-labs/dynamic-auth/actions/runs/11149638521/job/30988996400#step:1:2)2.04.5
  LTS
Runner Image
  Image: ubuntu-22.04
  Version: 20240922.1.0
  Included Software: https://github.com/actions/runner-images/blob/ubuntu22/20240922.1/images/ubuntu/Ubuntu2204-Readme.md
  Image Release: https://github.com/actions/runner-images/releases/tag/ubuntu22%2F20240922.1
Runner Image Provisioner
  2.0.[3](https://github.com/dynamic-labs/dynamic-auth/actions/runs/11149638521/job/30988996400#step:1:3)84.1
```

**To Reproduce**
1. Create a js file that imports `const core = require('@actions/core');`
2. Add it to your github action

```
/home/runner/work/_actions/<redacted>/main/.github/actions/release-helper/node_modules/@actions/core/lib/file-command.js:47
    const delimiter = `ghadelimiter_${crypto.randomUUID()}`;
                                      ^

ReferenceError: crypto is not defined
    at prepareKeyValueMessage (/home/runner/work/_actions/<redacted>/main/.github/actions/release-helper/node_modules/@actions/core/lib/file-command.js:47:39)
    at Object.setOutput (/home/runner/work/_actions/<redacted>/main/.github/actions/release-helper/node_modules/@actions/core/lib/core.js:168:105)
    at setOutput (/home/runner/work/_actions/<redacted>/main/.github/actions/release-helper/util.js:8:8)
    at Object.<anonymous> (/home/runner/work/_actions/<redacted>/main/.github/actions/release-helper/getVersion.js:20:1)
    at Module._compile (node:internal/modules/cjs/loader:1364:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1422:10)
    at Module.load (node:internal/modules/cjs/loader:1203:32)
    at Module._load (node:internal/modules/cjs/loader:1019:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:[128](https://github.com/<redacted>/actions/runs/11149638521/job/30988996400#step:2:139):12)
    at node:internal/main/run_main_module:28:49
```

[...]

PR #1842, opened 2024-10-04
Explicitly import `crypto`

joshmgross:
https://github.com/actions/toolkit/issues/1841

`crypto` has to be explicitly imported in versions before Node 19, so #1824 broke using these packages.

I've validated that importing `crypto` works as expected on Node 16, 18, and 20:

https://github.com/actions/toolkit/issues/1841#issuecomment-2394474883


luketomlinson (approved):

yacaovsnc (approved):

reemasardana:
> I've validated that importing crypto works as expected on Node 16, 18, and 20:

Were these manual tests? Anything we can easily automate to prevent future accidental breakages like this?

joshmgross:
> > I've validated that importing crypto works as expected on Node 16, 18, and 20:
> 
> Were these manual tests? Anything we can easily automate to prevent future accidental breakages like this?

We can run tests against other Node versions besides the current support action Node version. I've validated that this catches the issue described in #1841 - #1843.

In the future we should be clear about what Node versions we support for the toolkit, but until then I think it's reasonable to validate that the toolkit at least works with what users currently expect.

yacaovsnc (approved):

PR #1855, opened 2024-10-16
Allow passing non-async function to group

Cyberbeni:
The current version would give a `@typescript-eslint/require-await` error when there is no await inside and wouldn't accept a non-async block. This allows passing a non-async block (although still leaving the group function async, I don't know if it's possible to tie those together in TypeScript)


Cyberbeni:
ping @nebuk89

PR #1857, opened 2024-10-24
Integrate cache service v2

Link-:
Upgrading the cache package to integrate with the new cache service backend (cache service v2). Major changes to the API layer are involved with this change but the interfaces are backward compatible. The transition to the new service will undergo gradual rollout and will be seamless to end users.

### More context

[Deprecation notice for v1 and v2 of actions/cache](https://github.blog/changelog/2024-09-16-notice-of-upcoming-deprecations-and-changes-in-github-actions-services/)


packages/cache/src/internal/config.ts
```diff
@@ -0,0 +1,24 @@
...
+  if (!token) {
+    throw new Error('Unable to get the ACTIONS_RUNTIME_TOKEN env variable')
+  }
+  return token
+}
+
+// TODO: Use the feature flag to determine the cache service version
+export function getCacheServiceVersion(): string {
+  return process.env['ACTIONS_CACHE_SERVICE_VERSION'] || 'v1'
+}
```
robherley, line 12: when this is fixed, we should include some logic to check for results env too. we don't want to end up removing the feature flag after ship and attempting to hit v2 in GHES

packages/cache/package.json
```diff
@@ -38,6 +38,7 @@
   },
   "dependencies": {
     "@actions/core": "^1.10.0",
+    "@actions/artifact": "^2.1.7",
```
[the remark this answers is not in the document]
Link-: Ideally yes. For now, I'm just going to duplicate the logic so that we can decrease the complexity of the rollout

Liabaemt (commented):
Fix

Liabaemt:
H

Liabaemt (approved):

packages/cache/src/cache.ts
```diff
@@ -153,6 +203,106 @@ export async function restoreCache(
...
+
+    await extractTar(archivePath, compressionMethod)
+    core.info('Cache restored successfully')
+
+    return request.key
+  } catch (error) {
+    throw new Error(`Failed to restore: ${error.message}`)
+  } finally {
+    try {
+      await utils.unlinkFile(archivePath)
```
joshmgross, line 299:
```suggestion
      if (archivePath) {
        await utils.unlinkFile(archivePath)
      }
```

It's possible for this `archivePath` to not yet be set when we get here

packages/cache/src/cache.ts
```diff
@@ -153,6 +203,106 @@ export async function restoreCache(
...
+      key: primaryKey,
+      restoreKeys,
+      version: utils.getCacheVersion(
+        paths,
+        compressionMethod,
+        enableCrossOsArchive
+      )
+    }
+
+    const response: GetCacheEntryDownloadURLResponse =
```
joshmgross, line 256: These Twirp methods are typed, so we don't need to define the type here on `response`

```diff
diff --git a/packages/cache/src/cache.ts b/packages/cache/src/cache.ts
index 7d0cd00..2682b06 100644
--- a/packages/cache/src/cache.ts
+++ b/packages/cache/src/cache.ts
@@ -253,9 +253,7 @@ async function restoreCachev2(
       )
     }
 
-    const response: GetCacheEntryDownloadURLResponse =
-      await twirpClient.GetCacheEntryDownloadURL(request)
-
+    const response = await twirpClient.GetCacheEntryDownloadURL(request)
     if (!response.ok) {
       core.warning(`Cache not found for keys: ${keys.join(', ')}`)
       return undefined
@@ -503,8 +501,7 @@ async function saveCachev2(
       key,
       version
     }
-    const response: CreateCacheEntryResponse =
-      await twirpClient.CreateCacheEntry(request)
+    const response = await twirpClient.CreateCacheEntry(request)
     if (!response.ok) {
       throw new ReserveCacheError(
         `Unable to reserve cache with key ${key}, another job may be creating this cache.`

```

package.json
```diff
@@ -32,5 +32,19 @@
     "prettier": "^3.0.0",
     "ts-jest": "^29.1.1",
     "typescript": "^5.2.2"
+  },
+  "dependencies": {
```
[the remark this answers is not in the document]
Link-: These shouldn't be here 🤔

packages/cache/package.json
```diff
@@ -45,10 +45,13 @@
     "@azure/abort-controller": "^1.1.0",
     "@azure/ms-rest-js": "^2.6.0",
     "@azure/storage-blob": "^12.13.0",
-    "semver": "^6.3.1"
+    "@protobuf-ts/plugin": "^2.9.4",
```
Link-, line 48: These have been moved from the top level `package.json`. And `@protobuf-ts/plugin` was bumped up to `2.9.4`. Not sure why were using (and still using) `@protobuf-ts/plugin": "^2.2.3-alpha.1` in artifacts

packages/cache/src/cache.ts
```diff
@@ -153,6 +203,106 @@ export async function restoreCache(
...
+      core.info('Lookup only - skipping download')
+      return request.key
+    }
+
+    archivePath = path.join(
+      await utils.createTempDirectory(),
+      utils.getCacheFileName(compressionMethod)
+    )
+    core.debug(`Archive path: ${archivePath}`)
+    core.debug(`Starting download of artifact to: ${archivePath}`)
```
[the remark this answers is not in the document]
Link-: `cache` is correct, good catch

packages/cache/src/cache.ts
Link-: Good call

packages/cache/src/cache.ts
```diff
@@ -153,6 +203,106 @@ export async function restoreCache(
...
+ * Restores cache using the new Cache Service
+ *
+ * @param paths a list of file paths to restore from the cache
+ * @param primaryKey an explicit key for restoring the cache
+ * @param restoreKeys an optional ordered list of keys to use for restoring the cache if no cache hit occurred for key
+ * @param downloadOptions cache download options
+ * @param enableCrossOsArchive an optional boolean enabled to restore on windows any cache created on any platform
+ * @returns string returns the key for the cache hit, otherwise returns undefined
+ */
+async function restoreCachev2(
```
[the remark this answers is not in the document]
Link-: I think it's easier to reason about what each function is doing this way. We don't have to couple the behaviour of each with more shared functions. The utility functions are already extracted, and what remains is the procedural implementation.

packages/cache/src/cache.ts
```diff
@@ -153,6 +203,106 @@ export async function restoreCache(
...
+
+  let archivePath = ''
+  try {
+    const twirpClient = cacheTwirpClient.internalCacheTwirpClient()
+    const backendIds: utils.BackendIds = utils.getBackendIdsFromToken()
+    const compressionMethod = await utils.getCompressionMethod()
+
+    const request: GetCacheEntryDownloadURLRequest = {
+      workflowRunBackendId: backendIds.workflowRunBackendId,
+      workflowJobRunBackendId: backendIds.workflowJobRunBackendId,
```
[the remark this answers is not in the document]
Link-: Addressed in: https://github.com/github/actions-relaunch/issues/994

packages/cache/src/internal/cacheUtils.ts
```diff
@@ -143,3 +146,104 @@ export function isGhes(): boolean {
...
+  return crypto.createHash('sha256').update(components.join('|')).digest('hex')
+}
+
+export function getRuntimeToken(): string {
+  const token = process.env['ACTIONS_RUNTIME_TOKEN']
+  if (!token) {
+    throw new Error('Unable to get the ACTIONS_RUNTIME_TOKEN env variable')
+  }
+  return token
+}
```
johnsudol, line 181: Do we need this function both in here and in https://github.com/actions/toolkit/blob/8616c313a26622e237804459fff2308978539e20/packages/cache/src/internal/config.ts#L1-L7

Link-: Great catch, thank you

packages/artifact/package-lock.json
```diff
@@ -839,9 +839,10 @@
       }
     },
     "node_modules/cross-spawn": {
-      "version": "7.0.3",
-      "resolved": "https://registry.npmjs.org/cross-spawn/-/cross-spawn-7.0.3.tgz",
-      "integrity": "sha512-iRDPJKUPVEND7dHPO8rkbOnPpyDygcDFtWjpeWNCgy8WP2rXcxXL8TskReQl6OrB2G7+UJrags1q15Fudc7G6w==",
+      "version": "7.0.6",
```
Link-, line 842: This was needed to resolve the audit failures

packages/cache/src/internal/shared/cacheTwirpClient.ts
robherley, line 1: Ideally, it would be great if we could share this client between the artifact and cache packages

packages/cache/src/internal/blob/download-cache.ts
```diff
@@ -0,0 +1,30 @@
+import * as core from '@actions/core'
+
+import {
+  BlobClient,
+  BlockBlobClient,
+  BlobDownloadOptions
+} from '@azure/storage-blob'
+
+export async function DownloadCacheFile(
```
robherley, line 9: nit: no `PascalCase` for functions, we tend to use lower camel case

packages/cache/src/internal/blob/upload-cache.ts
```diff
@@ -0,0 +1,27 @@
+import * as core from '@actions/core'
+import {
+  BlobClient,
+  BlockBlobClient,
+  BlockBlobParallelUploadOptions
+} from '@azure/storage-blob'
+
+export async function UploadCacheFile(
```
robherley, line 8: nit: `uploadCacheFile`

packages/cache/src/internal/blob/upload-cache.ts
```diff
@@ -0,0 +1,27 @@
...
+  signedUploadURL: string,
+  archivePath: string
+): Promise<{}> {
+  // Specify data transfer options
+  const uploadOptions: BlockBlobParallelUploadOptions = {
+    blockSize: 4 * 1024 * 1024, // 4 MiB max block size
+    concurrency: 4, // maximum number of parallel transfer workers
+    maxSingleShotSize: 8 * 1024 * 1024 // 8 MiB initial transfer size
+  }
```
robherley, line 17: were these options derived from defaults?

packages/cache/src/cache.ts
```diff
@@ -153,6 +201,104 @@ export async function restoreCache(
...
+ * Restores cache using the new Cache Service
+ *
+ * @param paths a list of file paths to restore from the cache
+ * @param primaryKey an explicit key for restoring the cache
+ * @param restoreKeys an optional ordered list of keys to use for restoring the cache if no cache hit occurred for key
+ * @param downloadOptions cache download options
+ * @param enableCrossOsArchive an optional boolean enabled to restore on windows any cache created on any platform
+ * @returns string returns the key for the cache hit, otherwise returns undefined
+ */
+async function restoreCacheV2(
```
robherley, line 214: for the `*CacheV2` functions, do we want to throw `GHESNotSupportedError`?

robherley (approved):
non-blocking comments, 👍 given we'll follow up in future PRs with more test coverage before release

packages/cache/src/internal/blob/upload-cache.ts
Link-: Yep, we need to tune them as we go

packages/cache/src/internal/config.ts
```diff
@@ -0,0 +1,39 @@
...
+  const isGheHost = hostname.endsWith('.GHE.COM')
+  const isLocalHost = hostname.endsWith('.LOCALHOST')
+
+  return !isGitHubHost && !isGheHost && !isLocalHost
+}
+
+export function getCacheServiceVersion(): string {
+  // Cache service v2 is not supported on GHES. We will default to
+  // cache service v1 even if the feature flag was enabled by user.
+  if (isGhes()) return 'v1'
```
Link-, line 17: @robherley to handle GHES better, we always default to v1

packages/cache/src/cache.ts
Link-: I added a check to always default to v1 on GHES

packages/cache/src/cache.ts
```diff
@@ -153,6 +201,104 @@ export async function restoreCache(
...
+    await extractTar(archivePath, compressionMethod)
+    core.info('Cache restored successfully')
+
+    return request.key
+  } catch (error) {
+    throw new Error(`Failed to restore: ${error.message}`)
+  } finally {
+    try {
+      if (archivePath) {
+        await utils.unlinkFile(archivePath)
```
ericsciple, line 294: do you need to handle file not found error here?

packages/cache/src/cache.ts
```diff
@@ -258,3 +429,111 @@ export async function saveCache(
...
+    }
+
+    cacheId = parseInt(finalizeResponse.entryId)
+  } catch (error) {
+    const typedError = error as Error
+    core.warning(`Failed to save: ${typedError.message}`)
+  } finally {
+    // Try to delete the archive to save space
+    try {
+      await utils.unlinkFile(archivePath)
```
ericsciple, line 532: same question here: do you need to catch file-not-found error?

ericsciple (approved):

PR #1860, opened 2024-10-28
Update default retry options for http client for Artifacts 

SrRyan:
Updating default retry options for the Artifacts http client. This is to help alleviate job failures due to upstream platform restrictions when using the artifact action during a job. 

Rough math based on the code involve here and the backoff using `30,000ms` as the base retry with a multiplier of `1.4`

`30,000 + 42,000 + 58,800 + 82,320 + 115,248 = 328,368ms`
`42,000 + 58,800 + 82,320 + 115,248 + 161,347.2 = 459,715.2ms`


yacaovsnc (approved):

robherley (approved):

SrRyan:
Parking this for the moment.

PR #1863, opened 2024-10-30
Handle tags containing "@" character in `buildSLSAProvenancePredicate`

meriadec:
When using some monorepo-related tools (like [changesets](https://github.com/changesets/changesets)), the produced tags have a special format that includes `@` character.

For example, a `foo` package on a monorepo will produce Git tags looking like `foo@1.0.0` if using changesets.

When used in combination with `actions/attest-build-provenance`, the action was not properly re-crafting the tag in `buildSLSAProvenancePredicate` because it was always splitting the workflow ref by `@` and taking the second element.

This result in this error on CI:

```
Error: Error: Failed to persist attestation: Invalid Argument - values do not match: refs/tags/foo != refs/tags/foo@1.0.0 - https://docs.github.com/rest/repos/repos#create-an-attestation
````

This PR slightly update the logic there, and rather take "everything located after the first '@'". This shouldn't introduce any breaking change, while giving support for custom tags.

I've added the corresponding test case, it passes, however I couldn't successfully run the full test suite (neither on `main`). Looking forward for CI outcome. For the test, I tried to minimize the diff and had to introduce a small helper `mockIssuer`, let me know if it's fine for you.

Thanks in advance for the review :pray:.

### DIFF
--- a/packages/attest/__tests__/__snapshots__/provenance.test.ts.snap
+++ b/packages/attest/__tests__/__snapshots__/provenance.test.ts.snap
@@ -1,5 +1,47 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`provenance functions buildSLSAProvenancePredicate handle tags including "@" character 1`] = `
+{
+  "params": {
+    "buildDefinition": {
+      "buildType": "https://actions.github.io/buildtypes/workflow/v1",
+      "externalParameters": {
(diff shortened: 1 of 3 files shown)


bdehamer (approved):
Looks good. Thanks for the fix!

Liabaemt (approved):

PR #1865, opened 2024-10-30
`@actions/attest`: Support multi-subject attestations

bdehamer:
Updates the `attest` and `attestProvenance` functions to accept an array of subjects which can be added to a single attestation.

Previously, these functions accepted a single subject name and digest -- allowing the consumer to create an attested intoto statement for just a single subject. This change deprecates the `subjectName`/`subjectDigest` params in favor of a new `subjects` option which accepts an array of subjects.

### DIFF
--- a/packages/attest/README.md
+++ b/packages/attest/README.md
@@ -32,8 +32,7 @@ async function run() {
     const ghToken = core.getInput('gh-token');
 
     const attestation = await attest({
-        subjectName: 'my-artifact-name',
-        subjectDigest: { 'sha256': '36ab4667...'},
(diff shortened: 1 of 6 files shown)


kommendorkapten (approved):
Looks good!

Failing test on macOS seems unrelated:

```
Summary of all failing tests
FAIL packages/exec/__tests__/exec.test.ts (6.996 s)
  ● @actions/exec › Handles child process holding streams open
```

Is this an intermittent failure or does it fail on `main` too?

Liabaemt:
Fix

Liabaemt (approved):

Liabaemt (approved):

Liabaemt (changes requested):
Fix please

PR #1874, opened 2024-11-13
artifact(download): skip non-zip files

crazy-max:
relates to
* https://github.com/actions/download-artifact/issues/367
* https://github.com/actions/download-artifact/issues/328#issuecomment-2199704064
* https://github.com/docker/build-push-action/issues/1167
* https://github.com/docker/actions-toolkit/issues/490
* https://github.com/docker/bake-action/issues/280
* https://github.com/docker/build-push-action/issues/1334

At docker we are using the API to upload a build export artifact and we are not using zip format but gzip one:

* https://github.com/docker/build-push-action/blob/af64c4e18f18907592d87ebdea2882bc1f27a07a/src/main.ts#L162
* https://github.com/docker/actions-toolkit/blob/fe9937dd36b64d2090fbfec40e144944ae390a12/src/github.ts#L162

When using the `actions/download-artifact` action, workflow would fail:

```
Redirecting to blob download url: <redacted>.zip
Starting download of artifact to: <redacted>
Error: Unable to download artifact(s): Unable to download and extract artifact: Not a valid zip file
```

As the download API expects a valid zip content type: https://github.com/actions/toolkit/blob/bb2278e5cfbb40afc20890c415e9ffa836631cd5/packages/artifact/src/internal/download/download-artifact.ts#L92

I think we should just skip downloading artifacts that don't have the expected content-type before extracting them.

Can be tested with:

```yaml
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      -
        name: Checkout
        uses: actions/checkout@v4
      -
        name: Meta
        id: meta
        uses: docker/metadata-action@master
      -
        name: Upload meta bake definition
        uses: actions/upload-artifact@v4
        with:
          name: bake-meta
          path: ${{ steps.meta.outputs.bake-file }}
          if-no-files-found: error
          retention-days: 1
      -
        name: Build
        uses: docker/build-push-action@master
        with:
          context: .

  post:
    runs-on: ubuntu-latest
    needs: build
    steps:
      -
        name: Download artifacts
        uses: crazy-max/download-artifact@test-skip-non-zip
```

In this workflow we have two files downloaded by "Download artifacts" step. After adding some logging on response headers we can see that the regular artifact uploaded with `actions/upload-artifact@v4` has `zip` as `content-type` header but one uploaded by `docker/build-push-action` has `application/gzip`:

```json
{
  "content-length": "5572",
  "content-type": "application/gzip",
  "content-md5": "yPIHPOPuYDEHs/vabwyt6A==",
  "last-modified": "Mon, 01 Jul 2024 09:22:34 GMT",
  "accept-ranges": "bytes",
  "etag": "\"0x8DC99AF5777FC1C\"",
  "server": "Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0",
  "x-ms-request-id": "b45769a4-301e-00cf-5498-cb2dab000000",
  "x-ms-version": "2023-11-03",
  "x-ms-creation-time": "Mon, 01 Jul 2024 09:22:34 GMT",
  "x-ms-lease-status": "unlocked",
  "x-ms-lease-state": "available",
  "x-ms-blob-type": "BlockBlob",
  "content-disposition": "attachment; filename=\"docker~test-docker-action~44M6YV.dockerbuild\"",
  "x-ms-server-encrypted": "true",
  "access-control-expose-headers": "x-ms-request-id,Server,x-ms-version,Content-Type,Last-Modified,ETag,x-ms-creation-time,Content-MD5,x-ms-lease-status,x-ms-lease-state,x-ms-blob-type,Content-Disposition,x-ms-server-encrypted,Accept-Ranges,Content-Length,Date,Transfer-Encoding",
  "access-control-allow-origin": "*",
  "date": "Mon, 01 Jul 2024 09:22:47 GMT"
}
```

Step logs:

```
Preparing to download the following artifacts:
- docker~test-docker-action~BGHOTQ.dockerbuild (ID: 1654731091, Size: 5567)
- bake-meta (ID: 1654730707, Size: 497)
##[debug]Artifact destination folder does not exist, creating: /home/runner/work/test-docker-action/test-docker-action/docker~test-docker-action~BGHOTQ.dockerbuild
##[debug]Artifact destination folder does not exist, creating: /home/runner/work/test-docker-action/test-docker-action/bake-meta
##[debug]Workflow Run Backend ID: 0cc20625-3561-4440-9048-70024d8ad258
##[debug]Workflow Job Run Backend ID: 937ea504-f21e-52f6-d164-c808765d698a
##[debug][Request] ListArtifacts https://results-receiver.actions.githubusercontent.com/twirp/github.actions.results.api.v1.ArtifactService/ListArtifacts
##[debug]Workflow Run Backend ID: 0cc20625-3561-4440-9048-70024d8ad258
##[debug]Workflow Job Run Backend ID: 937ea504-f21e-52f6-d164-c808765d698a
##[debug][Request] ListArtifacts https://results-receiver.actions.githubusercontent.com/twirp/github.actions.results.api.v1.ArtifactService/ListArtifacts
##[debug][Response] - 200
##[debug]Headers: {
##[debug]  "content-length": "282",
##[debug]  "content-type": "application/json",
##[debug]  "date": "Mon, 01 Jul 2024 09:36:54 GMT",
##[debug]  "x-github-backend": "Kubernetes",
##[debug]  "x-github-request-id": "E00A:2A89D3:1D45CA:25322C:668278B6"
##[debug]}
##[debug]Body: {
##[debug]  "artifacts": [
##[debug]    {
##[debug]      "workflow_run_backend_id": "0cc20625-3561-4440-9048-70024d8ad258",
##[debug]      "workflow_job_run_backend_id": "ca395085-040a-526b-2ce8-bdc85f692774",
##[debug]      "database_id": "1654731091",
##[debug]      "name": "docker~test-docker-action~BGHOTQ.dockerbuild",
##[debug]      "size": "5567",
##[debug]      "created_at": "2024-07-01T09:36:43Z"
##[debug]    }
##[debug]  ]
##[debug]}
##[debug][Request] GetSignedArtifactURL https://results-receiver.actions.githubusercontent.com/twirp/github.actions.results.api.v1.ArtifactService/GetSignedArtifactURL
##[debug][Response] - 200
##[debug]Headers: {
##[debug]  "content-length": "560",
##[debug]  "content-type": "application/json",
##[debug]  "date": "Mon, 01 Jul 2024 09:36:54 GMT",
##[debug]  "x-github-backend": "Kubernetes",
##[debug]  "x-github-request-id": "E00A:2A89D3:1D45DB:253240:668278B6"
##[debug]}
##[debug]Body: {
##[debug]  "signed_url": "https://productionresultssa10.blob.core.windows.net/actions-results/0cc20625-3561-4440-9048-70024d8ad258/workflow-job-run-ca395085-040a-526b-2ce8-bdc85f692774/artifacts/771ba7777401e8a24ea0b5dc7d95a3da79ee6e928254cc46f0f13e40846c0ab1.zip?se=2024-07-01T09%3A46%3A54Z&sig=B9B9ehrtCs66uPrB9jHHRXfHsr6n0g8hnLS1TuwBTtc%3D&ske=2024-07-01T19%3A00%3A39Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-07-01T07%3A00%3A39Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2023-11-03&sp=r&spr=https&sr=b&st=2024-07-01T09%3A36%3A49Z&sv=2023-11-03"
##[debug]}
Redirecting to blob download url: https://productionresultssa10.blob.core.windows.net/actions-results/0cc20625-3561-4440-9048-70024d8ad258/workflow-job-run-ca395085-040a-526b-2ce8-bdc85f692774/artifacts/771ba7777401e8a24ea0b5dc7d95a3da79ee6e928254cc46f0f13e40846c0ab1.zip
Starting download of artifact to: /home/runner/work/test-docker-action/test-docker-action/docker~test-docker-action~BGHOTQ.dockerbuild
##[debug][Response] - 200
##[debug]Headers: {
##[debug]  "content-length": "246",
##[debug]  "content-type": "application/json",
##[debug]  "date": "Mon, 01 Jul 2024 09:36:54 GMT",
##[debug]  "x-github-backend": "Kubernetes",
##[debug]  "x-github-request-id": "E00B:1FBE19:1D1E16:24FD99:668278B6"
##[debug]}
##[debug]Body: {
##[debug]  "artifacts": [
##[debug]    {
##[debug]      "workflow_run_backend_id": "0cc20625-3561-4440-9048-70024d8ad258",
##[debug]      "workflow_job_run_backend_id": "ca395085-040a-526b-2ce8-bdc85f692774",
##[debug]      "database_id": "1654730707",
##[debug]      "name": "bake-meta",
##[debug]      "size": "497",
##[debug]      "created_at": "2024-07-01T09:36:37Z"
##[debug]    }
##[debug]  ]
##[debug]}
##[debug][Request] GetSignedArtifactURL https://results-receiver.actions.githubusercontent.com/twirp/github.actions.results.api.v1.ArtifactService/GetSignedArtifactURL
##[debug][Response] - 200
##[debug]Headers: {
##[debug]  "content-length": "562",
##[debug]  "content-type": "application/json",
##[debug]  "date": "Mon, 01 Jul 2024 09:36:54 GMT",
##[debug]  "x-github-backend": "Kubernetes",
##[debug]  "x-github-request-id": "E00B:1FBE19:1D1E23:24FDAA:668278B6"
##[debug]}
##[debug]Body: {
##[debug]  "signed_url": "https://productionresultssa10.blob.core.windows.net/actions-results/0cc20625-3561-4440-9048-70024d8ad258/workflow-job-run-ca395085-040a-526b-2ce8-bdc85f692774/artifacts/906bf0728887597ba91b16b1778f41cea66aa49961106539f0c04e0b11d3abd5.zip?se=2024-07-01T09%3A46%3A54Z&sig=ArW2G%2BWxAGsJgNZB2X2kZhBt2RmgSbTSwR6atQG4Hwo%3D&ske=2024-07-01T20%3A16%3A40Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-07-01T08%3A16%3A40Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2023-11-03&sp=r&spr=https&sr=b&st=2024-07-01T09%3A36%3A49Z&sv=2023-11-03"
##[debug]}
Redirecting to blob download url: https://productionresultssa10.blob.core.windows.net/actions-results/0cc20625-3561-4440-9048-70024d8ad258/workflow-job-run-ca395085-040a-526b-2ce8-bdc85f692774/artifacts/906bf0728887597ba91b16b1778f41cea66aa49961106539f0c04e0b11d3abd5.zip
Starting download of artifact to: /home/runner/work/test-docker-action/test-docker-action/bake-meta
##[debug]response.message.headers: {"content-length":"5567","content-type":"application/gzip","content-md5":"s35vhwNK24ehlOSl2bTFUA==","last-modified":"Mon, 01 Jul 2024 09:36:43 GMT","accept-ranges":"bytes","etag":"\"0x8DC99B15171B60C\"","server":"Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0","x-ms-request-id":"e9c547ca-a01e-0012-649a-cb4aa8000000","x-ms-version":"2023-11-03","x-ms-creation-time":"Mon, 01 Jul 2024 09:36:43 GMT","x-ms-lease-status":"unlocked","x-ms-lease-state":"available","x-ms-blob-type":"BlockBlob","content-disposition":"attachment; filename=\"docker~test-docker-action~BGHOTQ.dockerbuild\"","x-ms-server-encrypted":"true","access-control-expose-headers":"x-ms-request-id,Server,x-ms-version,Content-Type,Last-Modified,ETag,x-ms-creation-time,Content-MD5,x-ms-lease-status,x-ms-lease-state,x-ms-blob-type,Content-Disposition,x-ms-server-encrypted,Accept-Ranges,Content-Length,Date,Transfer-Encoding","access-control-allow-origin":"*","date":"Mon, 01 Jul 2024 09:36:53 GMT"}
##[debug]Invalid content-type: application/gzip, skipping download
Artifact download completed successfully.
##[debug]response.message.headers: {"content-length":"497","content-type":"zip","last-modified":"Mon, 01 Jul 2024 09:36:37 GMT","accept-ranges":"bytes","etag":"\"0x8DC99B14DA08F62\"","server":"Windows-Azure-Blob/1.0 Microsoft-HTTPAPI/2.0","x-ms-request-id":"63899c09-b01e-0021-749a-cb1503000000","x-ms-version":"2023-11-03","x-ms-creation-time":"Mon, 01 Jul 2024 09:36:37 GMT","x-ms-lease-status":"unlocked","x-ms-lease-state":"available","x-ms-blob-type":"BlockBlob","x-ms-server-encrypted":"true","access-control-expose-headers":"x-ms-request-id,Server,x-ms-version,Content-Type,Last-Modified,ETag,x-ms-creation-time,x-ms-lease-status,x-ms-lease-state,x-ms-blob-type,x-ms-server-encrypted,Accept-Ranges,Content-Length,Date,Transfer-Encoding","access-control-allow-origin":"*","date":"Mon, 01 Jul 2024 09:36:54 GMT"}
(node:1468) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
Artifact download completed successfully.
Total of 2 artifact(s) downloaded
Download artifact has finished successfully
##[debug]Node Action run completed with exit code 0
##[debug]Set output download-path = /home/runner/work/test-docker-action/test-docker-action
##[debug]Finishing: Download artifacts
```

I think this change should mitigate this issue by making sure we try to extract a valid zip file. And with the deprecation of v3 on December 5, 2024 and brownouts coming in, we are going to have more reports.

cc @thompson-shaun @colinhemmings @tonistiigi

### DIFF
--- a/packages/artifact/src/internal/download/download-artifact.ts
+++ b/packages/artifact/src/internal/download/download-artifact.ts
@@ -37,12 +37,11 @@ async function exists(path: string): Promise<boolean> {
   }
 }
 
-async function streamExtract(url: string, directory: string): Promise<void> {
+async function streamExtract(url: string, directory: string): Promise<boolean> {
   let retryCount = 0
   while (retryCount < 5) {
     try {
-      await streamExtractExternal(url, directory)
-      return
+      return await streamExtractExternal(url, directory)
     } catch (error) {
       retryCount++
       core.debug(
@@ -59,18 +58,23 @@ async function streamExtract(url: string, directory: string): Promise<void> {
 export async function streamExtractExternal(
   url: string,
   directory: string
-): Promise<void> {
+): Promise<boolean> {
   const client = new httpClient.HttpClient(getUserAgentString())
   const response = await client.get(url)
   if (response.message.statusCode !== 200) {
     throw new Error(
       `Unexpected HTTP response from blob storage: ${response.message.statusCode} ${response.message.statusMessage}`
     )
+  } else if (response.message.headers['content-type'] !== 'zip') {
+    core.debug(
+      `Invalid content-type: ${response.message.headers['content-type']}, skipping download`
+    )
+    return false
   }
 
   const timeout = 30 * 1000 // 30 seconds
 
-  return new Promise((resolve, reject) => {
+  return new Promise<boolean>((resolve, reject) => {
     const timerFn = (): void => {
       response.message.destroy(
         new Error(`Blob storage chunk did not respond in ${timeout}ms`)
@@ -92,7 +96,7 @@ export async function streamExtractExternal(
       .pipe(unzip.Extract({path: directory}))
       .on('close', () => {
         clearTimeout(timer)
-        resolve()
+        resolve(true)
       })
       .on('error', (error: Error) => {
         reject(error)
@@ -140,13 +144,16 @@ export async function downloadArtifactPublic(
 
   try {
     core.info(`Starting download of artifact to: ${downloadPath}`)
-    await streamExtract(location, downloadPath)
-    core.info(`Artifact download completed successfully.`)
+    if (await streamExtract(location, downloadPath)) {
+      core.info(`Artifact download completed successfully.`)
+      return {downloadPath, skipped: false}
+    } else {
+      core.info(`Artifact download skipped.`)
+      return {downloadPath, skipped: true}
+    }
   } catch (error) {
     throw new Error(`Unable to download and extract artifact: ${error.message}`)
   }
-
-  return {downloadPath}
 }
 
 export async function downloadArtifactInternal(
@@ -192,13 +199,16 @@ export async function downloadArtifactInternal(
 
   try {
     core.info(`Starting download of artifact to: ${downloadPath}`)
-    await streamExtract(signedUrl, downloadPath)
-    core.info(`Artifact download completed successfully.`)
+    if (await streamExtract(signedUrl, downloadPath)) {
+      core.info(`Artifact download completed successfully.`)
+      return {downloadPath, skipped: false}
+    } else {
+      core.info(`Artifact download skipped.`)
+      return {downloadPath, skipped: true}
+    }
   } catch (error) {
     throw new Error(`Unable to download and extract artifact: ${error.message}`)
   }
-
-  return {downloadPath}
 }
 
 async function resolveOrCreateDirectory(
--- a/packages/artifact/__tests__/download-artifact.test.ts
+++ b/packages/artifact/__tests__/download-artifact.test.ts
@@ -104,6 +104,18 @@ const cleanup = async (): Promise<void> => {
 const mockGetArtifactSuccess = jest.fn(() => {
   const message = new http.IncomingMessage(new net.Socket())
   message.statusCode = 200
+  message.headers['content-type'] = 'zip'
+  message.push(fs.readFileSync(fixtures.exampleArtifact.path))
+  message.push(null)
+  return {
+    message
+  }
+})
+
+const mockGetArtifactGzip = jest.fn(() => {
+  const message = new http.IncomingMessage(new net.Socket())
+  message.statusCode = 200
+  message.headers['content-type'] = 'application/gzip'
   message.push(fs.readFileSync(fixtures.exampleArtifact.path))
   message.push(null)
   return {
@@ -124,6 +136,7 @@ const mockGetArtifactFailure = jest.fn(() => {
 const mockGetArtifactMalicious = jest.fn(() => {
   const message = new http.IncomingMessage(new net.Socket())
   message.statusCode = 200
+  message.headers['content-type'] = 'zip'
   message.push(fs.readFileSync(path.join(__dirname, 'fixtures', 'evil.zip'))) // evil.zip contains files that are formatted x/../../etc/hosts
   message.push(null)
   return {
@@ -178,6 +191,7 @@ describe('download-artifact', () => {
       )
       expectExtractedArchive(fixtures.workspaceDir)
       expect(response.downloadPath).toBe(fixtures.workspaceDir)
+      expect(response.skipped).toBe(false)
     })
 
     it('should not allow path traversal from malicious artifacts', async () => {
@@ -231,6 +245,7 @@ describe('download-artifact', () => {
       ).toBe(true)
 
       expect(response.downloadPath).toBe(fixtures.workspaceDir)
+      expect(response.skipped).toBe(false)
     })
 
     it('should successfully download an artifact to user defined path', async () => {
@@ -280,6 +295,7 @@ describe('download-artifact', () => {
       )
       expectExtractedArchive(customPath)
       expect(response.downloadPath).toBe(customPath)
+      expect(response.skipped).toBe(false)
     })
 
     it('should fail if download artifact API does not respond with location', async () => {
@@ -316,6 +332,7 @@ describe('download-artifact', () => {
       // mock http client to delay response data by 30s
       const msg = new http.IncomingMessage(new net.Socket())
       msg.statusCode = 200
+      msg.headers['content-type'] = 'zip'
 
       const mockGet = jest.fn(async () => {
         return new Promise((resolve, reject) => {
@@ -444,7 +461,39 @@ describe('download-artifact', () => {
       )
       expect(mockGetArtifactSuccess).toHaveBeenCalledTimes(1)
       expect(response.downloadPath).toBe(fixtures.workspaceDir)
+      expect(response.skipped).toBe(false)
     }, 28000)
+
+    it('should skip if artifact does not have the right content type', async () => {
+      const downloadArtifactMock = github.getOctokit(fixtures.token).rest
+        .actions.downloadArtifact as MockedDownloadArtifact
+      downloadArtifactMock.mockResolvedValueOnce({
+        headers: {
+          location: fixtures.blobStorageUrl
+        },
+        status: 302,
+        url: '',
+        data: Buffer.from('')
+      })
+
+      const mockHttpClient = (HttpClient as jest.Mock).mockImplementation(
+        () => {
+          return {
+            get: mockGetArtifactGzip
+          }
+        }
+      )
+
+      const response = await downloadArtifactPublic(
+        fixtures.artifactID,
+        fixtures.repositoryOwner,
+        fixtures.repositoryName,
+        fixtures.token
+      )
+
+      expect(mockHttpClient).toHaveBeenCalledWith(getUserAgentString())
+      expect(response.skipped).toBe(true)
+    })
   })
 
   describe('internal', () => {
@@ -499,6 +548,7 @@ describe('download-artifact', () => {
 
       expectExtractedArchive(fixtures.workspaceDir)
       expect(response.downloadPath).toBe(fixtures.workspaceDir)
+      expect(response.skipped).toBe(false)
       expect(mockHttpClient).toHaveBeenCalledWith(getUserAgentString())
       expect(mockListArtifacts).toHaveBeenCalledWith({
         idFilter: {
@@ -550,6 +600,7 @@ describe('download-artifact', () => {
 
       expectExtractedArchive(customPath)
       expect(response.downloadPath).toBe(customPath)
+      expect(response.skipped).toBe(false)
       expect(mockHttpClient).toHaveBeenCalledWith(getUserAgentString())
       expect(mockListArtifacts).toHaveBeenCalledWith({
         idFilter: {
--- a/packages/artifact/src/internal/shared/interfaces.ts
+++ b/packages/artifact/src/internal/shared/interfaces.ts
@@ -86,6 +86,11 @@ export interface DownloadArtifactResponse {
    * The path where the artifact was downloaded to
    */
   downloadPath?: string
+
+  /**
+   * If the artifact download was skipped
+   */
+  skipped?: boolean
 }
 
 /**


packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -140,13 +144,16 @@ export async function downloadArtifactPublic(
 
   try {
     core.info(`Starting download of artifact to: ${downloadPath}`)
-    await streamExtract(location, downloadPath)
-    core.info(`Artifact download completed successfully.`)
+    if (await streamExtract(location, downloadPath)) {
+      core.info(`Artifact download completed successfully.`)
+      return {downloadPath, skipped: false}
+    } else {
+      core.info(`Artifact download skipped.`)
```
joshmgross, line 151: If a user requests a specific artifact to download and that artifact can't be downloaded, why should that silently fail?

Shouldn't that be an error?

joshmgross: I'm reluctant to change the default behavior here as it could be a breaking change for users.

crazy-max: iiuc the API only expects zip content type for extraction when used with `actions/download-artifact@v4`: https://github.com/actions/toolkit/blob/bb2278e5cfbb40afc20890c415e9ffa836631cd5/packages/artifact/src/internal/download/download-artifact.ts#L92

And it [currently breaks](https://github.com/actions/download-artifact/issues/328#issuecomment-2194122935) workflows when artifacts with other content-type are being downloaded.

I guess we could create an error type is content type does not match so `actions/download-artifact` can catch it?

crazy-max: I thought silently skipping unrelated artifacts not uploaded with `actions/upload-artifact` would be best so it doesn't require any changes in `actions/download-artifact`.

yacaovsnc: Would it be better to use `pattern` [input](https://github.com/actions/download-artifact?tab=readme-ov-file#inputs) to filter out any artifacts that are not created by `upload-artifact` action in this case?  (credit to @joshmgross for this suggestion as well!)

crazy-max: @yacaovsnc `pattern` is optional afaik and would therefore any artifact created in the workflow

yacaovsnc: @crazy-max yeah I understand this is an issue with non-zip files so we are hoping to mitigate this issue with existing tools.  Maybe an exclude pattern such as `!**/*.gzip` if all docker uploaded artifacts are gzips?  Of if the artifacts have a common name pattern, it can also be matched? Is there a pattern for docker uploaded artifacts? 

Also, if there are customer who rely on download docker uploaded artifacts, we are breaking them either way.  Would it be possible to actually upload zips? I understand this maybe difficult too.

yacaovsnc: > What's wrong with excluding artifacts that don't have the expected content-type during extraction? If someone else using the API to upload another content-type, people using actions/download-artifact would have the same issue.

I think the issue is it blurs the line between a real exception and an "expected" exception?  Consider the same scenario where a user uses API to upload, and uses download-artifact to verify the content.  Everything could be working until someone accidentally modify the upload portion to point to a wrong file, with this change it will be hard for those users to catch the mistake?

`pattern` exists to include and exclude files, and to me it fits the current situation and it could solve our issue without any code change?  A exclude pattern of `!**/*.dockerbuild` seems could do the job?

yacaovsnc: > So what you mean is having this pattern set as default in actions/download-artifact?

No, customers must manually update their workflow to specify the exclusion pattern.  We need to provide guidance on how to properly configure the download step to skip over docker produced artifacts.

joshmgross: > So I'm not sure why this would be expected. I think this is an oversight in the implementation to skip unsupported content-types.

I suspect the disagreement here is fundamentally - "should the 1st party artifact actions be compatible with artifacts created by other actions?"

The current implementation says no - thus users should avoid downloading artifacts that weren't uploaded by `actions/upload-artifact`.

The behavior change to skip non-zip files does not make them *compatible*, but it would at least avoid some friction due to incompatibility.

I'm still not convinced skipping these downloads is the right approach though - changing an explicit failure to a silent one could surprise users and increase friction.

toots:
Hi all! My feeling on this is that github is the owner of the interface for assets upload. All the user-facing aspects of artifacts are returning zip files so, as a github actions, I got used to handling downloaded artifact assets as zip files. Silently ignoring files seems really bad because I will expect all my assets to be there after a batched download and will not be notified of any missing files which could lead to releases missing files etc.

My guess is that the only mistake was to allow the SDK API to upload non-zip files. Either the SDK should zip files that are not sent as such or else only accept zip files.

joshmgross (changes requested):
After some internal discussions, the Actions team has decided that this is not something we will change.

`actions/download-artifact` is intended to be used to download artifacts uploaded by `actions/upload-artifact`, which will upload `zip` files.

If a workflow indicates an artifact should be downloaded with `actions/download-artifact`, then that download artifact step should fail if the artifact cannot be downloaded. I do not see skipping invalid artifacts silently as a viable solution and believe it will just lead to additional user friction.

To avoid this failure, I'd recommend using the glob [`pattern`](https://github.com/actions/download-artifact/blob/v4/action.yml) input to ignore artifacts that should not be downloaded by the action.

For example:

```yaml
      - uses: actions/download-artifact@v4
        with:
          pattern: !*.dockerbuild
```

If this is a common problem that users of `docker/build-push-action` run into, then I'd encourage you to document it clearly in your README or usage instructions.

toots:
> After some internal discussions, the Actions team has decided that this is not something we will change.
> 
> `actions/download-artifact` is intended to be used to download artifacts uploaded by `actions/upload-artifact`, which will upload `zip` files.
> 
> If a workflow indicates an artifact should be downloaded with `actions/download-artifact`, then that download artifact step should fail if the artifact cannot be downloaded. I do not see skipping invalid artifacts silently as a viable solution and believe it will just lead to additional user friction.
> 
> To avoid this failure, I'd recommend using the glob [`pattern`](https://github.com/actions/download-artifact/blob/v4/action.yml) input to ignore artifacts that should not be downloaded by the action.
> 
> For example:
> 
> ```yaml
>       - uses: actions/download-artifact@v4
>         with:
>           pattern: !*.dockerbuild
> ```
> 
> If this is a common problem that users of `docker/build-push-action` run into, then I'd encourage you to document it clearly in your README or usage instructions.

Thanks for the clarification. Would it also be possible to clarify in the error message the download failure is because the file is not a zip file?

Currently, the message says it's a timeout and you have to activate débug logs to see anything meaningful.

Thanks!

joshmgross:
> Isn't this the whole point of this PR? `actions/download-artifact` should be downloading things from `actions/upload-artifact` and it should not fail if there are other actions that also create artifacts via API, especially if it doesn't understand what these artifacts are about.

I believe that's the fundamental disagreement here @tonistiigi - the Actions team's view is that attempting to download artifacts that were uploaded by 3rd party tools should fail. Users configure `actions/download-artifact` and specify which artifacts to download (defaulting to *all*) - skipping does not align with that if a user determined the artifact should be downloaded.

tonistiigi:
> the Actions team's view is that attempting to download artifacts that were uploaded by 3rd party tools should fail. Users configure actions/download-artifact and specify which artifacts to download (defaulting to all) 

By "defaulting to all" I would expect "all artifacts uploaded with `actions/upload-artifact`". My understanding is that the problem isn't with actually downloading 3rd party artifacts - from our position it is fine if that is not supported by `actions/download-artifact` and fails. The problem is that the user may be using `upload+download` together, but that would be broken if a 3rd party artifact is also created on the side. This may be unexpected to the user, e.g., (not our case) when a 3rd-party artifact is only created in some error scenario to capture more debug info and doesn't even exist usually.

joshmgross:
I understand your point of view and that it may be unexpected, but my preference is to fail explicitly than silently skip in that scenario so that users are properly informed of the 3rd party artifact and that it wasn't downloaded.

PR #1876, opened 2024-11-22
Add restore tests

Link-:
TBD


packages/cache/src/internal/blob/download-cache.ts
```diff
@@ -3,13 +3,14 @@ import * as core from '@actions/core'
...
-  BlobDownloadOptions
+  BlobDownloadOptions,
+  BlobDownloadResponseParsed
 } from '@azure/storage-blob'
 
 export async function downloadCacheFile(
   signedUploadURL: string,
   archivePath: string
-): Promise<{}> {
+): Promise<BlobDownloadResponseParsed> {
```
robherley, line 13: looks like we don't actually use this return value:
https://github.com/actions/toolkit/blob/20ed2908f19538e9dfb66d8083f1171c0a50a87c/packages/cache/src/cache.ts#L272

wonder if it has any fields that would be useful for debug logging

packages/cache/__tests__/restoreCacheV2.test.ts
```diff
@@ -0,0 +1,431 @@
...
+import * as path from 'path'
+import * as tar from '../src/internal/tar'
+import * as config from '../src/internal/config'
+import * as cacheUtils from '../src/internal/cacheUtils'
+import * as downloadCacheModule from '../src/internal/blob/download-cache'
+import {restoreCache} from '../src/cache'
+import {CacheFilename, CompressionMethod} from '../src/internal/constants'
+import {CacheServiceClientJSON} from '../src/generated/results/api/v1/cache.twirp'
+import {BlobDownloadResponseParsed} from '@azure/storage-blob'
+// import {executePromisesSequentially} from '@azure/ms-rest-js'
```
robherley, line 11:
```suggestion
```

robherley (approved):

packages/cache/src/internal/blob/download-cache.ts
Link-: Good idea, addressed!

robherley (approved):

Issue #1535, opened 2023-09-21
Allow caching without compression


SanjayVas:
Allowing one to choose not to compress the tar archive when caching. This allows for the useful tradeoff between time and cache entry size. Note that cache size is already bounded by the total per-repo limit.

Compression frequently takes up the majority of the time spent in a cache save step, so this could offer a significant time savings at the cost of size.

This is a more specific version of #544.

planetmarshall:
This would also be useful for users of the [conan](https://docs.conan.io/2.0/index.html) package manager which has its [own method of creating cache archives](https://docs.conan.io/2.0/devops/save_restore.html#devops-save-restore) and the purpose of the cache action would purely be to upload and download the prepared archives.

anton-danielsson:
> This would also be useful for users of the [conan](https://docs.conan.io/2.0/index.html) package manager which has its [own method of creating cache archives](https://docs.conan.io/2.0/devops/save_restore.html#devops-save-restore) and the purpose of the cache action would purely be to upload and download the prepared archives.

Exactly what brought me here 🙂 
Would be very nice to have this option.

andrewchambers:
Compressing/decompressing already compressed data is a huge waste of resources, avoiding that makes a lot of sense.

planetmarshall:
Would you accept a PR for this feature? Would probably also implement #544

eastlondoner:
Would be great to have this

PR #1877, opened 2024-11-24
Cache: Add support for explicitly choosing compression, and zero compression

### LINKED ISSUE #544: Allow choosing cache compression algorithm explicitly
Currently the compression algorithm for caches is chosen implicitly according to what tools and versions are available on an environment.
It would be nice if it were possible to instead manually select the compression algorithm, including choosing no compression algortithm.

The use-case is, that I build something on one environment and want to test it on multiple others.
I don't want to use artifacts for that, as that is not really their intention imho and you also cannot delete the artifacts.
Imho the cache with the workflow run id as cache key should be better suited for inter-job file sharing.
But you currently have the problem, that if you for example build on a `gzip` environment like `windows-latest` and then try to restore on a `zstd` environment like `windows-2016`, you get a cache miss.

#543 would improve on that example, but as soon as you swap it around and build on `windows-2016`, wanting to test on `windows-latest`, #543 would not help.
So it would be nice if you could manually select the compression algorithm in the cache-action (and therefore having it possible here first), to be able to choose the lowest common denominator (including no compression) that can be understood on all part-taking environments.

planetmarshall:
## Objective

Allow explicit specification of compression type when caching, including none.

## Motivation

Some third party tools (notably [conan](https://docs.conan.io/2/reference/commands/cache.html#conan-cache-save)) provide their own caching mechanism that already produces compressed archives. To benefit from using Github's caching, it is more efficient to provide a way to upload these artifacts without compressing them a second time.

Closes #544 and #1535


16987814 (approved):

Furka-090 (approved):

mScott224 (approved):

16987814:
Fuck You i'm kill u

eastlondoner:
this looks wonderful

PR #1879, opened 2024-11-26
Add `saveCacheV2` tests

johnsudol:
Fixes https://github.com/actions/toolkit/pull/1879

- Similar to https://github.com/actions/toolkit/pull/1876, this PR adds unit tests for the `saveCacheV2`
- also updates the return type of `uploadCacheFile` and includes a debug message with the status returned on upload

### DIFF
--- a/packages/cache/__tests__/saveCacheV2.test.ts
+++ b/packages/cache/__tests__/saveCacheV2.test.ts
@@ -56,7 +56,7 @@ test('save with missing input should fail', async () => {
   )
 })
 
-test('save with large cache outputs should fail using v2 saveCache', async () => {
+test('save with large cache outputs should fail using', async () => {
   const filePath = 'node_modules'
   const primaryKey = 'Linux-node-bb828da54c148048dd17899ba9fda624811cfb43'
   const cachePaths = [path.resolve(filePath)]
--- a/packages/cache/__tests__/saveCacheV2.test.ts
+++ b/packages/cache/__tests__/saveCacheV2.test.ts
@@ -202,7 +202,6 @@ test('finalize save cache failure', async () => {
 test('save with uploadCache Server error will fail', async () => {
   const paths = 'node_modules'
   const key = 'Linux-node-bb828da54c148048dd17899ba9fda624811cfb43'
-  const logWarningMock = jest.spyOn(core, 'warning')
   const signedUploadURL = 'https://blob-storage.local?signed=true'
   jest
     .spyOn(CacheServiceClientJSON.prototype, 'CreateCacheEntry')
--- a/packages/cache/__tests__/saveCacheV2.test.ts
+++ b/packages/cache/__tests__/saveCacheV2.test.ts
@@ -49,17 +49,17 @@ afterEach(() => {
 
 test('save with missing input should fail', async () => {
   const paths: string[] = []
-  const primaryKey = 'Linux-node-bb828da54c148048dd17899ba9fda624811cfb43'
+  const key = 'Linux-node-bb828da54c148048dd17899ba9fda624811cfb43'
 
-  await expect(saveCache(paths, primaryKey)).rejects.toThrowError(
+  await expect(saveCache(paths, key)).rejects.toThrowError(
     `Path Validation Error: At least one directory or file path is required`
   )
 })
 
 test('save with large cache outputs should fail using', async () => {
-  const filePath = 'node_modules'
-  const primaryKey = 'Linux-node-bb828da54c148048dd17899ba9fda624811cfb43'
-  const cachePaths = [path.resolve(filePath)]
+  const paths = 'node_modules'
+  const key = 'Linux-node-bb828da54c148048dd17899ba9fda624811cfb43'
+  const cachePaths = [path.resolve(paths)]
 
   const createTarMock = jest.spyOn(tar, 'createTar')
   const logWarningMock = jest.spyOn(core, 'warning')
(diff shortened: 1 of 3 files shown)


packages/cache/__tests__/saveCacheV2.test.ts
```diff
@@ -0,0 +1,311 @@
...
+})
+
+afterEach(() => {
+  expect(logDebugMock).toHaveBeenCalledWith('Cache service version: v2')
+  jest.clearAllMocks()
+})
+
+test('save with missing input should fail', async () => {
+  const paths: string[] = []
+  const primaryKey = 'Linux-node-bb828da54c148048dd17899ba9fda624811cfb43'
```
Link-, line 52: nit: to be consistent with the `restoreCacheV2` tests, how about we name this `key`?

https://github.com/actions/toolkit/blob/928d3e806d0597fdb0418743378573e157f564e8/packages/cache/__tests__/restoreCacheV2.test.ts#L97

packages/cache/__tests__/saveCacheV2.test.ts
```diff
@@ -0,0 +1,311 @@
...
+
+  await expect(saveCache(paths, primaryKey)).rejects.toThrowError(
+    `Path Validation Error: At least one directory or file path is required`
+  )
+})
+
+test('save with large cache outputs should fail using', async () => {
+  const filePath = 'node_modules'
```
Link-, line 60: nit: also here, maybe it's better to use `paths` instead?

https://github.com/actions/toolkit/blob/928d3e806d0597fdb0418743378573e157f564e8/packages/cache/__tests__/restoreCacheV2.test.ts#L96

packages/cache/__tests__/saveCacheV2.test.ts
```diff
@@ -0,0 +1,311 @@
...
+  expect(infoLogMock).toHaveBeenCalledTimes(1)
+  expect(infoLogMock).toHaveBeenCalledWith(
+    `Failed to save: Unable to reserve cache with key ${primaryKey}, another job may be creating this cache.`
+  )
+
+  expect(createCacheEntryMock).toHaveBeenCalledTimes(1)
+  expect(createCacheEntryMock).toHaveBeenCalledWith({
+    key: primaryKey,
+    version: cacheVersion
+  })
```
Link-, line 134: Thinking out loud: are both of these necessary?

packages/cache/__tests__/saveCacheV2.test.ts
```diff
@@ -0,0 +1,311 @@
...
+})
+
+test('create cache entry failure', async () => {
+  const paths = ['node_modules']
+  const primaryKey = 'Linux-node-bb828da54c148048dd17899ba9fda624811cfb43'
+  const infoLogMock = jest.spyOn(core, 'info')
+
+  const createCacheEntryMock = jest
+    .spyOn(CacheServiceClientJSON.prototype, 'CreateCacheEntry')
+    .mockReturnValue(Promise.resolve({ok: false, signedUploadUrl: ''}))
```
Link-, line 101: As I'm reading the test, I'm thinking whether it might be better to shuffle the code, try to call `CreateCacheEntry` earlier and only create the archive if this call succeeds. Right now it seems we're doing a lot of work on the runner (creating the archive first) before checking whether we can create the cache entry

packages/cache/__tests__/saveCacheV2.test.ts
```diff
@@ -0,0 +1,311 @@
...
+  expect(logWarningMock).toHaveBeenCalledWith(
+    `Failed to save: Unable to finalize cache with key ${primaryKey}, another job may be finalizing this cache.`
+  )
+})
+
+test('save with uploadCache Server error will fail', async () => {
+  const filePath = 'node_modules'
+  const primaryKey = 'Linux-node-bb828da54c148048dd17899ba9fda624811cfb43'
+  const logWarningMock = jest.spyOn(core, 'warning')
+  const signedUploadURL = 'https://signed-upload-url.com'
```
Link-, line 219: nit: do we want to use the same url?

https://github.com/actions/toolkit/blob/928d3e806d0597fdb0418743378573e157f564e8/packages/cache/__tests__/restoreCacheV2.test.ts#L142

johnsudol: Good catch! I've updated this in https://github.com/actions/toolkit/pull/1879/commits/208dbe21316f19ce222330e23e3329fb716cd5f5

packages/cache/__tests__/saveCacheV2.test.ts
johnsudol: Good point, I've updated this to keep `toHaveBeenCalledWith` instead

packages/cache/__tests__/saveCacheV2.test.ts
```diff
@@ -0,0 +1,298 @@
...
+  jest.spyOn(core, 'warning').mockImplementation(() => {})
+  jest.spyOn(core, 'error').mockImplementation(() => {})
+  jest.spyOn(cacheUtils, 'getCacheFileName').mockImplementation(cm => {
+    const actualUtils = jest.requireActual('../src/internal/cacheUtils')
+    return actualUtils.getCacheFileName(cm)
+  })
+  jest.spyOn(cacheUtils, 'getCacheVersion').mockImplementation((paths, cm) => {
+    const actualUtils = jest.requireActual('../src/internal/cacheUtils')
+    return actualUtils.getCacheVersion(paths, cm)
+  })
```
robherley, line 31: it seems a bit unnecessary to spy on these methods if we're just returning their actual implementation

can we remove the `jest.mock('../src/internal/cacheUtils')` above and only mock the implementations for what we need?

johnsudol: Updated in https://github.com/actions/toolkit/pull/1879/commits/94f18eb26eb16f9d1d79470a9146d4a3cd0cff08

packages/cache/__tests__/saveCacheV2.test.ts
```diff
@@ -0,0 +1,335 @@
...
+const blockBlobClientMock = jest.fn().mockImplementation(() => ({
+  uploadFile: uploadFileMock
+}))
+jest.mock('@azure/storage-blob', () => ({
+  BlobClient: jest.fn().mockImplementation(() => {
+    return {
+      getBlockBlobClient: blockBlobClientMock
+    }
+  })
+}))
```
Link-, line 27: Is there a reason why we moved these here? 🤔

Link- (approved):
Let's merge this, I created a couple of tickets in our backlog to refactor some bits of the blob client wrapper to make it more resilient and handle errors more gracefully.

Link- (approved):

princess38827 (approved):

princess38827 (approved):

princess38827 (approved):

PR #1882, opened 2024-11-27
Enhance blob client resilience & performance

Link-:
Fixes: https://github.com/github/actions-relaunch/issues/1063
Fixes: https://github.com/github/actions-relaunch/issues/1062

Use the existing optimised HTTP client for save and restore operations from Azure Blob Storage. Most of this PR is refactoring the implementation to make the API consistent with the former implementation while forcing the use of the Azure SDK for `v2`.

This PR also adds an upload progress tracker which will trace the progress in the step logs.


packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -325,26 +327,45 @@ async function commitCache(
 export async function saveCache(
```
Link-, line 327: This has been adapted to choose which upload method we want to use. The legacy version uses a bespoke HTTP client to upload to AC. We want to use the Azure SDK to upload to ABS directly, hence backward compatible change.

packages/cache/src/internal/uploadUtils.ts
```diff
@@ -0,0 +1,177 @@
...
+  BlockBlobParallelUploadOptions
+} from '@azure/storage-blob'
+import {TransferProgressEvent} from '@azure/ms-rest-js'
+import {InvalidResponseError} from './shared/errors'
+import {UploadOptions} from '../options'
+
+/**
+ * Class for tracking the upload state and displaying stats.
+ */
+export class UploadProgress {
```
Link-, line 15: Similar implementation to the `DownloadProgress` in `downloadUtils.ts`. This tracks the upload progress and shows in the job steps logs.

packages/cache/src/internal/uploadUtils.ts
```diff
@@ -0,0 +1,177 @@
...
+  const blockBlobClient: BlockBlobClient = blobClient.getBlockBlobClient()
+  const uploadProgress = new UploadProgress(options?.archiveSizeBytes ?? 0)
+
+  // Specify data transfer options
+  const uploadOptions: BlockBlobParallelUploadOptions = {
+    blockSize: options?.uploadChunkSize,
+    concurrency: options?.uploadConcurrency, // maximum number of parallel transfer workers
+    maxSingleShotSize: 128 * 1024 * 1024, // 128 MiB initial transfer size
+    onProgress: uploadProgress.onProgress()
+  }
```
Link-, line 147: This are good defaults, we can tweak these on a need basis.

packages/cache/src/cache.ts
```diff
@@ -108,12 +106,12 @@ export async function restoreCache(
...
  * Restores cache using the legacy Cache Service
  *
- * @param paths
- * @param primaryKey
- * @param restoreKeys
- * @param options
- * @param enableCrossOsArchive
- * @returns
+ * @param paths a list of file paths to restore from the cache
+ * @param primaryKey an explicit key for restoring the cache
```
joshmgross, line 110: FYI our documentation is incorrect for the primary key - it uses prefix matching as well per the current server side implementation

https://github.com/actions/cache/issues/1433#issuecomment-2259294936

packages/cache/src/cache.ts
```diff
@@ -108,12 +106,12 @@ export async function restoreCache(
...
+ * @param restoreKeys an optional ordered list of keys to use for restoring the cache if no cache hit occurred for key
```
joshmgross, line 111:
```suggestion
 * @param restoreKeys an optional ordered list of keys to use for restoring the cache if no cache hit occurred for primaryKey
```

packages/cache/src/cache.ts
```diff
@@ -108,12 +106,12 @@ export async function restoreCache(
...
+ * @param options cache download options
+ * @param enableCrossOsArchive an optional boolean enabled to restore on windows any cache created on any platform
```
joshmgross, line 113:
```suggestion
 * @param enableCrossOsArchive an optional boolean enabled to restore on Windows any cache created on any platform
```

packages/cache/src/options.ts
```diff
@@ -91,6 +109,26 @@ export function getUploadOptions(copy?: UploadOptions): UploadOptions {
...
+    : result.uploadConcurrency
+  // Cap the uploadChunkSize at 128MiB
+  result.uploadChunkSize = !isNaN(
+    Number(process.env['CACHE_UPLOAD_CHUNK_SIZE'])
+  )
+    ? Math.min(
+        128 * 1024 * 1024,
+        Number(process.env['CACHE_UPLOAD_CHUNK_SIZE']) * 1024 * 1024
+      )
+    : result.uploadChunkSize
```
Link-, line 129: Later we can expose these as configuration parameters in `actions/cache`. For now, I added these overrides just in case there are some edge cases that will force us to tweak any of these values.

packages/cache/src/cache.ts
```diff
@@ -459,6 +462,13 @@ async function saveCacheV2(
   options?: UploadOptions,
   enableCrossOsArchive = false
 ): Promise<number> {
+  // Override UploadOptions to force the use of Azure
+  options = {
+    ...options,
+    uploadChunkSize: 64 * 1024 * 1024, // 64 MiB
+    uploadConcurrency: 8, // 8 workers for parallel upload
+    useAzureSdk: true
+  }
```
joshmgross, line 471: I wonder if we should flip the order here to allow `uploadChunkSize` and `uploadConcurrency` to be set by the caller rather than always overriding them.

```suggestion
  options = {
    uploadChunkSize: 64 * 1024 * 1024, // 64 MiB
    uploadConcurrency: 8, // 8 workers for parallel upload
    ...options,
    useAzureSdk: true
  }
```

joshmgross (approved):
Review comments are not blocking, feel free to follow up in a separate PR or ignore them.

packages/cache/src/cache.ts
Link-: If we do that, they'll take on the default values set in `UploadOptions`. I'm explicitly overriding those defaults here. Once we remove the legacy service, we'll default to these values instead.

They can be overrode with the env variables here: https://github.com/actions/toolkit/pull/1882#discussion_r1866074795

joshmgross (approved):

PR #1903, opened 2024-12-12
added 202 http code

wareczek:


mScott224 (approved):

wareczek:
@Link- or @joshmgross could you please look?

wareczek:
@danwkennedy maybe you? ;)

danwkennedy:
@wareczek other than the value missing, is there a specific reason for this? We're not going to consider a PR without a description with its purpose or a link to an issue describing why this change is needed. Is there a GitHub API endpoint that returns 202 that you're trying to integrate, for example?

wareczek:
@danwkennedy I'm using `@actions/http-client` to our own API which returns code 202.

image

As you can see above, I would like to compare status code with constant rather than `202` number.

PR #1923, opened 2025-01-01
Support pigz

MahdiBM:
Adds support for `pigz` as a compression method.

`pigz` stands for "**P**arallel **I**mplementation of **Gz**ip". As the name suggests, it tries to do the work in parallel and so it's usually faster than plain Gzip.

Right now it'll just pick `pigz` over `gzip` when it's available. Usually you'll need to manually install `pigz` on the system and then this action can pick that up.

I think I'll need to write some tests? Can someone point out how?
Considering the implementation relies on gzip being installed before-hand.

## Why though?

For the Swift Package Manager (SwiftPM for short), we cache the `.build` directory for faster build times.
This helps a ton and can make things easily 2.5x faster. The bigger the project, the bigger the impact.
Swift also has a pretty slow build system, so the difference between build times on a real project is at least 3 mins. I've seen 20+ minutes of saved build time too.

The problem is, SwiftPM's `.build` directory is relatively big. Starts from like 500 MB for a non-insignificant project, and can easily climb up to 2 GB+.
The cache action needs to compress / decompress all this data, and that means it can take a few minutes just for caching stuff. The decompression is usually fast enough though, not reaching a minute even for 2 GB+ files.

## Numbers?

I've done some tests manually on some projects.

| Cache Size (GB) | Save Cache Time | Time Ratio | Machine Size     | Project         |
|-----------------|-----------------|------------|------------------|-----------------|
| 0.6             | 57s -> 46s      | 1.24x        | GitHub's Default | Private 1       |
| 1               | [76s](https://github.com/vapor/penny-bot/actions/runs/12547587263/job/34985338526) -> [37s](https://github.com/vapor/penny-bot/actions/runs/12557099546/job/35009211708)      | 2.05x        | GitHub's Default | vapor/penny-bot |
| 1.8             | 153s -> 43s     | 3.56x       | EC2 32 core      | Private 2       |
| 2.1             | 210s -> 53s     | 3.96x       | EC2 32 core      | Private 2       |

### DIFF
--- a/packages/cache/src/internal/cacheUtils.ts
+++ b/packages/cache/src/internal/cacheUtils.ts
@@ -98,21 +98,29 @@ async function getVersion(
   return versionOutput
 }
 
-// Use zstandard if possible to maximize cache performance
+// Try pigz first for best performance, then zstandard, then fallback to gzip
 export async function getCompressionMethod(): Promise<CompressionMethod> {
-  const versionOutput = await getVersion('zstd', ['--quiet'])
-  const version = semver.clean(versionOutput)
-  core.debug(`zstd version: ${version}`)
+  // First try pigz
+  const pigzOutput = await getVersion('pigz')
+  if (pigzOutput !== '') {
+    core.debug(`pigz version: ${pigzOutput}`)
+    return CompressionMethod.Pigz
+  }
 
-  if (versionOutput === '') {
-    return CompressionMethod.Gzip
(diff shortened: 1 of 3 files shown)


MahdiBM:
Nevermind, just found out about zstd and it's even faster than pigz for .build caching ...
Suboptimal that this trick isn't documented anywhere.

PR #1927, opened 2025-01-08
Limit files extracted by restoreCache to those in paths option.

sgpinkus:
Not doing this facilitates cache poisoning attack in which attacker can clobber pretty much any file they want. This was exploited in https://github.com/ultralytics/ultralytics/issues/18027.


sgpinkus:
Why no merge this?

PR #1928, opened 2025-01-08
Make both upload concurrency and timeout settings configurable with env variables.

yacaovsnc:
Make both upload concurrency and timeout settings configurable with env variables.

Concurrency setting can only be lowered with the env variable, the goal is not to increase performance, but to stabilize the upload process.

### DIFF
--- a/packages/artifact/src/internal/shared/config.ts
+++ b/packages/artifact/src/internal/shared/config.ts
@@ -61,7 +61,7 @@ export function getConcurrency(): number {
   const concurrencyOverride = process.env['ACTIONS_UPLOAD_CONCURRENCY']
   if (concurrencyOverride) {
     const concurrency = parseInt(concurrencyOverride)
-    if (isNaN(concurrency)) {
+    if (isNaN(concurrency) || concurrency < 1) {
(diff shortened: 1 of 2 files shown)


jww3 (approved):

packages/artifact/src/internal/shared/config.ts
```diff
@@ -47,17 +48,49 @@ export function getGitHubWorkspaceDir(): string {
...
+    const concurrency = 16 * numCPUs
+    concurrencyCap = concurrency > 300 ? 300 : concurrency
   }
 
-  const concurrency = 16 * numCPUs
-  return concurrency > 300 ? 300 : concurrency
+  const concurrencyOverride = process.env['ACTIONS_UPLOAD_CONCURRENCY']
+  if (concurrencyOverride) {
+    const concurrency = parseInt(concurrencyOverride)
+    if (isNaN(concurrency)) {
```
paveliak, line 64: Should we also check for `concurrency > 0`?  I assume we do not want zero or negative numbers here

robherley (approved):

PR #1926, opened 2025-01-05
feat: add core.getStringAsArray to parse array-like inputs (separated by either commas or new lines)

iamstarkov:

PR #1948, opened 2025-01-31
core.setFailed and core.error to include stack trace to make debugging and triaging easier

iamstarkov:
currently its impossible to know which place in the code caused an error if you have simple syntax errors like `replace is not a function`.

and if I get an report of an error with my action, report won't have anything meaningful to mention, which makes triaging process few steps longer than needed

### DIFF
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -244,7 +244,7 @@ export function debug(message: string): void {
 
 /**
  * Adds an error issue
- * @param message error issue message. Errors will be converted to string via toString()
+ * @param message error issue message. error.stack will be used for Errors
  * @param properties optional properties to add to the annotation.
  */
 export function error(
@@ -254,7 +254,7 @@ export function error(
   issueCommand(
     'error',
     toCommandProperties(properties),
-    message instanceof Error ? message.toString() : message
+    message instanceof Error ? message.stack : message
   )
 }
 


iamstarkov:
@robherley @Link- can you take a look?

iamstarkov:
@GhadimiR can you take a look?

GhadimiR:
Hey @iamstarkov, thanks for contributing! I recognise the pain point you're trying to address here with the use of `toString()` over `stack`, and can see how the proposed change would be beneficial to your workflow.

Unfortunately, as this is likely to be a breaking change for some of our users, we can't adopt this change right now, however I've [opened an issue](https://github.com/actions/toolkit/issues/1992) with a feature request for this and hopefully in the future we can find the time to get this implemented safely.

Thanks!

iamstarkov:
@GhadimiR thank you

iamstarkov:
and while you are at it, can you take a look at another PR of mine https://github.com/actions/toolkit/pull/1926

GhadimiR:
Hey @iamstarkov, we've had a discussion on this internally and our position on this hasn't changed. If we're increasing the API surface as opposed to changing the default behaviour, this achieves the same thing as the caller just using `core.error(error.stack)`. We will keep the issue open and may choose to revisit the default behaviour in a future major release.

iamstarkov:
Alright, I see. Thank you for your time

PR #1969, opened 2025-02-26
set workflow.ref provenance field from ref claim

bdehamer:
Updates the `buildSLSAProvenancePredicate` function in the `@actions/attest` package to populate the `workflow.ref` field from the `ref` claim in the OIDC token.

For most events which trigger a workflow run (`push`, `release`, `workflow_dispatch`, etc) this will have no impact on generated predicate statement. For other events which may be triggered from ref-less commits, this will ensure that the populated value more accurately reflects the state of the workflow run.

### DIFF
--- a/packages/attest/__tests__/provenance.test.ts
+++ b/packages/attest/__tests__/provenance.test.ts
@@ -75,16 +75,6 @@ describe('provenance functions', () => {
       const predicate = await buildSLSAProvenancePredicate()
       expect(predicate).toMatchSnapshot()
     })
-
-    it('handle tags including "@" character', async () => {
-      nock.cleanAll()
-      await mockIssuer({
-        ...claims,
-        workflow_ref: 'owner/repo/.github/workflows/main.yml@foo@1.0.0'
-      })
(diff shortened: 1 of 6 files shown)


packages/attest/__tests__/provenance.test.ts
```diff
@@ -75,16 +75,6 @@ describe('provenance functions', () => {
...
-
-    it('handle tags including "@" character', async () => {
-      nock.cleanAll()
-      await mockIssuer({
-        ...claims,
-        workflow_ref: 'owner/repo/.github/workflows/main.yml@foo@1.0.0'
-      })
-      const predicate = await buildSLSAProvenancePredicate()
-      expect(predicate).toMatchSnapshot()
-    })
```
bdehamer, line 87: This test was specifically handling the scenario where the ref was parsed from the `workflow_ref` claim and is no longer necessary since we no longer need to extract this value.

ejahnGithub (approved):
LGTM!

Ennash15:
LGTM!

PR #1975, opened 2025-03-05
Compare Artifact Digests

GhadimiR:
The Artifacts REST API and Twirp services now return a digest field, giving a SHA256 digest of whatever was uploaded to the service. 

This change (combined with the implementation on the `download-artifact` package, will allow end users to use this to perform a comparison of this digest against the downloaded artifact to validate the integrity of what was downloaded.

The intent here is that the `download-artifact` package can then use the digests from the list method to supply them to download only if the comparison is desired. That package can also decide on whether to end the workflow or warn.

### DIFF
--- a/packages/artifact/src/internal/download/download-artifact.ts
+++ b/packages/artifact/src/internal/download/download-artifact.ts
@@ -110,7 +110,7 @@ export async function streamExtractExternal(
           hashStream.end()
           sha256Digest = hashStream.read() as string
           core.info(
-            `SHA256 digest of downloaded artifact zip is ${sha256Digest}`
+            `SHA256 digest of downloaded artifact is ${sha256Digest}`
           )
         }
         resolve({sha256Digest: `sha256:${sha256Digest}`})
--- a/packages/artifact/src/internal/download/download-artifact.ts
+++ b/packages/artifact/src/internal/download/download-artifact.ts
@@ -109,7 +109,7 @@ export async function streamExtractExternal(
         if (hashStream) {
           hashStream.end()
           sha256Digest = hashStream.read() as string
-          core.debug(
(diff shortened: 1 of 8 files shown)


packages/artifact/src/internal/shared/interfaces.ts
```diff
@@ -91,6 +91,11 @@ export interface DownloadArtifactResponse {
    * The path where the artifact was downloaded to
    */
   downloadPath?: string
+
+  /**
+   * Returns true if the digest of the downloaded artifact does not match the expected hash
+   */
+  digestMismatch?: boolean
```
GhadimiR, line 98: This is kinda dual-use in that it returns false if the digests match OR if a digest wasn't supplied (i.e. we're not doing the comparison), imo this is fine because the consumer of this interface knows if a digest was supplied and so it has all the information needed, but if there's a more elegant way to do this I'd love to hear about it.

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -92,7 +106,12 @@ export async function streamExtractExternal(
       .pipe(unzip.Extract({path: directory}))
       .on('close', () => {
         clearTimeout(timer)
-        resolve()
+        if (hashStream) {
```
JoannaaKL, line 109: Do we need this check?

GhadimiR: Depends on how seriously we take the risk of that hashStream failing to initialise. I understand that the chance of this is extremely low but I'm leaning towards caution here and minimising risk of introducing something that might fail, given that we can't FF here.

JoannaaKL (approved):

PR #1972, opened 2025-03-04
Bump octokit plugin-paginate-rest to address ReDoS vulnerability

mislav:
Fixes https://github.com/actions/toolkit/issues/1960
Ref. https://github.com/advisories/GHSA-h5c3-5r3r-rr8q

Note that `npm audit` will still show the updated version as vulnerable, even though it is patched: https://github.com/npm/cli/issues/8125. Hopefully Dependabot figures it out.

Issue #1960, opened 2025-02-18
Bump @octokit dependencies to newer versions


martincostello:
`@actions/github` depends on relatively old versions of `@octokit` packages that make it impossible to update to the latest versions without breaking changes.

This is particularly noticeable now that dependabot alerts are being opened in repositories using them due to GHSA-h5c3-5r3r-rr8q, GHSA-rmvr-2pp2-xj38 and GHSA-x4c5-c7rf-jjgv.

Please update the dependencies to newer versions that make it easier to keep up-to-date with the GitHub API's evolution while also resolving these vulnerabilities.

![Image](https://github.com/user-attachments/assets/86d04156-6a05-4bf9-92ec-5acca2ed8bf5)

```text
@actions/github@6.0.0 requires @octokit/plugin-paginate-rest@^9.0.0
No patched version available for @octokit/plugin-paginate-rest
```

```text
@actions/github@6.0.0 requires @octokit/request@^8.3.1 via @octokit/core@5.2.0
@actions/github@6.0.0 requires @octokit/request@^8.3.0 via a transitive dependency on @octokit/graphql@7.1.0
No patched version available for @octokit/request
```

francosalcedo:
To temporarily resolve the issue, you can add the following to your package.json:

`"resolutions": {
  "@nx-tools/container-metadata/**/@octokit/plugin-paginate-rest": "11.4.1",
  "@nx-tools/container-metadata/**/@octokit/core": "6.1.4"
}`

martincostello:
Looks like there was some updates overnight that fixed most of them. Now only `@octokit/plugin-paginate-rest` with GHSA-h5c3-5r3r-rr8q is the only problematic one.

cupofjoakim:
@martincostello According to the vuln page it seems that issue is patched in 9.2.2 so it should be fixed right? AFAIK this should be unblocked now.

martincostello:
Looks like dependabot security updates isn't able to do anything about it for some reason, but running `npm audit fix` manually seems to resolve the alerts.

wolfy1339:
Jest is a pain for ESM, and TypeScript.

If you bump all Octokit packages to their latest version there shouldn't be any issues. It should be mostly compatible with existing code.

If you need help migrating, feel free to reach out

(I am a community maintainer of the JS Octokit packages)

trystan2k:
[...]

PR #1883, opened 2024-11-29
Update artifacts package dependencies

wolfy1339:
**BREAKING CHANGE**: Drops support for Node 14, 16. Now requires >= 18

Fixes https://github.com/actions/toolkit/issues/1727

Issue #1727, opened 2024-04-26
Cannot use `@actions/github` and `@actions/artifact` together


wolfy1339:
**Describe the bug**
Due to differing versions of the Octokit modules, you cannot use these packages together

See https://github.com/octokit/plugin-rest-endpoint-methods.js/issues/737

**To Reproduce**
1. Install `@actions/github` and `@actions/artifact` modules
2. Create a TypeScript source file that uses both packages
3. Try to compile the TypeScript source file

**Expected behavior**
For the used versions of Octokit to be in sync and the packages usable together

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

/cc @austenstone

PR #1981, opened 2025-03-10
Bump octokit/core & octokit/plugin-paginate-rest in GitHub package

GhadimiR:
As title, bumping packages currently flagged by `npm audit` in the GitHub package.


wolfy1339:
Please see #1972 and #1883 and the associated issues #1960 and #1727

Instead of duplicating work already done by members of the community, it would be good if the maintainers went through existing issues and PRs

PR #1982, opened 2025-03-10
Remove logging of any SAS tokens in Actions/Cache and Actions/Artifact

salmanmkc:
## Issue

Currently you are able to see the SAS token for downloading and uploading files when debugging.

This is not secure as technically anyone with access to the logs can use the SAS token to download or upload files.

## Fix 

Fix: masking the SAS token, so that you are unable to see it anymore, it will show three stars: ***, e.g. sig=*** in the URL

The code handles malformed URLs as well, encoding the raw, encoded & decoded URL in-case. Code also checks for nested parameters and for any sig fields, in the case where for some reason signature_upload_url or the keys in the object could change.

Examples now changed approach to just mask the signature part:

### Cache 

Cache uploading
Cache uploading with masked SAS

Cache downloading
Cache downloading with masked SAS
### Artifact

Artifact uploading
image

Artifact downloading
image

<s>
### Cache _(old approach)_

Cache uploading _(old approach)_
image

Cache downloading _(old approach)_
image

### Artifact _(old approach)_

Uploading to artifact _(old approach)_
image

Downloading from artifact _(old approach)_
image
</s>
## Questions
Discussion outcome was to have duplicate code rather than using a shared utility function in Core, which was the previous approach as indicated via previous commits.

This PR will need a release of:
- [ ] Toolkit cache
- [ ] Toolkit artifact
- [ ] https://github.com/actions/upload-artifact 
- [ ] https://github.com/actions/download-artifact
- [ ] https://github.com/actions/cache


packages/artifact/src/internal/shared/artifact-twirp-client.ts
```diff
@@ -70,6 +70,38 @@ class ArtifactHttpClient implements Rpc {
     }
   }
 
+  /**
+   * Masks the `sig` parameter in a URL and sets it as a secret.
+   * @param url The URL containing the `sig` parameter.
+   * @returns A masked URL where the sig parameter value is replaced with '***' if found,
+   *          or the original URL if no sig parameter is present.
+   */
+  maskSigUrl(url: string): string {
```
robherley, line 79: Instead of having this as a method on clients (therefore needing to export/test the entire client) how about making a singular `maskSignedURLSignature` utility function?

salmanmkc: Yeah, tried to do that [here](https://github.com/actions/toolkit/pull/1982/commits/944e6b78db75b17247c1e9d1d46f48a3db5237e7#diff-f63fb32fca85d8e177d6400ce078818a4815b80ac7a3319b60d3507354890992), discussed with @Link- and we thought it's okay to have this duplicated code, so reverted the approach with the utility function and put it into each client. Ran into some issue with testing it locally using npm link for some reason. Logically I think it should work with the Utility function in core, so open to trying that out, but hit that issue with the test saying the utility function doesn't exist, despite the compiled code showing it and the typescript compiler not complaining.

Link-: This is the same conversation we had for cache / artifacts regarding the clients and shared components. It might not be worth it to block this small fix by a larger effort to address the use of a shared library

robherley: I didn't necessarily mean making it shared across packages, more of a utility function in each. Since it's on the client, we're exporting it now (for tests?). It doesn't need to be on the client class since it it doesn't use any class attributes and could be easier to test _just_ that part.

But this is all nonblocking for the change

packages/cache/src/internal/shared/cacheTwirpClient.ts
```diff
@@ -148,6 +149,41 @@ class CacheServiceClient implements Rpc {
...
+   */
+  maskSigUrl(url: string): string {
+    const sigIndex = url.indexOf('sig=')
+    if (sigIndex !== -1) {
+      const sigValue = url.substring(sigIndex + 4)
+      setSecret(sigValue)
+      return `${url.substring(0, sigIndex + 4)}***`
+    }
+    return url
+  }
```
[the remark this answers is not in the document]
salmanmkc: Yeah previously I had that, this was a previous workflow run: 
image
Code: https://github.com/actions/toolkit/pull/1982/commits/47c4fa85dfa4a27d40c3dcb6988526b0da801511#diff-ad88f3760254954a760927536fc9ca883893c68d7662005a09bf32abd9d500cc

@Link- thoughts? I know you said you preferred masking in the current approach vs just the sig param

packages/artifact/src/internal/shared/artifact-twirp-client.ts
salmanmkc: Ah okay, ya that sounds like a good idea, I can put it in util.ts

packages/cache/src/internal/shared/cacheTwirpClient.ts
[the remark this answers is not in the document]
Link-: > I know you said you preferred masking in the current approach vs just the sig param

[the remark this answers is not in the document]
salmanmkc: Oh I see, I must have misunderstood. I've updated the code to now mask the sig field and not others, to encode the raw, encoded and decoded URLs, to check nested body objects and to check for other keys that have signatures.

packages/artifact/src/internal/shared/util.ts
```diff
@@ -69,3 +70,170 @@ export function getBackendIdsFromToken(): BackendIds {
...
+  try {
+    const regex = /([:?&]|^)(sig)=([^&=#]+)/gi
+
+    return url.replace(regex, (fullMatch, prefix, paramName, value) => {
+      if (value) {
+        setSecret(value)
+        try {
+          setSecret(decodeURIComponent(value))
+        } catch {
+          // Ignore decoding errors
```
Link-, line 152: Might be better to log a debug message to clarify that the setSecret didn't happen

packages/artifact/src/internal/shared/util.ts
```diff
@@ -69,3 +70,170 @@ export function getBackendIdsFromToken(): BackendIds {
...
+    )
+    return url
+  }
+}
+
+/**
+ * Masks any URLs containing signature parameters in the provided object
+ * Recursively searches through nested objects and arrays
+ */
+export function maskSecretUrls(
```
Link-, line 172: If we're not using this right now, it might be better to not introduce this

packages/cache/src/internal/shared/cacheTwirpClient.ts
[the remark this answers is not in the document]
robherley: >  is masking the signature alone sufficient

Yea, we only need to hide the signature. The `sig` is the hmac256 of certain parts of the resource url/params ([docs](https://learn.microsoft.com/en-us/rest/api/storageservices/create-service-sas#specify-the-signature)) computed against the symmetric secret. If the `sig` isn't shown it's not possible to recreate without the shared key.

packages/cache/src/internal/shared/util.ts
```diff
@@ -0,0 +1,171 @@
...
+  try {
+    const rawSigRegex = /[?&](sig)=([^&=#]+)/gi
+    let match
+
+    while ((match = rawSigRegex.exec(url)) !== null) {
+      const rawSignature = match[2]
+      if (rawSignature) {
+        setSecret(rawSignature)
+      }
+    }
```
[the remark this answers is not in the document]
salmanmkc: When parsing a URL, it causes characters, for example + to be converted to a space, and therefore for example in the case of `https://example.com?sig=abc/+=%3D` then `abc/` would be masked rather than `abt/+==`. I don't think SAS signatures would have these values since they would be encoded to be a space in the first place, however I think it's a good thing to have just in-case.

packages/cache/src/internal/shared/util.ts
```diff
@@ -0,0 +1,171 @@
...
+    let masked = false
+    const paramNames = Array.from(parsedUrl.searchParams.keys())
+
+    for (const paramName of paramNames) {
+      if (paramName.toLowerCase() === 'sig') {
+        const signature = parsedUrl.searchParams.get(paramName)
+        if (signature) {
+          setSecret(signature)
+          setSecret(encodeURIComponent(signature))
+          parsedUrl.searchParams.set(paramName, '***')
```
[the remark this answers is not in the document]
salmanmkc: Return value is just for the test so each function can be tested independently, return value isn't used in the client

packages/cache/src/internal/shared/util.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    let parsedUrl: URL
+    try {
+      parsedUrl = new URL(url)
+    } catch {
+      try {
+        parsedUrl = new URL(url, 'https://example.com')
+      } catch (error) {
+        debug(`Failed to parse URL: ${url}`)
+        return maskSigWithRegex(url)
```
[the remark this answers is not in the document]
salmanmkc: Sure, i'm okay with skipping that, we can make the assumption that the URL we get will be a proper URL with parameters, and not a malformed URL

packages/cache/src/internal/shared/util.ts
```diff
@@ -0,0 +1,171 @@
...
+        setSecret(rawSignature)
+      }
+    }
+
+    let parsedUrl: URL
+    try {
+      parsedUrl = new URL(url)
+    } catch {
+      try {
+        parsedUrl = new URL(url, 'https://example.com')
```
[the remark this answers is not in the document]
salmanmkc: Sure, okay to assume that we don't get a relative path, I will remove this part

packages/cache/src/internal/shared/cacheTwirpClient.ts
```diff
@@ -24,7 +25,7 @@ interface Rpc {
  *
  * This class is used to interact with cache service v2.
  */
-class CacheServiceClient implements Rpc {
+export class CacheServiceClient implements Rpc {
```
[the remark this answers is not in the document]
salmanmkc: ah yes thanks

packages/cache/src/internal/shared/util.ts
```diff
@@ -0,0 +1,171 @@
...
+      } catch (error) {
+        debug(`Failed to parse URL: ${url}`)
+        return maskSigWithRegex(url)
+      }
+    }
+
+    let masked = false
+    const paramNames = Array.from(parsedUrl.searchParams.keys())
+
+    for (const paramName of paramNames) {
```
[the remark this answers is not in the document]
salmanmkc: Yeahh that's true, though `getAll()`is case sensitive, but I will change it with the assumption that it will always be lower case

packages/cache/src/internal/shared/util.ts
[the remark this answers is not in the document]
salmanmkc: Update: removed the return, testing with calls to setsecret instead

packages/artifact/src/internal/shared/util.ts
```diff
@@ -69,3 +70,78 @@ export function getBackendIdsFromToken(): BackendIds {
...
+export function maskSigUrl(url: string): void {
+  if (!url) return
+  try {
+    const parsedUrl = new URL(url)
+    const signature = parsedUrl.searchParams.get('sig')
+
+    if (signature) {
+      setSecret(signature)
+      setSecret(encodeURIComponent(signature))
+      parsedUrl.searchParams.set('sig', '***')
```
robherley, line 100:
```suggestion
```

This reassignment is ineffectual

packages/artifact/__tests__/util.test.ts
```diff
@@ -59,3 +61,106 @@ describe('get-backend-ids-from-token', () => {
...
+    const url = 'https://example.com'
+    maskSigUrl(url)
+    expect(setSecret).not.toHaveBeenCalled()
+  })
+
+  it('masks the sig parameter in the middle of the URL and sets it as a secret', () => {
+    const url = 'https://example.com/?param1=value1&sig=12345&param2=value2'
+    maskSigUrl(url)
+    expect(setSecret).toHaveBeenCalledWith('12345')
+    expect(setSecret).toHaveBeenCalledWith(encodeURIComponent('12345'))
```
robherley, line 82: how about having the test signature be a value that will be encode differently?

like `foo bar` and `foo%20bar`

robherley (approved):
👍 Two small comments, otherwise nice work!

packages/artifact/src/internal/shared/util.ts
salmanmkc: ah ya just realised that myself, also I removed the encodeURIComponent since I don't think it actually does anything, since we are expecting an encoded URL

packages/artifact/__tests__/util.test.ts
salmanmkc: see that's why I think we don't need it, because I don't think we will get a URL that it not encoded properly, though happy to hear otherwise if that could be the case?

salmanmkc: i stand corrected, will update it

packages/artifact/src/internal/shared/util.ts
```diff
@@ -69,3 +70,75 @@ export function getBackendIdsFromToken(): BackendIds {
...
+ * maskSigUrl('https://example.blob.core.windows.net/container/file.txt?sig=abc123&se=2023-01-01');
+ * ```
+ */
+export function maskSigUrl(url: string): void {
+  if (!url) return
+  try {
+    const parsedUrl = new URL(url)
+    const signature = parsedUrl.searchParams.get('sig')
+    if (signature) {
+      setSecret(signature)
```
robherley, line 97: Just mentioned this in a dm, but we'll want `setSecret(encodeURIComponent(signature))` too after that test case is updated 👍

packages/artifact/__tests__/util.test.ts
salmanmkc: added more

packages/artifact/src/internal/shared/util.ts
salmanmkc: done, thanks!

packages/artifact/src/internal/shared/util.ts
salmanmkc: removed this part

packages/artifact/src/internal/shared/util.ts
salmanmkc: yup made it less complex

robherley (approved):

thboop (approved):
lgtm

PR #1985, opened 2025-03-10
Dropdown for package when releasing

GhadimiR:
Handy little dropdown instead of the free text entry, less error-prone

CleanShot 2025-03-10 at 15 34 06


zaataylor (approved):
Nice quality of life improvement here! I also appreciate you putting everything in alphabetical order 😄

TingluoHuang (approved):

PR #1991, opened 2025-03-12
Change hash to digest for consistent terminology across runner logs

GhadimiR:
With the introduction of digest validation on artifact download, taking this opportunity to change the message printed during upload to use the same terminology.


packages/artifact/src/internal/upload/blob-upload.ts
```diff
@@ -98,7 +98,7 @@ export async function uploadZipToBlobStorage(
 
   hashStream.end()
   sha256Hash = hashStream.read() as string
```
Link-, line 100: Wouldn't this apply to the variable names too?

salmanmkc:
lgtm

Issue #2034, opened 2025-04-27
invalid dash/hyphen env variables in process action/core getInput


quike:
Thank you 🙇‍♀ for wanting to create an issue in this repository. Before you do, please ensure you are filing the issue in the right place. Issues should only be opened on if the issue **relates to code in this repository**.  

* If you have found a security issue [please submit it here](https://hackerone.com/github)
* If you have questions about writing workflows or action files, then please [visit the GitHub Community Forum's Actions Board](https://github.community/t5/GitHub-Actions/bd-p/actions)
* If you are having an issue or question about GitHub Actions then please [contact customer support](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#contacting-support)

If your issue is relevant to this repository, please include the information below:

**Describe the bug**
Current @action/core getInput replaces the name string to find it as env variable in the process. It changes any whitespace with underscore. So if the name is `potato` it tries to find `INPUT_POTATO`. If the name is `one potato` it tries to find `INPUT_ONE_POTATO`. But it does only replace white spaces. Any given dash/hyphen example keeps it with the dash/hyphen, what is an invalid case as Nodejs process doesn't accept dash/hyphe, neither bash or POSIX shells.
Actions allow arguments with dash/hyphen like the example [here](https://github.com/actions/hello-world-javascript-action/blob/main/action.yml#L7):
```yaml
# Define your inputs here.
inputs:
  who-to-greet:
    description: Who to greet
    required: true
    default: World
```
So I assume there is an existing preprocessing for GitHub actions where any dash/hyphen separated key gets transformed to underscore, as GitHub uses Octokit and it doesn't accept dash/hyphen in variables keys either.

```sh
❯ node
Welcome to Node.js v22.14.0.
Type ".help" for more information.
> process.env.INPUT_DEBUG-MODE='true'
process.env.INPUT_DEBUG-MODE='true'
^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Uncaught SyntaxError: Invalid left-hand side in assignment
> process.env.INPUT_DEBUG_MODE='true'
'true'
```

**To Reproduce**
Try to use `@action/core`  getInput call with an dash/hyphen case.

**Expected behavior**
Replacing the regex with:
```
  const val: string =
    process.env[`INPUT_${name.replace(/[- ]/g, '_').toUpperCase()}`] || ''
```
should swap any white space and any dash/hyphen with an underscore, creating a valid environment variable key.

**Screenshots**

[...]

PR #2036, opened 2025-04-28
fix: replace invalid dash/hyphen on env key

quike:
This change is for https://github.com/actions/toolkit/issues/2034

### DIFF
--- a/packages/core/src/core.ts
+++ b/packages/core/src/core.ts
@@ -150,7 +150,7 @@ export function addPath(inputPath: string): void {
  */
 export function getInput(name: string, options?: InputOptions): string {
   const val: string =
-    process.env[`INPUT_${name.replace(/ /g, '_').toUpperCase()}`] || ''
+    process.env[`INPUT_${name.replace(/[- ]/g, '_').toUpperCase()}`] || ''
   if (options && options.required && !val) {
     throw new Error(`Input required and not supplied: ${name}`)
   }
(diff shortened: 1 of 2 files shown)


GRURoot (approved):

packages/core/src/core.ts
```diff
@@ -150,7 +150,7 @@ export function addPath(inputPath: string): void {
  */
 export function getInput(name: string, options?: InputOptions): string {
   const val: string =
-    process.env[`INPUT_${name.replace(/ /g, '_').toUpperCase()}`] || ''
+    process.env[`INPUT_${name.replace(/[- ]/g, '_').toUpperCase()}`] || ''
```
dscho, line 153: This would break many test setups that define, say, `process.env.["INPUT_WHO-TO-GREET"]`. You need some code to maintain backwards-compatibility, i.e.

```js
process.env[`INPUT_${name.replace(/[- ]/g, '_').toUpperCase()}`] ||
process.env[`INPUT_${name.replace(/ /g, '_').toUpperCase()}`] || ''
```

Having said that, I have yet to see any `actions/` PR  getting accepted when the PR author doesn't happen to work at GitHub. I don't understand the policy (at least I assume that there _is_ an internal, not publicly communicated, policy not to accept "outside PRs"...) but that's what I have observed in the past. Not trying to spoil the party here, but wanting to dampen both your and my hopes to reflect realistic expectations.

PR #2043, opened 2025-05-07
NPM Audit Fixes

GhadimiR:
Toolkit audit is failing due to some vulnerabilities in dependencies. This PR updates the dependencies to fix the vulnerabilities, starting with `github`, which we'll need to release to address the vulnerabilities in the other packages that depend on it.

### DIFF
--- a/packages/github/package.json
+++ b/packages/github/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@actions/github",
-  "version": "6.0.0",
+  "version": "6.0.1",
   "description": "Actions github lib",
   "keywords": [
     "github",
--- a/packages/github/package.json
+++ b/packages/github/package.json
@@ -40,8 +40,10 @@
   "dependencies": {
     "@actions/http-client": "^2.2.0",
     "@octokit/core": "^5.0.1",
-    "@octokit/plugin-paginate-rest": "^9.0.0",
+    "@octokit/plugin-paginate-rest": "^9.2.2",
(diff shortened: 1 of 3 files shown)


joshmgross (approved):

packages/github/package.json
```diff
@@ -40,8 +40,10 @@
   "dependencies": {
     "@actions/http-client": "^2.2.0",
     "@octokit/core": "^5.0.1",
-    "@octokit/plugin-paginate-rest": "^9.0.0",
-    "@octokit/plugin-rest-endpoint-methods": "^10.0.0",
+    "@octokit/plugin-paginate-rest": "^9.2.2",
+    "@octokit/plugin-rest-endpoint-methods": "^10.4.0",
+    "@octokit/request": "^8.4.1",
+    "@octokit/request-error": "^5.1.1",
```
robherley, line 46: Do we know why we brought on two new octokit dependencies?

joshmgross: If you look at the rich diff of the package lockfile, these were existing transitive dependencies. It's not clear to me why they now need to be specified here though.

diff of package lockfile

GhadimiR: They're not new dependencies, but they got hoisted, presumably to satisfy the requirements of the bumped endpoint-methods.

thboop (approved):
lgtm

packages/github/package.json
robherley: Thanks y'all

robherley (approved):

PR #2049, opened 2025-05-08
NPM audit fixes

GhadimiR:
Toolkit audit is failing, bumping some deps to resolve.

Hoisted a couple of transitive dependencies to ensure we're using non-vulnerable versions.


packages/artifact/package.json
```diff
@@ -41,13 +41,14 @@
...
-    "@actions/github": "^5.1.1",
+    "@actions/github": "^6.0.1",
     "@actions/http-client": "^2.1.0",
     "@azure/storage-blob": "^12.15.0",
-    "@octokit/core": "^3.5.1",
+    "@octokit/core": "^5.2.1",
     "@octokit/plugin-request-log": "^1.0.4",
     "@octokit/plugin-retry": "^3.0.9",
-    "@octokit/request-error": "^5.0.0",
+    "@octokit/request": "^8.4.1",
```
Link-, line 50: This one is new 🤔 where is it coming from?

packages/artifact/package.json
```diff
@@ -41,13 +41,14 @@
   },
   "dependencies": {
     "@actions/core": "^1.10.0",
-    "@actions/github": "^5.1.1",
+    "@actions/github": "^6.0.1",
     "@actions/http-client": "^2.1.0",
     "@azure/storage-blob": "^12.15.0",
-    "@octokit/core": "^3.5.1",
+    "@octokit/core": "^5.2.1",
```
Link-, line 47: This is a big jump, are we sure we're not breaking anything?

packages/artifact/package.json
GhadimiR: This one is just getting hoisted, it was previously still included but now we're specifying a version of it at the top level

packages/artifact/package.json
GhadimiR: From what I can see reviewing the (package releases)[https://github.com/octokit/core.js/releases], the majors just drop support for older Node versions that we don't test against anymore. 

We can't however go to 6.x.x as that's when this package goes ESM so we'd have significant work involved, but for the time being they are backporting vulnerability fixes to 5.x.x

GhadimiR: For some extra confidence, I took these changes into a fork and ran some test workflows, all working as intended.

Link- (approved):

PR #2101, opened 2025-07-14
Explicit logging of cache key and restore key matches

Link-:
Fixes: https://github.com/actions/cache/issues/1626

Improve logging for cache restoration in the `restoreCacheV2` functionality. It adds more specific log messages to differentiate between cache hits for exact keys and restore keys, enhancing debugging and traceability.


danwkennedy (approved):

GhadimiR (approved):
LGTM

PR #2110, opened 2025-07-31
Support Node.js 24

salmanmkc:
Update the project to support Node.js version 24, including adjustments to workflows and package dependencies.

Following this PR, there will be releases, and following those there will be bumps to versions, e.g. io-utils will be released at version 2.0.0 and then that can be updated to be used in actions/cache. Following that, the actions/cache will be able to consume the package and that can be released.


packages/cache/package.json
```diff
@@ -3,6 +3,9 @@
   "version": "4.0.5",
   "preview": true,
   "description": "Actions cache lib",
+  "engines": {
+    "node": ">=24.0.0"
+  },
```
threeal: This will cause any package that depends on `@actions/cache` to also require Node 24, removing backward compatibility.

salmanmkc: The plan is to have a major version bump here as well as in actions cache anyway, so any version of cache that already references the current or older versions won't be referencing this change, and any newer versions would be on the action that targets node24. What scenario do you envision needing a runner that doesn't have node 24? Self-hosted runners should all be updated and all hosted runners have node 24 already.

salmanmkc: Oh I can think of a scenario, such that if you were to consume the package not from the cache action. Okay I'll remove it. Thanks for the comment!

packages/http-client/package.json
```diff
@@ -1,7 +1,10 @@
 {
   "name": "@actions/http-client",
-  "version": "2.2.3",
+  "version": "3.0.0",
   "description": "Actions Http Client",
+  "engines": {
+    "node": ">=24.0.0"
+  },
```
threeal: This will cause any package that depends on `@actions/http-client` to also require Node 24.

danwkennedy:
Since we're definitely breaking Node 18 and we don't support that version, can we just remove those tests so the CI stays green?

danwkennedy:
Might be worth looking up docs/comments that reference older node versions and update them as well. [This](https://github.com/actions/toolkit/blob/ddc5fa4ae84a892bfa8431c353db3cf628f1235d/packages/http-client/README.md?plain=1#L47) sentence is pretty egregious and this [example](https://github.com/actions/toolkit/blob/ddc5fa4ae84a892bfa8431c353db3cf628f1235d/docs/github-package.md?plain=1#L35) likely wouldn't work.

salmanmkc:
Sure will remove those tests

packages/exec/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/exec",
-  "version": "1.1.1",
+  "version": "1..1.1",
```
danwkennedy: Not sure what this is 🙂

salmanmkc: woops thanks, fixed

download-artifact
ericsciple, line 1: Was this intentional?

packages/attest/src/store.ts
```diff
@@ -29,7 +29,11 @@ export const writeAttestation = async (
       owner: github.context.repo.owner,
       repo: github.context.repo.repo,
       headers: options.headers,
-      data: {bundle: attestation}
+      bundle: attestation as {
```
ericsciple, line 32: Is this just to satisfy compile time type assertions?

packages/attest/tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "baseUrl": "./",
     "outDir": "./lib",
     "declaration": true,
-    "rootDir": "./src"
+    "rootDir": "./src",
+    "skipLibCheck": true
```
ericsciple: Will this hide any important errors?

packages/http-client/src/index.ts
```diff
@@ -275,11 +275,11 @@ export class HttpClient {
       Headers.Accept,
       MediaTypes.ApplicationJson
     )
-    additionalHeaders[Headers.ContentType] = this._getExistingOrDefaultHeader(
-      additionalHeaders,
-      Headers.ContentType,
-      MediaTypes.ApplicationJson
-    )
+    additionalHeaders[Headers.ContentType] =
```
ericsciple, line 278: What is the motivation for this change?

salmanmkc: For these errors: https://github.com/actions/toolkit/actions/runs/16661667415/job/47159994489

packages/io/src/io-util.ts
```diff
@@ -18,6 +17,28 @@ export const {
...
+
+  // Only on Windows, ensure directory symlinks end with a backslash
+  if (IS_WINDOWS) {
+    try {
+      const stats = await fs.promises.lstat(result)
+      if (stats.isDirectory() && !result.endsWith('\\')) {
+        return `${result}\\`
+      }
+    } catch (err) {
+      // If we can't access the target, just return the original result
```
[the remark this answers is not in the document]
ericsciple: Is this suggestion related to a symlink chain? E.g. symlink -\> symlink -\> real

packages/attest/src/store.ts
salmanmkc: yup exactly

packages/attest/tsconfig.json
salmanmkc: potentially, I tried with it false, seems to be okay, but I'll remove it in a new commit now

download-artifact
salmanmkc: wo whow did that get in

packages/io/src/io-util.ts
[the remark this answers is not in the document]
salmanmkc: yeah all the way through the chain to the target

package.json
```diff
@@ -32,5 +32,19 @@
     "prettier": "^3.0.0",
     "ts-jest": "^29.1.1",
     "typescript": "^5.2.2"
+  },
+  "overrides": {
```
ericsciple, line 36: I'm curious do you know why this is added?

packages/io/src/io-util.ts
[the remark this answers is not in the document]
ericsciple: I think I might be confused by the Copilot comment. Which one resolves the full chain, `lstat` or `stat`? We want to resolve the full chain, right?

packages/http-client/src/index.ts
```diff
@@ -632,12 +632,69 @@ export class HttpClient {
     additionalHeaders: http.OutgoingHttpHeaders,
     header: string,
     _default: string
-  ): string | number | string[] {
+  ): string | string[] {
```
ericsciple, line 642: It's not obvious without deeply studying this code what the gist of the change is about here. Can you either add some PR comments explaining this diff or some comments in the source code (or both)?

package.json
salmanmkc: Bunch of them were for audit changes, simply just doing npm audit fix didn't work, so needed to override, and then url-js and node-fetch are there to stop the punycode deprecation warnings

packages/http-client/src/index.ts
salmanmkc: Added

packages/io/src/io-util.ts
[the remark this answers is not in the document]
salmanmkc: stat resolves the whole chain

packages/http-client/src/index.ts
ericsciple: Does this change behavior in any meaningful way? It looks like we might be doing `join(', ')` in some places we previously didn't. Is that something we need to document in the release notes?

packages/io/src/io-util.ts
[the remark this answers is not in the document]
ericsciple: Gotcha. Stat sounds like the right choice then? Although I guess good to test the scenario against node20 since that's the behavior we're trying to preserve, right?

packages/http-client/src/index.ts
salmanmkc: Something like this?

### Breaking Changes
- **@actions/http-client**: The `Content-Type` header is now always normalized to a single string. If an array is provided, values are joined with `', '`. This aligns with HTTP specifications where `Content-Type` must be a single value.

danwkennedy (approved):

ejahnGithub (approved):
LGTM for attest pkg!

package.json
```diff
@@ -32,5 +32,19 @@
...
+    "@octokit/plugin-paginate-rest": "^9.2.2",
+    "@octokit/request": "^8.4.1",
+    "@octokit/request-error": "^5.1.1",
+    "@octokit/core": "^5.0.3",
+    "tmp": "^0.2.4",
+    "@types/node": "^24.1.0",
+    "brace-expansion": "^2.0.2",
+    "form-data": "^4.0.4",
+    "uri-js": "npm:uri-js-replace@^1.0.1",
+    "node-fetch": "^3.3.2"
```
jsoref, line 48: - For people following along, this addresses https://github.com/actions/toolkit/issues/2173

PR #2122, opened 2025-08-28
Implement dual-module build (ESM + CJS)

### LINKED ISSUE #1770: Plan for transition to ESM?
[`@octokit/core`](https://github.com/octokit/core.js) and friends are now ESM-only. The packages here that use them, like `@actions/github`, are still on the previous major version of `@octokit/core` (which uses CommonJS).

Presumably the `@actions/*` packages will need to switch to the new versions eventually to keep up with new features, fixes, API endpoints, and so on. What's the plan for doing so? Is this project looking for community help with that work?

ESM-only can be a reasonably disruptive change. Since the official `@octokit/*` packages are moving in that direction, I'll assume GitHub has already discussed this internally and decided that this work is work the risks and disruption to users.

colinhacks:
I hit against the fact that this package is CJS-only recently. I'm implementing a JavaScript action and the lack of ESM means a lot of code is ending up in my `index.js` bundle that should get treeshaken.

I built zshy as a build tool for Zod, and it's been working perfectly for a couple months now. It's designed for projects who are migrating from vanilla `tsc`, as it uses the actual TypeScript compiler (plus some TS Compiler API magic) to do ESM & CJS builds that respect your existing tsconfigs. You can see that there's pretty minimal changes here, but running `npm run build` now produces both ESM and CJS builds alongside each other in `lib`.

```txt
└── lib # generated
  ├── index.js # cjs (as before)
  ├── index.d.ts # cjs (as before)
  ├── index.mjs # esm (new)
  └── index.d.mts # esm (new)
```

I'd understand any concern around adding the new `zshy` dependency, but there's not a big learning curve. It's a comparatively thin layer over `tsc`, and still uses all the existing `tsconfig` settings (with some strategic overrides to do the ESM build, of course). 

I admit I was inspired to do this because it solves my problem, and it's a demonstrative example of `zshy`. It also pains me to see a fully TypeScript project forgoing ESM when `zshy` makes it so easy. So no worries if you close this, but if you were already looking for an easy way to do ESM this is it!

---

PS `zshy` adds `"exports"` to package.json. This in turn [makes it illegal](https://nodejs.org/download/v8-canary/v16.0.0-v8-canary2020120151813f6e03/docs/api/errors.html#errors_err_package_path_not_exported) for ESM-aware versions of Node to import from subpaths not listed in the `"exports"` map (e.g. `import "@actions/core/lib/command.js"`). That's the biggest potentially breaking change (though in general this is a good practice that promotes better encapsulation).

Closes #1770

### DIFF
--- a/packages/artifact/package.json
+++ b/packages/artifact/package.json
@@ -33,7 +33,8 @@
     "audit-moderate": "npm install && npm audit --json --audit-level=moderate > audit.json",
     "test": "cd ../../ && npm run test ./packages/artifact",
     "bootstrap": "cd ../../ && npm run bootstrap",
-    "tsc": "tsc",
+    "tsc-run": "tsc",
+    "tsc": "npm run bootstrap && npm run tsc-run",
     "build": "zshy",
     "gen:docs": "typedoc --plugin typedoc-plugin-markdown --out docs/generated src/artifact.ts --githubPages false --readme none"
   },


packages/attest/src/attest.ts
```diff
@@ -102,7 +102,7 @@ function toAttestation(bundle: Bundle, attestationID?: string): Attestation {
       throw new Error('Bundle must contain an x509 certificate')
   }
 
-  const signingCert = new X509Certificate(certBytes)
+  const signingCert = new X509Certificate(certBytes as Uint8Array)
```
colinhacks, line 105: There were a couple `tsc` errors that stem from some incompatibilities in some recent versions of `@types/node`. Namely the addition of `SharedArrayBuffer`. This is a simple solution that is sound and works well. I think a "proper" solution would involve standardizing on a single version of `@types/node` throughout the monorepo (there are currently several different versions floating around) but I didn't want to touch too much code.

jenseng:
Since all actions runners should now be using node 20 or greater, it might even be worth dropping CJS altogether, since require(ESM) is now possible 🤞

Regardless of the approach taken, big +1 on at least adding ESM output. This will make it possible to treeshake down to just a few KB (depending on what's needed), rather than almost a megabyte (since things like undici get pulled in, etc).

danwkennedy:
We've done a big push to upgrade these packages to ESM. We don't have plans to support both CJS and ESM. Closing this now.

PR #2131, opened 2025-09-24
Use native AbortController

trivikr:
Artifact already uses native AbortController
https://github.com/actions/toolkit/blob/f58042f9cc16bcaa87afaa86c2974a8c771ce1ea/packages/artifact/src/internal/upload/blob-upload.ts#L32


trivikr:
Closing as the custom AbortController dependency was removed in https://github.com/actions/toolkit/commit/ae29a2751b6d0fe32eae88ebb45b35cad5d9769f

PR #2161, opened 2025-10-17
fix: bumping max list artifact to 2k

functionstackx:
running into use cases where each workflow has more than 1k artifacts that i need to download

### DIFF
--- a/packages/artifact/src/internal/find/list-artifacts.ts
+++ b/packages/artifact/src/internal/find/list-artifacts.ts
@@ -59,7 +59,7 @@ export async function listArtifactsPublic(
   const totalArtifactCount = listArtifactResponse.total_count
   if (totalArtifactCount > maximumArtifactCount) {
     warning(
-      `Workflow run ${workflowRunId} has more than ${maximumArtifactCount} artifacts. Results will be incomplete as only the first ${maximumArtifactCount} artifacts will be returned`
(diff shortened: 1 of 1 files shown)


packages/artifact/src/internal/find/list-artifacts.ts
```diff
@@ -11,8 +11,8 @@ import {internalArtifactTwirpClient} from '../shared/artifact-twirp-client'
 import {getBackendIdsFromToken} from '../shared/util'
 import {ListArtifactsRequest, Timestamp} from '../../generated'
 
-// Limiting to 1000 for perf reasons
-const maximumArtifactCount = 1000
+// Limiting to 2000 for perf reasons
+const maximumArtifactCount = 2000
```
Link-, line 15: @functionstackx - Instead of making this change, I recommend that you add a helper function to `actions/toolkit/packages/artifact/src/internal/shared/config.ts` which will read an environment variable let's call it: `ACTIONS_ARTIFACT_MAX_ARTIFACT_LIST_COUNT`

The default being `1000` if this env var is not set.

This way we make this configurable and avoid causing others to hit their primary rate limits with this change.

Link-: The change here is not sufficient alone of course, we'll need to cut a new release for this package, pull it in: https://github.com/actions/download-artifact and also cut a new minor release for that action.

austenstone:
Continuing the work here
https://github.com/actions/toolkit/pull/2165

PR #2218, opened 2025-12-14
feat(toolkit/cache): add compression level support

### LINKED ISSUE #1535: Allow caching without compression

### LINKED ISSUE #2050: [Feature request] Allow configuration of zstd compression level
I'd like to configure zstd compression for my cache, and was surprised to find this is not configurable. Can this be supported by an env var?

For example, [sccache](https://github.com/mozilla/sccache/blob/main/docs/Configuration.md) supports `SCCACHE_CACHE_ZSTD_LEVEL` (default: 3) which significantly modifies cache size.

My understanding is that this repo is the source of the zstd functionality in the `actions/cache` Action (and in turn, the likes of `prefix-dev/setup-pixi` which delegate caching to the cache action and hence to this toolkit action).

As such, the ability to configure the zstd compression level in this toolkit action would be immediately useful to me if it allowed configuration via environment variable (as is done in sccache).

StephenHodgson:
- Added compression level support end-to-end: option plumbed through options/cache/cacheHttpClient/tar so tar creation sets gzip/zstd env and upload options carry the level.
- Expanded tests: updated saveCache/saveCacheV2 expectations for new parameter, added size-driven compression tests in tar suite, adjusted options tests for clamping/overrides, and kept tar tests prettier-compliant.
- Docs: refreshed cache README to describe compression-level behavior and defaults.
- closes #1535
- closes #2050
- blocks https://github.com/actions/cache/pull/1689

### DIFF
--- a/packages/cache/src/internal/tar.ts
+++ b/packages/cache/src/internal/tar.ts
@@ -237,6 +237,8 @@ async function getCompressionProgram(
 ): Promise<string[]> {
   const cacheFileName = utils.getCacheFileName(compressionMethod)
   const normalizedCompressionLevel = normalizeCompressionLevel(compressionLevel)
+  // zstd treats level 0 as invalid (gzip supports it), so clamp zstd to 1+ to avoid errors
+  // See: https://github.com/facebook/zstd/issues/1680
   const zstdCompressionLevel = Math.max(1, normalizedCompressionLevel)
   const BSD_TAR_ZSTD =
     tarPath.type === ArchiveToolType.BSD &&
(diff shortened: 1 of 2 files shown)


planetmarshall:
#1877 Implements similar functionality

StephenHodgson:
@planetmarshall similar but not the same.

Issue #2226, opened 2026-01-03
tool-cache downloadTool does not support relative redirects


tobil4sk:
**Describe the bug**
If `tc.downloadTool` is used with a url that returns a relative redirect then the function fails with the error: `TypeError: Invalid URL` as it attempts to parse the relative url as an absolute url.

**To Reproduce**

1. Apply this patch with a test to check the behaviour:
```diff
diff --git a/packages/tool-cache/__tests__/tool-cache.test.ts b/packages/tool-cache/__tests__/tool-cache.test.ts
index 4712ab5f..c4329275 100644
--- a/packages/tool-cache/__tests__/tool-cache.test.ts
+++ b/packages/tool-cache/__tests__/tool-cache.test.ts
@@ -111,6 +111,22 @@ describe('@actions/tool-cache', function () {
     expect(fs.statSync(downPath).size).toBe(35)
   })
 
+  it('downloads a 35 byte file after a relative redirect', async () => {
+    nock('http://example.com')
+      .persist()
+      .get('/relative-redirect')
+      .reply(303, undefined, {
+        location: 'bytes/35'
+      })
+
+    const downPath: string = await tc.downloadTool(
+      'http://example.com/relative-redirect'
+    )
+
+    expect(fs.existsSync(downPath)).toBeTruthy()
+    expect(fs.statSync(downPath).size).toBe(35)
+  })
+
   it('handles error from response message stream', async () => {
     nock('http://example.com')
       .persist()
```
2. Run the repository tests and view the `TypeError: Invalid URL` failure.

**Expected behavior**
The relative location should be resolved with the current url as the base instead of as an absolute url.

**Additional context**
The support for relative redirects is missing in the @actions/http-client package.

sverdlov93:
We're hitting this in production — our GitHub Action downloads a binary from a server that returns a **302 with a relative `Location` header**, and `@actions/http-client` crashes with `TypeError: Invalid URL` on every CI run.

The root cause is `new URL(redirectUrl)` at [packages/http-client/src/index.ts:405](https://github.com/actions/toolkit/blob/main/packages/http-client/src/index.ts#L405) — relative URLs are valid per RFC 7231 §7.1.2 but require a base parameter: `new URL(redirectUrl, requestUrl)`.

We've had to work around this by resolving redirects manually before calling `tc.downloadTool`, which shouldn't be necessary.

PR #2227 has the correct one-line fix. Would be great to see it merged and released.

PR #2227, opened 2026-01-03
fix(http-client): fix relative redirects

tobil4sk:
See: #2226

This solves the root cause of the bug, but for tool-cache to work correctly, the tool-cache's http-client version needs to be updated with this patch.

URL constructor parameter documentation can be found here:
https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#parameters

### DIFF
--- a/packages/http-client/__tests__/basics.test.ts
+++ b/packages/http-client/__tests__/basics.test.ts
@@ -159,6 +159,18 @@ describe('basics', () => {
     expect(obj.url).toBe('https://postman-echo.com/get')
   })
 
+  it('does basic get request with hostname and relative redirects', async () => {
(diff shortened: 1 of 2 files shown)


sverdlov93:
+1 — this fix is correct and we need it. We're hitting the same issue in production: our download server returns a relative `Location` header on 302, and `new URL(redirectUrl)` without a base throws `TypeError: Invalid URL`. We've had to work around it by resolving redirects manually before calling `tc.downloadTool`.

`new URL(redirectUrl, info.parsedUrl)` is the right fix per RFC 7231 §7.1.2 — relative references in Location headers are valid and the base should be the request URI.

Would love to see this merged and released.

Issue #2217, opened 2025-12-14
@actions/glob@0.5.0 (dependency of @actions/cache) missing Node.js 24 update


MikeMcC399:
## Situation

[@actions/cache@5.0.1](https://github.com/actions/toolkit/blob/main/packages/cache/RELEASES.md) depends on:

- [@actions/glob@0.5.0](https://github.com/actions/toolkit/blob/main/packages/glob/RELEASES.md) which in turn depends on:

    - [@actions/core@1.11.1](https://github.com/actions/toolkit/blob/main/packages/core/RELEASES.md)
    - [@actions/exec@1.1.1](https://github.com/actions/toolkit/blob/main/packages/exec/RELEASES.md)
    - [@actions/io@1.1.3](https://github.com/actions/toolkit/blob/main/packages/io/RELEASES.md)

whereas [@actions/cache@5.0.1](https://github.com/actions/toolkit/blob/main/packages/cache/RELEASES.md) also depends:

- directly on the following higher major versions that are labeled as compatible with Node.js 24:

    - [@actions/core@2.0.1](https://github.com/actions/toolkit/blob/main/packages/core/RELEASES.md)
    - [@actions/exec@2.0.0](https://github.com/actions/toolkit/blob/main/packages/exec/RELEASES.md)
    - [@actions/io@2.0.0](https://github.com/actions/toolkit/blob/main/packages/io/RELEASES.md)

This leaves a mixture of pre Node.js 24 toolkit dependencies installed together with ones which have been updated for Node.js 24 compatibility.

Continues to be true to @actions/cache@5.0.2 (currently missing from release notes) - see https://github.com/actions/toolkit/issues/2239

## Steps to reproduce

```shell
cd $(mktemp -d)
npm install @actions/cache
npm ls --all | grep @actions
```

## Logs

```
$ npm ls --all | grep @actions
└─┬ @actions/cache@5.0.1
  ├─┬ @actions/core@2.0.1
  │ ├── @actions/exec@2.0.0 deduped
  │ └── @actions/http-client@3.0.0 deduped
  ├─┬ @actions/exec@2.0.0
  │ └── @actions/io@2.0.0 deduped
  ├─┬ @actions/glob@0.5.0
  │ ├─┬ @actions/core@1.11.1
  │ │ ├─┬ @actions/exec@1.1.1
  │ │ │ └── @actions/io@1.1.3
  │ │ └─┬ @actions/http-client@2.2.3
[...]

PR #2237, opened 2026-01-08
Bump version to 2.2.0 and update dependencies for @actions/core, @actions/github, and @actions/http-client

TingluoHuang:


salmanmkc (approved):

crazy-max:
If `@actions/glob` could be updated as well, that would be nice :pray: https://github.com/actions/toolkit/blob/d3a48dd52f5f55ff28d5547d428dea98fba8ca25/packages/glob/package.json#L40-L41

(relates to https://github.com/actions/toolkit/issues/2217)

bdehamer (approved):

PR #2243, opened 2026-01-16
Don't retry 429s returned from the cache service

GhadimiR:
These are only returned in exceptional situations, it's better to not slow down the run retrying these when cache is best effort.

### DIFF
--- a/packages/cache/src/internal/shared/cacheTwirpClient.ts
+++ b/packages/cache/src/internal/shared/cacheTwirpClient.ts
@@ -111,6 +111,7 @@ class CacheServiceClient implements Rpc {
         }
 
         // Handle rate limiting - don't retry, just warn and exit
+        // For more info, see https://docs.github.com/en/actions/reference/limits
         if (statusCode === HttpCodes.TooManyRequests) {
           const retryAfterHeader = response.message.headers['retry-after']
           if (retryAfterHeader) {


packages/cache/src/internal/shared/cacheTwirpClient.ts
```diff
@@ -109,6 +109,20 @@ class CacheServiceClient implements Rpc {
...
         }
+
+        // Handle rate limiting - don't retry, just warn and exit
+        if (statusCode === HttpCodes.TooManyRequests) {
+          const retryAfterHeader = response.message.headers['retry-after']
+          if (retryAfterHeader) {
+            const parsedSeconds = parseInt(retryAfterHeader, 10)
+            if (!isNaN(parsedSeconds) && parsedSeconds > 0) {
+              warning(
+                `You've hit a rate limit, your rate limit will reset in ${parsedSeconds} seconds`
```
Link-, line 121: Maybe it's best to also include a link to the actions limit page here so that folks can revisit the limits

packages/cache/RELEASES.md
```diff
@@ -1,5 +1,9 @@
 # @actions/cache Releases
 
+### 5.0.3
+
+Fail cache operations on rate limit errors from the cache service to prevent slowing down impacted runs [2243](https://github.com/actions/toolkit/pull/2243).
```
Link-:
```suggestion
Prevent retries for rate limited cache operations [2243](https://github.com/actions/toolkit/pull/2243).
```

packages/cache/src/internal/shared/cacheTwirpClient.ts
Link-: For posterity, we're not retrying on the retryAfter schedule to prevent workflow run delays. It's best to fail fast in this case. We might want to make this configurable though.

Link-:
For posterity, we're not retrying on the retry after schedule to prevent workflow run delays. It's best to fail fast in this case. We might want to make this configurable though.

PR #2253, opened 2026-01-26
Artifact download: don't unzip non-zip artifacts

danwkennedy:
## Description

Being able to download un-zipped artifacts would be really useful. To support this in the future and to avoid some bugs we've seen before, we can check the `content-type` header to make sure the artifact is a zip before we attempt to unzip it. Right now, if the file isn't a zip (which can happen if someone extends/writes their own `actions/upload-artifact`), `actions/download-artifact` will fail in the extraction.

## Changes:

- Check whether downloaded file is a zip. We first check the `content-type` but also the URL file extension for backwards compatibility reasons:
    - Extract if it is (unchanged behavior)
    - Skip extraction if it isn't (new behavior)
- Allow callers to pass a `skipDecompress` flag in case they want to override our new behavior 

Example output from a custom `actions/download-artifact` build:

Non-zipped:
```
Downloading raw file (non-zip) to: /workspaces/actions/runner/_layout/_work/direct-upload/direct-upload/GitHub_Invertocat_Black.pdf
SHA256 digest of downloaded artifact is 9d30aac7deaf046f3e6ca4a46a6d03bcc347c595c64521dc103f4bad313aaf65
Artifact download completed successfully.
Total of 1 artifact(s) downloaded
Download artifact has finished successfully
```

Zipped:
```
SHA256 digest of downloaded artifact is 76ef7d397dc0352904bab221f72893dcc65a6db9e37cf64eeec244a57f4dcfff
Artifact download completed successfully.
Total of 1 artifact(s) downloaded
Download artifact has finished successfully
```


.vscode/launch.json
zaataylor, line 1: Let's make sure we don't commit this one

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -75,49 +79,85 @@ export async function streamExtractExternal(
...
+      clearTimeout(timer)
+      if (hashStream) {
+        hashStream.end()
+        sha256Digest = hashStream.read() as string
+        core.info(`SHA256 digest of downloaded artifact is ${sha256Digest}`)
+      }
+      resolve({sha256Digest: `sha256:${sha256Digest}`})
+    }
+
+    if (isZip && !skipDecompress) {
```
zaataylor, line 153: 👍

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -75,49 +79,85 @@ export async function streamExtractExternal(
     )
   }
 
+  const contentType = response.message.headers['content-type'] || ''
+  const mimeType = contentType.split(';', 1)[0].trim().toLowerCase()
```
zaataylor, line 83: From IANA standard [here](https://www.iana.org/assignments/media-types/application/zip), I see the `application/zip` MIME type has no required or optional parameters.

If the split we're doing here is for that, do we need it? Unsure 🤔 Probably won't _hurt_ us to have it just in case.

packages/artifact/src/internal/download/download-artifact.ts
```diff
@@ -75,49 +79,85 @@ export async function streamExtractExternal(
...
+  const isZip =
+    mimeType === 'application/zip' ||
+    mimeType === 'application/x-zip-compressed' ||
+    mimeType === 'zip'
```
zaataylor: Is just plain `zip` actually a valid/expected type?

packages/artifact/__tests__/download-artifact.test.ts
```diff
@@ -641,12 +651,284 @@ describe('download-artifact', () => {
...
+        fixtures.workspaceDir
+      )
+
+      expect(mockHttpClient).toHaveBeenCalledWith(getUserAgentString())
+      // After decoding and sanitizing, should just be 'passwd'
+      const sanitizedFileName = 'passwd'
+      const savedFilePath = path.join(fixtures.workspaceDir, sanitizedFileName)
+      expect(fs.existsSync(savedFilePath)).toBe(true)
+      expect(fs.readFileSync(savedFilePath, 'utf8')).toBe(rawFileContent)
+    })
```
zaataylor, line 979: I'd like to see an equivalent check that nothing was written outside the directory for the _encoded_ malicious file name, much like what we have for the test above this one. The specific part of that test I'm talking about is this part:

```typescript
// Verify the file was NOT written outside the workspace directory
      const maliciousPath = path.resolve(
        fixtures.workspaceDir,
        maliciousFileName
      )
      expect(fs.existsSync(maliciousPath)).toBe(false)
```

packages/artifact/src/internal/download/download-artifact.ts
danwkennedy: Yeah, it's technically correct 😄 We were looking for the mime type and the one we were looking for happened to not have optionals. I say we keep it.

packages/artifact/src/internal/download/download-artifact.ts
danwkennedy: Yeah, we should drop that. Funnily enough, there's _another_ type `application/zip-compressed` that is actually the preferred one by [mimetype.io](https://mimetype.io/application/zip-compressed). I'll swap them out.

.vscode/launch.json
danwkennedy: I'm on the fence here. It actually makes debugging tests simpler since you already have the configuration set. I also lifted this from `actions/upload-artifact`:
https://github.com/actions/upload-artifact/blob/main/.vscode/launch.json

packages/artifact/__tests__/download-artifact.test.ts
danwkennedy: Why not both encoded and decoded? 😄

crazy-max:
So I guess maintainers changed their mind related to https://github.com/actions/toolkit/pull/1874#pullrequestreview-2655200101?

danwkennedy:
> So I guess maintainers changed their mind related to [#1874 (review)](https://github.com/actions/toolkit/pull/1874#pullrequestreview-2655200101)?

@crazy-max that's correct. However, we're being more intentional with that feature. You've already seen the changes to upload. We're also changing the backend to handle this more gracefully (right now, the backend assumes everything will be a zip no matter what).

GhadimiR (approved):

PR #2255, opened 2026-01-26
fix(github): import from core directly

### LINKED ISSUE #2254: Broken types upgrading to @actions/github v8
## Describe the bug
**@actions/github v8** is exporting broken types.

## To Reproduce
My CI (type check) breaking on upgrade should be self explanatory :
https://github.com/kevinmarrec/workflows/pull/53

Still, you can check locally the types going crazy by reproducing the behavior:
1. Clone https://github.com/kevinmarrec/workflows (main)
2. Install dependencies
3. Check types in **filesize-diff/src/comment.ts** ✅
4. Checkout to **renovate/actions-github-8.x** branch
5. Install dependencies
6. Restart TypeScript Server **AND/OR** Reload IDE to refresh types cache
7. Check types in **filesize-diff/src/comment.ts** ❌ **(any)**

## Expected behavior
As the release did not specify related breaking change (except minimal Node.js version), the upgrade must be smooth and the types correct.

## Additional context
Maybe unrelevant but here is a related Gemini conversation :

<details>
<summary>Conversation</summary>
<br>
It is clear from the files provided that the upgrade to @actions/github v8.0.0 introduces a significant regression in type inference. 

The issue stems from the fact that v8 is a major architectural shift. 

It upgrades internal @octokit dependencies from v5/v10 to v7/v17. 

These newer Octokit versions have transitioned to pure ESM and have changed how types are exported and merged, which the @actions/github wrapper is currently failing to propagate.

### ⚠️ Issue Report: Broken Type Inference in v8.0.0

There is a confirmed issue with the @actions/github package after the v8 upgrade where the getOctokit client loses its type definitions and defaults to any.

#### 1. Loss of Type Safety (The "Any" Leak)

As seen in your comparison screenshots, the TypeScript LSP (Language Server Protocol) is no longer able to resolve the return type of the Octokit client:

    In v7: The octokit constant is correctly typed as an instance of GitHub & { rest: RestEndpoints; ... }.

    In v8: The octokit constant is inferred as any.

#### 2. CI Failures (noImplicitAny)

Because the client is now any, every subsequent property access (like .rest.issues.listComments) also returns any.

    The Bug: When you pass a callback to a method like .find(), the parameters (e.g., comment) cannot be inferred.

    The Error: Your CI is failing with error TS7006: Parameter 'comment' implicitly has an 'any' type. This occurs because your project likely has strict type checking enabled, which forbids implicit any types.

#### 3. Underlying Dependency Conflict

The diff shows that @actions/github v8 upgraded @octokit/core to ^7.0.6 and @octokit/plugin-rest-endpoint-methods to ^17.0.0. These versions changed the way RestEndpointMethodTypes are structured. 

The wrapper in @actions/github has not been updated to correctly map these new structures to the GitHub class it exports.
Recommendation

Until a patch is released for @actions/github to fix the getOctokit return type, you have two options:

- Temporary Workaround: Manually cast the response of your Octokit calls. You will need to install @octokit/plugin-rest-endpoint-methods as a dev dependency and use the RestEndpointMethodTypes to explicitly type your variables.
- Rollback: Pin your version to ^7.0.0 in package.json and configure Renovate to ignore the v8.0.0 update until a fix (likely a v8.0.1 or v8.1.0) is available.

</details>

## References

#2249

43081j:
The core `package.json` doesn't export `dist-types`, so typescript can
no longer resolve this path. That means it also can't resolve anything
that depends on it so they all become `any`.

It seems we can just import from core directly.

Fixes #2254.

EDIT:

there's actually all sorts of places importing `dist-types`, but its certainly not in the export map. so what's the expectation here? do you want to export it or do you want to stop importing from `dist-types`?

### DIFF
--- a/packages/github/src/github.ts
+++ b/packages/github/src/github.ts
@@ -2,7 +2,7 @@ import * as Context from './context'
 import {GitHub, getOctokitOptions} from './utils'
 
 // octokit + plugins
-import {OctokitOptions, OctokitPlugin} from '@octokit/core/dist-types/types'
+import {OctokitOptions, OctokitPlugin} from '@octokit/core/types'
 
 export const context = new Context.Context()
 
--- a/packages/github/src/internal/utils.ts
+++ b/packages/github/src/internal/utils.ts
@@ -1,6 +1,6 @@
 import * as http from 'http'
 import * as httpClient from '@actions/http-client'
-import {OctokitOptions} from '@octokit/core/dist-types/types'
+import {OctokitOptions} from '@octokit/core'
 import {ProxyAgent, fetch} from 'undici'
 
 export function getAuthString(
--- a/packages/github/src/utils.ts
+++ b/packages/github/src/utils.ts
@@ -2,8 +2,7 @@ import * as Context from './context'
 import * as Utils from './internal/utils'
 
 // octokit + plugins
-import {Octokit} from '@octokit/core'
-import {OctokitOptions} from '@octokit/core/dist-types/types'
+import {Octokit, type OctokitOptions} from '@octokit/core'
 import {restEndpointMethods} from '@octokit/plugin-rest-endpoint-methods'
 import {paginateRest} from '@octokit/plugin-paginate-rest'
 


43081j:
closing since #2260 duplicates this fix

PR #2256, opened 2026-01-26
Artifact upload: support uploading single un-zipped files

danwkennedy:
## Description

Sometimes we want to upload un-zipped files:

- Because the file is already compressed: we recommend uploading `tar` files to handle file permissions. Zipping those files is unhelpful since it requires double decompressing.
- Browsers can handle displaying common file types directly without needing to download the file to disk (think `pdf`, `png`/`jpg`, `json`, `html`, etc.). To support something like that in the UI, we can't have them zipped
- Some users only have a single file to upload. Zipping that is somewhat redundant and just causes more clicks to view/use the file.
- Mobile browsers don't really support zipped files (at least not well)

This PR adds support for those scenarios.

The new CI run for this PR demonstrates the browser behavior for this. Note that this is enabled for this repo only, please wait until the feature is fully rolled out before attempting elsewhere.

https://github.com/user-attachments/assets/3ce84346-31fa-43f9-ab1b-b558660d6f82

## Changes:

- Support skipping the zip step when uploading a single file
- Support passing the mime type when we create the artifact
- Added tests to the CI to validate

## Notes

- Only single file uploads are supported right now (future updates can add support for multiple file uploads).
- When `skipArchive` is `true`, the `name` parameter is replaced with the file name.
- The old zip flow should be fully supported with this change if `skipArchive` is `false`. Notably, this'll mean the file type is not present in the artifact name. The mime type passed will still be `application/zip`, though.

### DIFF
--- a/packages/artifact/src/internal/shared/interfaces.ts
+++ b/packages/artifact/src/internal/shared/interfaces.ts
@@ -54,6 +54,7 @@ export interface UploadArtifactOptions {
    * If true, the artifact will be uploaded without being archived (zipped).
    * This is only supported when uploading a single file.
    * When using this option, the artifact will not be compressed.
+   * When using this option, the name parameter passed to the upload is ignored. Instead, the name of the file is used as the name of the artifact.
    */
   skipArchive?: boolean
 }
--- a/packages/artifact/RELEASES.md
+++ b/packages/artifact/RELEASES.md
@@ -2,7 +2,7 @@
 
 ## 6.2.0
 
-- Support uploading single un-archived files (not zipped). Direct uploads are only supported for artifacts version 7+ (based on the major version of `actions/upload-artifact`). Callers must pass the `skipArchive` option to `uploadArtifact`. Only single files can be uploaded at a time right now. Default behavior should remain unchanged if `skipArchive = false`.
+- Support uploading single un-archived files (not zipped). Direct uploads are only supported for artifacts version 7+ (based on the major version of `actions/upload-artifact`). Callers must pass the `skipArchive` option to `uploadArtifact`. Only single files can be uploaded at a time right now. Default behavior should remain unchanged if `skipArchive = false`. When `skipArchive = true`, the name of the file is used as the name of the artifact for consistency with the downloads: you upload `artifact.txt`, you download `artifact.txt`.
 
 ## 6.1.0
 


crazy-max:
Thanks for this! The ability to upload single un-zipped files with the appropriate MIME type will unlock real workflows where ZIP packaging is actively harmful. I previously opened #1874, which aimed at improving download behavior when artifacts are not zipped. That PR was ultimately closed, but it surfaced a real pain point: today the toolkit and `actions/download-artifact` assume ZIP-formatted content, leading to broken downloads if the source isn't zipped.

While here it focuses on the upload side, it aligns with that same broader goal: support first-class non-zip artifacts. Together, these capabilities would cover both ends of the artifact lifecycle for non-zipped files and improve compatibility with existing external producers/consumers of artifacts. For example, workflows using gzip blobs (e.g., `docker/build-push-action`) currently fail during download because the toolkit can't extract non-zip content. So I guess the download case will be addressed as follow-up?

packages/artifact/src/internal/upload/types.ts
```diff
@@ -0,0 +1,82 @@
...
+  '.otf': 'font/otf',
+  '.eot': 'application/vnd.ms-fontobject'
+}
+
+/**
+ * Gets the MIME type for a file based on its extension
+ */
+export function getMimeType(filePath: string): string {
+  const ext = path.extname(filePath).toLowerCase()
+  return mimeTypes[ext] || 'application/octet-stream'
```
crazy-max, line 81: Would it be possible to have an extra opt to enforce the desired mime type? In our case we are uploading a `.dockerbuild` file that is `application/gzip` type.

crazy-max: Something similar to https://github.com/crazy-max/toolkit/commit/00e8704113236d10f1c2e51244108abe754eed95 so we can call `uploadArtifact` directly in https://github.com/crazy-max/docker-actions-toolkit/commit/8027d57aef4d4ac0adee542c876b1b9b421ac88f#diff-a248ff5ec76368168c745273fee0601b243545fdca37aa3bec76ef9e9365fdb7R172 as with ESM only we can't use internal helpers anymore: https://github.com/docker/actions-toolkit/pull/968#issuecomment-3823147317

crazy-max: Or maybe it is fine to have `application/octet-stream` if this is just use as content type in azure blob storage?

danwkennedy: Heh, I was literally going to ask that? I think it's actually ok to have an option for specifying though. I hadn't thought of files with no extension. I'm trying to come up with another example though, `.dockerbuild` would probably work fine since I'd expect us to default to a basic binary stream 🙂 . The types mostly help with specifying how the browser should handle the download in the UI: a PDF or PNG will be opened in a tab in the browser instead of downloading, etc.

crazy-max: > I think it's actually ok to have an option for specifying though.

Would be great!

> I hadn't thought of files with no extension. 

Actually `.dockerbuild` is the file extension :sweat_smile:

For example one of the artifact in this run: https://github.com/docker/build-push-action/actions/runs/21586122185

image

Like: https://github.com/docker/build-push-action/actions/runs/21586122185/artifacts/5341267407

> The types mostly help with specifying how the browser should handle the download in the UI: a PDF or PNG will be opened in a tab in the browser instead of downloading, etc.

Ah ok that makes sense! So maybe a `contentType` opt is not necessary then.

crazy-max:
@danwkennedy Is there any sense of timing for when this might move forward? This PR would unblock https://github.com/docker/actions-toolkit/pull/972#issuecomment-3842531063. Thanks 🙏

danwkennedy:
@crazy-max sorry for the wait. We're going to pick up the rollout next week at the earliest.

crazy-max:
> @crazy-max sorry for the wait. We're going to pick up the rollout next week at the earliest.

Awesome thx!

packages/artifact/src/internal/upload/upload-artifact.ts
```diff
@@ -12,33 +13,53 @@ import {
...
+    }
+
+    if (files.length > 1) {
+      throw new Error(
+        'skipArchive option is only supported when uploading a single file'
+      )
+    }
+
+    artifactFileName = path.basename(files[0])
+    name = artifactFileName
```
philip-gai, line 51: This will silently replace the user-provided filename, which might be surprising behavior 😄 If they skip archiving could they still keep the name they provided?

packages/artifact/src/internal/shared/interfaces.ts
```diff
@@ -50,6 +50,12 @@ export interface UploadArtifactOptions {
    * For large files that are not easily compressed, a value of 0 is recommended for significantly faster uploads.
    */
   compressionLevel?: number
+  /**
+   * If true, the artifact will be uploaded without being archived (zipped).
+   * This is only supported when uploading a single file.
+   * When using this option, the artifact will not be compressed.
+   */
+  skipArchive?: boolean
```
philip-gai, line 59: I think having an optional `mimeType?: string` property here would be good. Allowing the caller to override would keep the auto-detect as the default while giving users an escape hatch.

packages/artifact/src/internal/upload/upload-artifact.ts
```diff
@@ -12,33 +13,53 @@ import {
...
   validateArtifactName(name)
   validateRootDirectory(rootDirectory)
 
   const zipSpecification: UploadZipSpecification[] = getUploadZipSpecification(
     files,
     rootDirectory
   )
-  if (zipSpecification.length === 0) {
+
+  if (!options?.skipArchive && zipSpecification.length === 0) {
```
philip-gai: Consider:
1. Skipping `getUploadZipSpecification()` entirely when `skipArchive` is true, or
2. Adding an explicit `fs.existsSync(files[0])` check in the `skipArchive` block above.

packages/artifact/src/internal/upload/stream.ts
```diff
@@ -0,0 +1,53 @@
...
+  // Create a read stream from the file and pipe it to the upload stream
+  const fileStream = fs.createReadStream(sourcePath, {
+    highWaterMark: bufferSize
+  })
+
+  fileStream.on('error', error => {
+    core.error('An error has occurred while reading the file for upload')
+    core.info(String(error))
+    uploadStream.destroy(
+      new Error('An error has occurred during file read for the artifact')
```
philip-gai, line 46: The original error details are logged at `info` level while the generic message is at `error` level but I think we should put the error details in error level logging right?

packages/artifact/src/internal/upload/upload-artifact.ts
```diff
@@ -51,7 +72,8 @@ export async function uploadArtifact(
     workflowRunBackendId: backendIds.workflowRunBackendId,
     workflowJobRunBackendId: backendIds.workflowJobRunBackendId,
     name,
-    version: 4
+    mimeType: StringValue.create({value: contentType}),
+    version: 7
```
philip-gai, line 83: Why did this jump from version 4 to 7?

packages/artifact/src/generated/results/api/v1/artifact.ts
```diff
@@ -852,7 +573,7 @@ export const ListArtifactsRequest = new ListArtifactsRequest$Type();
 class ListArtifactsResponse$Type extends MessageType<ListArtifactsResponse> {
     constructor() {
         super("github.actions.results.api.v1.ListArtifactsResponse", [
-            { no: 1, name: "artifacts", kind: "message", repeat: 1 /*RepeatType.PACKED*/, T: () => ListArtifactsResponse_MonolithArtifact }
+            { no: 1, name: "artifacts", kind: "message", repeat: 2 /*RepeatType.UNPACKED*/, T: () => ListArtifactsResponse_MonolithArtifact }
```
philip-gai, line 576: This is a breaking wire-format change right? Has the server-side code been updated to match?

philip-gai (commented):
Some comments, lmk if you have any questions! 🚀

packages/artifact/src/internal/upload/upload-artifact.ts
danwkennedy: I thought of this and there are a few issues with that:
- If the caller points at a file directly, doesn't that also indicate naming intent?
- The caller is not required to provide a name, in which case it's `artifact`
- What do we do with file extension? Do we override just that?

In the end, the first issue was more compelling: if you tell me to upload `report.xml`, I would expect to see that downloaded. If you want to rename that file, you can do that before uploading it 🙂

packages/artifact/src/internal/upload/upload-artifact.ts
danwkennedy: I was matching the action version for `actions/upload-artifact`, not the toolkit package version. They matched for `v4` but since then, we've had breaking action changes but not toolkit changes. Since most callers will be using the action, I went with that.

packages/artifact/src/generated/results/api/v1/artifact.ts
danwkennedy: Yeah, this was an unrelated proto change that didn't get propagated until I pulled my changes for this.

packages/artifact/src/internal/upload/stream.ts
danwkennedy: Good catch.

packages/artifact/src/internal/upload/upload-artifact.ts
philip-gai: Yeah I think that makes sense, considering `name` before was the name of the whole artifact, now `name` is just the name of a single file so its not used the same way it was before anyway 👍🏼

packages/artifact/src/internal/shared/interfaces.ts
philip-gai: Can be a future enhancement!

packages/artifact/src/internal/upload/upload-artifact.ts
philip-gai: Maybe we need a note on the `name` option to say that it does NOT get used for single-file uploads

packages/artifact/src/internal/shared/interfaces.ts
danwkennedy: Yeah, this isn't possible yet with the backend the way that it is. For now, we need a strong link between the extension in the name and the mime type. We've future proofed the API but we still need to do some work.

packages/artifact/src/internal/upload/upload-artifact.ts
danwkennedy: Done

packages/artifact/src/internal/upload/upload-artifact.ts
danwkennedy: Added to doc block for the `skipArchive` option as well as the release notes.

crazy-max:
@danwkennedy Related to https://github.com/actions/toolkit/pull/2256#pullrequestreview-3727699188 I made some tests on this repo https://github.com/crazy-max/upload-artifact-skiparchive/actions/runs/22435286802 using latest changes from main of https://github.com/actions/upload-artifact.

```yaml
jobs:
  json:
    runs-on: ubuntu-latest
    steps:
      -
        name: Checkout
        uses: actions/checkout@v6
      -
        name: Upload artifact
        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f
        with:
          name: github-repo-1c96c578b3.json
          path: github-repo-1c96c578b3.json
          archive: false

  dockerbuild:
    runs-on: ubuntu-latest
    steps:
      -
        name: Checkout
        uses: actions/checkout@v6
      -
        name: Upload artifact
        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f
        with:
          name: docker~actions-toolkit~LF571I.dockerbuild
          path: docker~actions-toolkit~LF571I.dockerbuild
          archive: false
```

It seems to work fine for the `json` job and downloaded artifact is not a zip one as expected: https://github.com/crazy-max/upload-artifact-skiparchive/actions/runs/22435286802/artifacts/5669038505 but for the `dockerbuild` job with a `.dockerbuild` file it still makes a zip archive: https://github.com/crazy-max/upload-artifact-skiparchive/actions/runs/22435286802/artifacts/5669038532

danwkennedy:
@crazy-max Could you rerun the whole pipeline? I've enabled the feature on that repo explicitly (I know the `json` flow worked but I want to make sure). This time, enable debug logs on the action as well (there's a checkbox on the rerun popup that'll do it for you).

crazy-max:
> @crazy-max Could you rerun the whole pipeline? I've enabled the feature on that repo explicitly (I know the `json` flow worked but I want to make sure). This time, enable debug logs on the action as well (there's a checkbox on the rerun popup that'll do it for you).

@danwkennedy Thanks! Looks good after re-run: https://github.com/crazy-max/upload-artifact-skiparchive/actions/runs/22435286802

Artifact: https://github.com/crazy-max/upload-artifact-skiparchive/actions/runs/22435286802/artifacts/5672814224

danwkennedy:
@crazy-max cool, I haven't fully rolled out the final flag (it's a percentage right now so I don't know which repos have it and which don't). The request/response look correct when I tested it in the browser:

Screenshot 2026-02-26 at 08 40 02

PR #2125, opened 2025-09-08
fix(core): eliminate import cycles, improve ESM friendliness

justfalter:
This change eliminates two import cycles found in `@actions/core`.
These import cycles would prevent rollupjs from removing unused code
within `@actions/core`, bloating the resultant packaged js. 
This change also marks `@actions/core` as having no side-effects, allowing
for more efficient code-elimination when tree-shaking.

Anecdotally, with these changes, a project I work on that only uses core's `info` and
`setState` was reduced in size from 1.0MiB to approximately 120KiB.

- `core.ts` <-> `oidc-utils.ts`:
  - `core.ts` importing from `oidc-utils.ts` for the purpose of
    forming and exporting `getIDToken()`.
  - `oidc-utils.ts` was importing `debug()` and `setSecret` from
    `core.ts`.
- `core.ts` <-> `utils.ts`:
  - `core.ts` importing `toCommandProperties` and `toCommandValue` from
    `utils.ts`.
  - `utils.ts` importing `AnnotationProperties` from `core.ts`.

I resolved these issues by:

- Moving the definition of `AnnotationProperties` from `core.ts` into
    `annotation.ts`, and having `utils.ts` import from there.
- Moving `getIDToken` from `core.ts` into `oidc-utils.ts`.
- Moving `core.ts` -> `api.ts`
- Moving all package exports back into `core.ts` from `api.ts`, including the export
of `getIDToken`.

Partially fixes #1436, where one of the import cycles were first
described.

PR #2272, opened 2026-01-28
`@actions/core`: convert to ESM module

danwkennedy:
## Description

We're upgrading all the packages in this repo to ESM modules. `@actions/core` can go next since `@actions/exec` and `@actions/http-client` are already modules.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -6,10 +6,13 @@ module.exports = {
   testMatch: ['**/__tests__/*.test.ts'],
   moduleNameMapper: {
     '^(\\.{1,2}/.*)\\.js$': '$1',
+    '^@actions/core$': '<rootDir>/packages/core/lib/core.js',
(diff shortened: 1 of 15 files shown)


lokesh755 (approved):

schneidergithub:
Man this breaking change might create havoc and a support surge, good luck!

esatterwhite:
@danwkennedy I'm not sure I see the point to exclude common js here. node 20+ support both equally fine

adding support for both is just 1 config change. no code changes


```json
  "exports": {
    ".": {
      "types": "./lib/core.d.ts",
      "import": "./lib/core.js",
      "default": "./lib/core.js"
    }
  },
```

justfalter:
It would have been very helpful if the import cycles in `@actions/core` were eliminated --- almost none of the code can be eliminated: https://github.com/actions/toolkit/pull/2125

PR #2273, opened 2026-01-29
`@actions/glob`: convert to an ESM module

danwkennedy:
## Description

We're upgrading all the packages in this repo to ESM modules. `@actions/glob` can go next since all its dependencies are ESM modules.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -14,7 +14,8 @@ module.exports = {
     '^@actions/http-client/lib/auth$': '<rootDir>/packages/http-client/lib/auth.js',
     '^@actions/http-client/lib/interfaces$': '<rootDir>/packages/http-client/lib/interfaces.js',
     '^@actions/github$': '<rootDir>/packages/github/lib/github.js',
-    '^@actions/github/lib/utils$': '<rootDir>/packages/github/lib/utils.js'
+    '^@actions/github/lib/utils$': '<rootDir>/packages/github/lib/utils.js',
+    '^@actions/glob$': '<rootDir>/packages/glob/lib/glob.js'
   },
   transform: {
     '^.+\\.(ts|js)$': ['ts-jest', {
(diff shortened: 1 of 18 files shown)


lokesh755 (approved):

rajuchula12-coder:
- uses: actions/upload-artifact@v7
  with:
    # Name of the artifact to upload.
    # Optional. Default is 'artifact'
    name:

    # A file, directory or wildcard pattern that describes what to upload
    # Required.
    path:

    # The desired behavior if no files are found using the provided path.
    # Available Options:
    #   warn: Output a warning but do not fail the action
    #   error: Fail the action with an error message
    #   ignore: Do not output any warnings or errors, the action does not fail
    # Optional. Default is 'warn'
    if-no-files-found:

    # Duration after which artifact will expire in days. 0 means using default retention.
    # Minimum 1 day.
    # Maximum 90 days unless changed from the repository settings page.
    # Optional. Defaults to repository settings.
    retention-days:

    # The level of compression for Zlib to be applied to the artifact archive.
    # The value can range from 0 to 9.
    # For large files that are not easily compressed, a value of 0 is recommended for significantly faster uploads.
    # Optional. Default is '6'
    compression-level:

    # If true, an artifact with a matching name will be deleted before a new one is uploaded.
    # If false, the action will fail if an artifact for the given name already exists.
    # Does not fail if the artifact does not exist.
    # Optional. Default is 'false'
    overwrite:

    # Whether to include hidden files in the provided path in the artifact
    # The file contents of any hidden files in the path should be validated before
    # enabled this to avoid uploading sensitive information.
    # Optional. Default is 'false'
    include-hidden-files:

    # Whether to zip the artifact files before upload
    # If 'false', only a single file can be uploaded. The name of the file will be used as the artifact name (the 'name' parameter is ignored)
    # Optional. Default is 'true'
    archive:

PR #2298, opened 2026-02-07
fix(http-client): correctly handle ipv6 literals

cristicbz:
Right now, attempting to access a literal IPv6 endpoint results in a `getaddrinfo ENOTFOUND`, because `URL.hostname` returns bracketed IPv6 literals, like `[2001:db8::1]`, while node expects them unbracketed; the mismatch results in it trying to resolve the IPv6 address as a domain, and failing. 

Fixes https://github.com/actions/cache/issues/1718 .


cristicbz:
pin @danwkennedy maybe? don't wanna nag, just no idea how to get this moving

Issue #2305, opened 2026-02-19
CVE-2026-26996 high severity vulnerability in minimatch@3 used in @actions/glob & @actions/cache


MikeMcC399:
## Situation

`npm audit` reports a high severity vulnerability CVE-2026-26996 (https://github.com/advisories/GHSA-3ppc-4f35-3m26) in [minimatch@3.1.2](https://www.npmjs.com/package/minimatch/v/3.1.2) using [@actions/glob@0.6.1](https://github.com/actions/toolkit/blob/main/packages/glob/RELEASES.md)

[minimatch@3.1.2](https://www.npmjs.com/package/minimatch/v/3.1.2) is a legacy version, released on Feb 15, 2022

The latest version is [minimatch@10.2.1](https://www.npmjs.com/package/minimatch)

## Steps to reproduce

```shell
cd $(mktemp -d)
npm install @actions/cache @actions/glob
npm audit
```

## Logs

```text
$ npm audit
# npm audit report

minimatch  <10.2.1
Severity: high
minimatch has a ReDoS via repeated wildcards with non-matching literal in pattern - https://github.com/advisories/GHSA-3ppc-4f35-3m26
No fix available
node_modules/minimatch
  @actions/glob  *
  Depends on vulnerable versions of minimatch
  node_modules/@actions/glob
    @actions/cache  *
    Depends on vulnerable versions of @actions/glob
    node_modules/@actions/cache

3 high severity vulnerabilities

Some issues need review, and may require choosing
a different dependency.
```

## Suggestion

Update to [minimatch@10.2.1](https://www.npmjs.com/package/minimatch) or above

MikeMcC399:
This vulnerability affects also `@actions/cache` which has a direct dependency on `@actions/glob`

MikeMcC399:
See PR https://github.com/actions/toolkit/pull/2306 for a suggested partial solution

MikeMcC399:
Are there any maintainers available to comment?

MikeMcC399:
Closing, as the repro steps no longer show a vulnerability. Instead `minimatch@3.1.5` is installed, which now satisfies the requirements of https://github.com/advisories/GHSA-3ppc-4f35-3m26.

PR #2306, opened 2026-02-19
update minimatch to v10

isaacs:
This fixes a ReDOS and gets onto a supported version

### DIFF
--- a/packages/glob/package-lock.json
+++ b/packages/glob/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@actions/core": "^3.0.0",
-        "minimatch": "^3.0.4"
+        "minimatch": "^10.2.2"
       }
     },
     "node_modules/@actions/core": {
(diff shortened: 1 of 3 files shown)


MikeMcC399:
~~This would partially fix issue https://github.com/actions/toolkit/issues/2305~~ Edit: now closed

https://github.com/actions/toolkit/blob/main/README.md#note however says:

> ## Note
>
> Thank you for your interest in this GitHub repo, however, right now we are not taking contributions. 

Hopefully maintainers will take account of the suggestion in any case.

matherm-aboehm:
@MikeMcC399 
At the same time https://github.com/actions/toolkit/security says:

> If you discover a security issue in this repo, please submit it through the [GitHub Security Bug Bounty](https://hackerone.com/github)
>
> Thanks for helping make GitHub Actions safe for everyone.

Maybe it's better to report it there to get attention?

MikeMcC399:
@matherm-aboehm 

> Maybe it's better to report it there to get attention?

That normally only applies to unpublished security issues and that is a different process. I would not expect going through that route would speed anything up.

This issue requires attention from maintainers of this repo. It's their responsibility to monitor the repo and to respond.

MikeMcC399:
Also, I notice that `minimatch@3.1.5` is now installed, which remediates the vulnerability. So this PR is still relevant for good housekeeping purposes, however the criticality is now lower.

PR #2320, opened 2026-02-25
custom user-agent string for attestation API reqs

bdehamer:
This pull request updates the `@actions/attest` package to improve how the package version is accessed and to enhance the user agent string sent with requests to the attestation API.

Follows the pattern for including the `ACTIONS_ORCHESTRATION_ID` value in the user agent string introduced in https://github.com/actions/toolkit/pull/2229

### DIFF
--- a/packages/attest/src/package-version.cjs
+++ b/packages/attest/src/package-version.cjs
@@ -1,0 +1,7 @@
+// This file exists as a CommonJS module to read the version from package.json.
+// In an ESM package, using `require()` directly in .ts files requires disabling
+// ESLint rules and doesn't work reliably across all Node.js versions.
+// By keeping this as a .cjs file, we can use require() naturally and export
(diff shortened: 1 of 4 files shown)


packages/attest/src/package-version.cjs
```diff
@@ -0,0 +1,7 @@
+// This file exists as a CommonJS module to read the version from package.json.
+// In an ESM package, using `require()` directly in .ts files requires disabling
+// ESLint rules and doesn't work reliably across all Node.js versions.
+// By keeping this as a .cjs file, we can use require() naturally and export
+// the version for the ESM modules to import.
+const packageJson = require('../package.json')
+module.exports = {version: packageJson.version}
```
bdehamer, line 7: I picked-up this trick from the `@actions/cache` package -- allows us to read the version string from the `package.json`

PR #2357, opened 2026-03-23
@actions/glob: extend hashFiles options

### LINKED ISSUE #1035: Support hashing files outside the workspace directory
Currently [`hashFiles` ignores files outside the workspace](https://github.com/actions/toolkit/blob/main/packages/glob/src/internal-hash-files.ts#L16-L19). This means that it cannot be used to hash files which appear within an Action. In turn this means that packages such as `actions/setup-python` cannot be used with files from the Action (see for example https://github.com/actions/setup-python/issues/361), making them hard to use within composite Actions.

It would be great if `hashFiles` could support these files as it would enable more powerful composite Actions.

priyagupta108:
This pull request extends `@actions/glob` `hashFiles` to support safer and more flexible hashing by adding options for allowed `roots`, an opt-in to include files outside those roots (`allowFilesOutsideWorkspace`), and exclude patterns (`exclude`).

Closes #1035

### DIFF
--- a/packages/glob/__tests__/hash-files.test.ts
+++ b/packages/glob/__tests__/hash-files.test.ts
@@ -123,6 +123,114 @@ describe('globber', () => {
       '4e911ea5824830b6a3ec096c7833d5af8381c189ffaa825c3503a5333a73eadc'
     )
   })
+
+  it('hashes files in allowed roots only', async () => {
+    const root = path.join(getTestTemp(), 'roots-hashfiles')
+    const dir1 = path.join(root, 'dir1')
+    const dir2 = path.join(root, 'dir2')
+    await fs.mkdir(dir1, {recursive: true})
+    await fs.mkdir(dir2, {recursive: true})
+    await fs.writeFile(path.join(dir1, 'file1.txt'), 'test 1 file content')
+    await fs.writeFile(path.join(dir2, 'file2.txt'), 'test 2 file content')
+
+    const broadPattern = `${root}/**`
+
+    const hashDir1Only = await hashFiles(broadPattern, '', {roots: [dir1]})
+    expect(hashDir1Only).not.toEqual('')
+
+    const hashDir2Only = await hashFiles(broadPattern, '', {roots: [dir2]})
+    expect(hashDir2Only).not.toEqual('')
+
+    expect(hashDir1Only).not.toEqual(hashDir2Only)
+
+    const hashBoth = await hashFiles(broadPattern, '', {roots: [dir1, dir2]})
+    expect(hashBoth).not.toEqual(hashDir1Only)
+    expect(hashBoth).not.toEqual(hashDir2Only)
+
+    const hashDir1Again = await hashFiles(broadPattern, '', {roots: [dir1]})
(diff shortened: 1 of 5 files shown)


packages/glob/src/internal-hash-files.ts
```diff
@@ -4,41 +4,223 @@ import * as fs from 'fs'
...
+      } as IMinimatchOptions),
+      workspaceRelativeMatcher: new Minimatch(normalizedPattern, {
+        ...MINIMATCH_OPTIONS,
+        matchBase: isBasenamePattern
+      } as IMinimatchOptions)
+    }
+  })
+}
+
+function isExcluded(
```
luketomlinson, line 71: Will this exclude files _outside_ the default workspace? It seems it only looks for exclusions relative to the `workspaceForRelativeMatch`, should we look for exclusions relative to all allowed roots too?

packages/glob/src/internal-hash-files.ts
```diff
@@ -4,41 +4,223 @@ import * as fs from 'fs'
 import * as stream from 'stream'
 import * as util from 'util'
 import * as path from 'path'
+import minimatch from 'minimatch'
```
luketomlinson: why do we do this vs using the 
```ts
import {Minimatch, type MinimatchOptions} from 'minimatch'
```
That `internal-pattern.ts` uses

packages/glob/__tests__/hash-files.test.ts
```diff
@@ -123,6 +123,114 @@ describe('globber', () => {
...
+    const insideOnly = await hashFiles(`${dir1}/*`, '', {roots: [dir1]})
+    expect(insideOnly).not.toEqual('')
+
+    const patterns = `${dir1}/*\n${outsideDir}/*`
+    const defaultSkip = await hashFiles(patterns, '', {roots: [dir1]})
+
+    expect(defaultSkip).toEqual(insideOnly)
+  })
+
+  it('allows files outside roots if opted-in (hash changes)', async () => {
```
luketomlinson, line 178: it might be worth adding a test to test _outside_ the GITHUB_WORKSPACE which is set to `__dirname`. All the `getTestTemp` paths are relative to `__dirname` which is GITHUB_WORKSPACE

priyagupta108: Added a test that hashes files in a real dir outside GITHUB_WORKSPACE (via os.tmpdir()), asserting they're skipped by default and only included with `allowFilesOutsideWorkspace: true`.

packages/glob/src/internal-hash-files.ts
priyagupta108: Switched to the named import to match `internal-pattern.ts`.

packages/glob/src/internal-hash-files.ts
priyagupta108: Updated `isExcluded` to match relative exclude patterns against every allowed root (plus the workspace). Added a test covering exclusions across multiple roots.

luketomlinson (approved):

PR #2364, opened 2026-04-07
feat(github): add orchestration ID to user-agent in getOctokitOptions

salmanmkc:
## Summary

Appends `actions_orchestration_id/{sanitizedId}` to the user-agent string when the `ACTIONS_ORCHESTRATION_ID` environment variable is set. This follows the same pattern already used in `@actions/http-client` and `@actions/attest`.

### `packages/github/src/internal/utils.ts`
- Added `getUserAgentWithOrchestrationId(baseUserAgent?)` — reads env var, sanitizes with `/[^a-z0-9_.-]/gi`, appends to user-agent. Trims whitespace-only values.

### `packages/github/src/utils.ts`
- `getOctokitOptions()` now calls the helper to include orchestration ID automatically
- Re-exports `getUserAgentWithOrchestrationId` for downstream consumers (e.g. `actions/github-script`) via `@actions/github/lib/utils`

### `packages/github/__tests__/orchestration.test.ts`
- 14 deterministic unit tests covering:
  - Helper function (env var present/absent, sanitization, whitespace, allowed chars)
  - `getOctokitOptions` integration (auto-append, preserve caller UA, no mutation)
  - Public re-export wiring verification

### `packages/github/RELEASES.md`
- Added 9.1.0 release notes

## Why

Actions orchestration needs to track which orchestration triggered API calls. The http-client and attest packages already do this — `@actions/github` was the missing piece. This unblocks [actions/github-script#708](https://github.com/actions/github-script/pull/708) from needing a local workaround.

## Testing

All 14 tests pass. No existing tests affected.

Part of https://github.com/github/c2c-actions/issues/10001

### DIFF
--- a/packages/github/src/internal/utils.ts
+++ b/packages/github/src/internal/utils.ts
@@ -49,8 +49,10 @@ export function getUserAgentWithOrchestrationId(
   const orchId = process.env['ACTIONS_ORCHESTRATION_ID']?.trim()
   if (orchId) {
     const sanitizedId = orchId.replace(/[^a-z0-9_.-]/gi, '_')
+    const tag = `actions_orchestration_id/${sanitizedId}`
+    if (baseUserAgent?.includes(tag)) return baseUserAgent
     const ua = baseUserAgent ? `${baseUserAgent} ` : ''
-    return `${ua}actions_orchestration_id/${sanitizedId}`
+    return `${ua}${tag}`
   }
   return baseUserAgent
 }
--- a/packages/github/src/internal/utils.ts
+++ b/packages/github/src/internal/utils.ts
@@ -42,3 +42,15 @@ export function getProxyFetch(destinationUrl): typeof fetch {
 export function getApiBaseUrl(): string {
   return process.env['GITHUB_API_URL'] || 'https://api.github.com'
 }
+
+export function getUserAgentWithOrchestrationId(
+  baseUserAgent?: string
+): string | undefined {
+  const orchId = process.env['ACTIONS_ORCHESTRATION_ID']?.trim()
+  if (orchId) {
+    const sanitizedId = orchId.replace(/[^a-z0-9_.-]/gi, '_')
+    const ua = baseUserAgent ? `${baseUserAgent} ` : ''
+    return `${ua}actions_orchestration_id/${sanitizedId}`
+  }
+  return baseUserAgent
+}
--- a/packages/github/src/utils.ts
+++ b/packages/github/src/utils.ts
@@ -23,6 +23,8 @@ export const GitHub = Octokit.plugin(
   paginateRest
 ).defaults(defaults)
 
+export {getUserAgentWithOrchestrationId} from './internal/utils.js'
+
 /**
  * Convience function to correctly format Octokit Options to pass into the constructor.
  *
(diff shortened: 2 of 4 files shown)


packages/github/src/internal/utils.ts
```diff
@@ -42,3 +42,15 @@ export function getProxyFetch(destinationUrl): typeof fetch {
...
 }
+
+export function getUserAgentWithOrchestrationId(
+  baseUserAgent?: string
+): string | undefined {
+  const orchId = process.env['ACTIONS_ORCHESTRATION_ID']?.trim()
+  if (orchId) {
+    const sanitizedId = orchId.replace(/[^a-z0-9_.-]/gi, '_')
+    const ua = baseUserAgent ? `${baseUserAgent} ` : ''
+    return `${ua}actions_orchestration_id/${sanitizedId}`
```
TingluoHuang: will concat the string together cause invalid user-agent format?

packages/github/src/utils.ts
```diff
@@ -23,6 +23,8 @@ export const GitHub = Octokit.plugin(
   paginateRest
 ).defaults(defaults)
 
+export {getUserAgentWithOrchestrationId} from './internal/utils.js'
+
 /**
  * Convience function to correctly format Octokit Options to pass into the constructor.
```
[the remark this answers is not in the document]
salmanmkc: was already there, not related to change, good thing to fix tho

packages/github/src/internal/utils.ts
salmanmkc: The concatenation should be fine — the User-Agent spec ([RFC 9110 §10.1.5](https://httpwg.org/specs/rfc9110.html#field.user-agent)) is just space-separated `product/version` tokens, and `actions_orchestration_id/<sanitizedId>` fits that since the ID is sanitized to `[a-z0-9_.-]`.

Octokit already concatenates the same way internally (`${options.userAgent} ${userAgentTrail}`), so the final header ends up looking like:

```
actions_orchestration_id/abc123 octokit-core.js/6.1.3 Node.js/20
```

Also worth noting this is the same approach used by [`@actions/http-client`](https://github.com/actions/toolkit/blob/main/packages/http-client/src/index.ts#L826) and [`@actions/attest`](https://github.com/actions/toolkit/blob/main/packages/attest/src/internal/utils.ts#L11) already in the toolkit.

packages/github/src/internal/utils.ts
```diff
@@ -42,3 +42,15 @@ export function getProxyFetch(destinationUrl): typeof fetch {
 export function getApiBaseUrl(): string {
   return process.env['GITHUB_API_URL'] || 'https://api.github.com'
 }
+
+export function getUserAgentWithOrchestrationId(
+  baseUserAgent?: string
+): string | undefined {
+  const orchId = process.env['ACTIONS_ORCHESTRATION_ID']?.trim()
+  if (orchId) {
```
ericsciple, line 50: should we check if base user agent already contains the orchestration ID? Or is that unlikely?

ericsciple (dismissed):

salmanmkc (commented):
Good catch — since `getUserAgentWithOrchestrationId` is exported as public API, downstream consumers (e.g., `github-script`) could call it directly while also using `getOctokitOptions()` which already appends the ID. A simple `.includes()` guard prevents double-appending.

ericsciple (approved):

sanjayravichandran15-gif (commented):
Delete

PR #2379, opened 2026-04-17
feat(cache): opt-in env var to prefer BSD tar on Windows

zeitlinger:
## Summary

Adds an opt-in env var `ACTIONS_CACHE_PREFER_BSD_TAR_ON_WINDOWS` that routes
cache archive/extract through `C:\Windows\System32\tar.exe` (libarchive bsdtar)
instead of Git-for-Windows' MSYS GNU tar. Default behavior is unchanged.

Refs [actions/cache#752](https://github.com/actions/cache/issues/752).

## Why

On hosted `windows-latest` runners, extract of a many-small-files cache is
dominated by MSYS tar's POSIX-shim + fork-per-file overhead. Concrete numbers
from a fresh bench on `windows-2025` hosted runner (568 MB mise install dir,
~40k files, zstd):

| archiver | create (s) | extract (s) | archive (MB) |
|---|---:|---:|---:|
| MSYS tar + zstd (current default) | 16–20 | ~80* | 567 |
| **bsdtar + zstd (this PR, opt-in)** | **14–16** | **16–19** | 568 |

*Measured in production CI (`jdx/mise-action` on flint). Bench harness below.

That's **~4× faster extract** with no archive-size penalty. bsdtar makes native
Win32 syscalls and has zstd built in (no external `zstd.exe` fork-per-file).

Windows Defender exclusions on the target dir only shave ~12 s off the MSYS
path and near-zero off bsdtar — confirming the bottleneck is tar, not AV scans.

## Why opt-in (not default)

- Zero risk to existing users — default path unchanged.
- BSD vs GNU tar have subtle behavioral differences (symlink handling, trailing
  slashes, permission mapping). Archives written by one are readable by the
  other in all formats this package uses (posix ustar + zstd), but switching
  the default deserves its own follow-up once this has soaked.
- Opt-in lets consumers like `jdx/mise-action` enable the fast path today.

## Design

`getTarPath()` on `win32`:

1. If `ACTIONS_CACHE_PREFER_BSD_TAR_ON_WINDOWS=true` **and** `C:\Windows\System32\tar.exe` exists → return BSD.
2. Else existing logic: GNU tar if present, else BSD if present.

The existing `BSD_TAR_ZSTD` two-command path (zstd + tar piped via temp file)
handles the BSD case end-to-end — no changes needed elsewhere.

## Test plan

- [x] New unit test `zstd extract tar prefers BSDtar when opt-in env var set`
      — mirrors the existing BSDtar test but keeps GNU tar available, sets the
      env var, asserts BSD path is selected and the two-command extract is
      invoked. All 12 tar tests pass.
- [x] `npx prettier --check` and `npx eslint` clean on changed files.
- [ ] Downstream verification: `actions/cache` consumer (e.g. a mise-action
      cache hit) with the env var set shows the extract-time improvement.

## Bench reproducer

Workflow used to produce the numbers above (ran on `windows-2025`, populated
mise install dir):
<https://github.com/grafana/flint/blob/ci/bench-win-extract/.github/workflows/bench-win-extract.yml>

### DIFF
--- a/packages/cache/__tests__/tar.test.ts
+++ b/packages/cache/__tests__/tar.test.ts
@@ -138,6 +138,8 @@ test('zstd extract tar prefers BSDtar when opt-in env var set', async () => {
   if (IS_WINDOWS) {
     const mkdirMock = jest.spyOn(io, 'mkdirP')
     const execMock = jest.spyOn(exec, 'exec')
+    const previousPreferBsdTar =
+      process.env['ACTIONS_CACHE_PREFER_BSD_TAR_ON_WINDOWS']
     // GNU tar still available — without the env var this would take the GNU path.
     jest
       .spyOn(utils, 'getGnuTarPathOnWindows')
@@ -185,7 +187,12 @@ test('zstd extract tar prefers BSDtar when opt-in env var set', async () => {
         }
       )
     } finally {
-      delete process.env['ACTIONS_CACHE_PREFER_BSD_TAR_ON_WINDOWS']
+      if (previousPreferBsdTar === undefined) {
+        delete process.env['ACTIONS_CACHE_PREFER_BSD_TAR_ON_WINDOWS']
+      } else {
+        process.env['ACTIONS_CACHE_PREFER_BSD_TAR_ON_WINDOWS'] =
+          previousPreferBsdTar
+      }
     }
   }
 })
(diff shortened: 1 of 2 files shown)


zeitlinger:
@robherley can you take a look?

PR #2380, opened 2026-04-17
feat:  add support for commonjs loaders

### LINKED ISSUE #2294: No "exports" main defined in @actions/core/package.json @v3
**Describe the bug**
Error when building any code using @v3.
 No "exports" main defined in @actions/core/package.json

**To Reproduce**
Try to build this code:

import { setOutput } from '@actions/core';
setOutput('x', 1);

**Expected behavior**
Code should build as normal.

**Screenshots**
n/a

 - OS: [e.g. iOS] linux
 - Browser [e.g. chrome, safari] n/a

**Additional context**

esatterwhite:
When this package was converted to ecmascript modules, it only defined an `import`
declaration in the package manifest making it impossible to load via
commonjs / require. Nodejs 22.12.0 added support for loading ecmascript
modules via require and commonjs modules via import. As long as the file
extension is `.js` and there is no top-level await, the same code can be
loaded by either mechanism. This change adds the `require` and `default`
declarations to the `exports` section of package.json making it possible
for this package to be used in either project type.

Fixes: #2294

### DIFF
--- a/packages/artifact/package.json
+++ b/packages/artifact/package.json
@@ -16,7 +16,9 @@
   "exports": {
     ".": {
       "types": "./lib/artifact.d.ts",
-      "import": "./lib/artifact.js"
+      "import": "./lib/artifact.js",
+      "require": "./lib/artifact.js",
+      "default": "./lib/artifact.js"
     }
   },
   "directories": {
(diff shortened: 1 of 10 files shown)


esatterwhite:
@danwkennedy thoughts?

PR #2383, opened 2026-04-21
release new versions for a few packages

aiqiaoy:


packages/glob/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@actions/glob",
-  "version": "0.6.1",
+  "version": "0.7.0",
```
salmanmkc, line 3: due to [this](https://github.com/actions/toolkit/blob/0607d7a54bc6b6da17f93071c3f3401143d16c05/packages/glob/RELEASES.md?plain=1#L5), i'm thinking since minimatch v3→v10 changes matching behavior (backslash handling, brace expansion), we may want to bump it via a major version?

packages/core/package-lock.json
salmanmkc, line 28: also are you going to do a npm install again to get http client to bump after releasing http client? would need a new pr after

packages/glob/package.json
aiqiaoy: Typically major version bump is when we introduce breaking change for the api. None of the methods in glob changed so I went with minor version bump

salmanmkc: cool sounds good

salmanmkc (approved):

Issue #798, opened 2021-05-07
Tool-Cache: .tar.gz tests fail on Windows


rcowsill:
**Describe the bug**
On Windows the tests for the tool-cache package currently don't all pass. The .tar.gz tests fail with an error saying `tar` wasn't found.

**To Reproduce**
On a Windows machine:
1. Clone actions/toolkit
2. On main branch, run the following commands:
    ```
    npm i
    npm run bootstrap
    npm run build
    npm run test -- packages/tool-cache
    ```

**Expected behavior**
All tests pass on Windows

**Actual behavior**
```
 FAIL  packages/tool-cache/__tests__/tool-cache.test.ts
  @actions/tool-cache
    V downloads a 35 byte file (16ms)
    V downloads a 35 byte file (dest) (20ms)
    V downloads a 35 byte file (dest requires mkdirp) (17ms)
    V downloads a 35 byte file after a redirect (18ms)
    V handles error from response message stream (71ms)
    V retries error from response message stream (23ms)
    V has status code in exception dictionary for HTTP error code responses (17ms)
    V works with redirect code 302 (15ms)
    V installs a binary tool and finds it (24ms)
    V installs a 7z and finds it (295ms)
    V extracts a 7z to a directory that does not exist (275ms)
    V extract 7z using custom 7z tool (37ms)
    × extract .tar.gz (20ms)
    × extract .tar.gz to a directory that does not exist (20ms)
    × extract .tar.xz (18ms)
    V installs a zip and finds it (440ms)
    V installs a zip and extracts it to specified directory (440ms)
    V extract zip to a directory that does not exist (433ms)
    V works with a 502 temporary failure (17ms)
    V doesn't retry 502s more than 3 times (18ms)
    V retries 429s (17ms)
    V doesn't retry 404 (13ms)
    V supports authorization headers (13ms)
    V supports custom headers (14ms)
    V supports authorization and custom headers (14ms)

  ? @actions/tool-cache > extract .tar.gz

    Unable to locate executable file: tar. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also verify the file has a valid extension for an executable file.

      188 |     if (!result) {
      189 |       if (ioUtil.IS_WINDOWS) {
    > 190 |         throw new Error(
          |               ^
      191 |           `Unable to locate executable file: ${tool}. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also verify the file has a valid extension for an executable file.`
      192 |         )
[...]

Issue #547, opened 2020-08-06
Is it possible to use cache service URL as a HTTP cache?


vlsi:
**Describe the enhancement**

It would be nice if caching could be used as an HTTP service. Modern build systems like Gradle, Bazel, Pants can use a remote cache to avoid building the same tasks across different builds.

The current `@actions/cache` allows to cache files, however, there's no way to fetch data on demand, and it is hard to make incremental updates to the cache.

Can the HTTP cache service be exposed to the build scripts? For instance, read-only for pull-requests, and read-write for branch-based builds.

**Additional information**

See https://cirrus-ci.org/guide/writing-tasks/#http-cache

fkorotkov:
I wrote an action to start a proxy server that has the compatible API. See https://github.com/cirruslabs/http-cache-action

vlsi:
There's Gradle-tailored action as well: https://github.com/burrunan/gradle-cache-action :)

vlsi:
@fkorotkov , FYI: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/1205#issue-701759426

```
Error: commitCache failed: ﻿{"$id":"1","innerException":null,"message":"Request was blocked due to exceeding usage of resource 'Count' in namespace ''.","typeName":"Microsoft.TeamFoundation.Framework.Server.RequestBlockedException, Microsoft.TeamFoundation.Framework.Server","typeKey":"RequestBlockedException","errorCode":0,"eventId":3000}
    at /home/runner/work/_actions/burrunan/gradle-cache-action/v1.4/dist/cache-action-entrypoint.js:15662:11
    at Generator.throw (<anonymous>)
    at rejected (/home/runner/work/_actions/burrunan/gradle-cache-action/v1.4/dist/cache-action-entrypoint.js:15615:27)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

fkorotkov:
Will be great to get some official comments about concurrent usage of the API. 🤔

PR #2393, opened 2026-05-01
Update minor versions of cache and artifact dependencies

Samirat:
Across the board updates to minor versions of dependencies of the cache and artifact actions in toolkit.

### DIFF
--- a/packages/artifact/RELEASES.md
+++ b/packages/artifact/RELEASES.md
@@ -1,5 +1,17 @@
 # @actions/artifact Releases
 
+## 6.2.2
+
+- Bump dependency versions ([#2393](https://github.com/actions/toolkit/pull/2393)):
+  - `@actions/core` to `^3.0.1`
+  - `@actions/github` to `^9.1.1`
+  - `@actions/http-client` to `^4.0.1`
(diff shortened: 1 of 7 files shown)


Mikhai56 (approved):
Okay

philip-gai (approved):

salmanmkc (approved):
probably not a concern but based on these prs, [`octokit/request.js` PR #798 - json-with-bigint](https://github.com/octokit/request.js/pull/798) and [`octokit/endpoint.js` PR #547 - BigInt path params](https://github.com/octokit/endpoint.js/pull/547)

these could break if any becomes a big int, however I don't think we'll ever reach that soon

https://github.com/actions/toolkit/blob/cf80afb3922317b98dd74d27cba7cf1032c1fe50/packages/artifact/src/internal/find/list-artifacts.ts#L177

https://github.com/actions/toolkit/blob/cf80afb3922317b98dd74d27cba7cf1032c1fe50/packages/artifact/src/internal/find/get-artifact.ts#L64

everything else looks good to me

Mikhai56 (approved):

Mikhai56 (approved):

PR #2414, opened 2026-05-20
feat: `@actions/cache` optional path validation during restore

jasongin:
# Optional path validation during cache restore

## New `restoreCache()` option

A new opt-in `DownloadOptions.pathValidation` on `@actions/cache` streams a downloaded cache archive through an in-process validator before handing it to system `tar` for extraction. Each entry's path (and, for symlinks/hardlinks, its target) must resolve under one of the declared cache `paths` — anything that escapes (`../...`, absolute, UNC, drive-relative, NUL bytes, unsupported entry types) is caught by the validator.

```js
await cache.restoreCache(paths, key, restoreKeys, {
  pathValidation: 'error' // 'off' (default) | 'warn' | 'error'
})
```

- `'off'` — legacy behavior, validation is skipped.
- `'warn'` — validate; on violations, log via `core.warning` / `core.debug` and extract anyway (legacy extraction, no allow-list).
- `'error'` — validate; on violations throw `CacheIntegrityError` (`code: 'PATH_VIOLATION'`, with `violations` attached). No files are extracted. On a clean archive, extraction is additionally **restricted to exactly the members the validator approved** (see below).

## Architecture: node-tar for listing, system `tar` for extraction — made sound with an approved-member allow-list

This is the biggest design call in the PR, and the area most changed in the latest revision.

**Listing/validation uses node-tar v7's `Parser` in-process.**
- Pure JS, deterministic across platforms, exposes structured entry metadata. No subprocess, no parsing `tar -tv` text output that varies by GNU/BSD/Windows.
- We're inspecting untrusted attacker-controlled bytes; doing it in JS avoids any shell-quoting/argv concern.
- node-tar surfaces `TAR_BAD_ARCHIVE` / `TAR_ENTRY_INVALID` / `TAR_ENTRY_ERROR` / `TAR_ABORT` as discrete codes we promote to parse errors.
- Gzip is handled natively; for zstd we spawn the system `zstd` binary for the same `--long=30` window behavior the extract path uses.

**Extraction stays on system `tar`.**
- The `tar -P` (`--absolute-names`) option must be used when extracting cache archives, because cache paths frequently live outside `$GITHUB_WORKSPACE` (`~/.npm`, `~/.cargo`, `~/.gradle/caches`, etc.). `cacheUtils.resolvePaths` rewrites every path relative to the workspace, so anything outside becomes `../../...`, which GNU tar refuses without `-P`, and node-tar's extractor refuses by default — swapping it in wholesale would break essentially every common dependency cache.
- The existing extract path is heavily tuned for cross-platform quirks (BSD/GNU detection, `--force-local`, `--delay-directory-restore`, `zstdmt` / `zstd -T0`, the BSD-tar-on-Windows-with-zstd workaround). Re-implementing all of that on node-tar's writer is a much larger, riskier change.
- Zero behavior change for users who don't opt in.

**Closing the listing↔extraction gap (new).** The original design relied on an informal argument that "anything the validator approves, system tar also accepts." That argument is weak against *path-channel parser differentials*: a crafted entry (notably via PAX extended headers) can make node-tar compute a safe-looking path while system `tar` places the bytes somewhere else entirely. Rather than move validation onto system `tar` — which would require GNU tar everywhere and break BSD-tar / Windows `tar.exe` / many self-hosted runners — this revision keeps node-tar for listing and instead **constrains the extractor to the validator's approved set**:

- In `'error'` mode with a clean archive, `listAndValidate` now returns the exact member names node-tar derived (`approvedNames`) alongside the violations.
- Those names are written to a temporary NUL-separated file (mode `0o600`, in `os.tmpdir()`, randomized name) and passed to system `tar` via `--null --no-recursion -T <file>`.
  - `--null` MUST precede `-T` on GNU tar.
  - `--no-recursion` prevents an approved directory from implicitly pulling in unapproved children.
  - On GNU tar we additionally pass `--no-wildcards --no-wildcards-match-slash --anchored` as defense in depth; bsdtar lacks these, but glob metacharacters are already rejected during validation, so its `fnmatch()`-based `-T` matching degrades to exact-name matching.
- The temp file is unlinked in a `finally`.

The effect: a member that system `tar` would extract to a different (escaped) path than node-tar computed simply **isn't on the list and is never extracted**. The parser differential becomes a no-op instead of a trusted assumption. `'warn'` mode deliberately does *not* use the allow-list — on any violation it falls back to extracting everything, matching legacy behavior.

**Performance tradeoff.** Validation is a full pre-pass over the archive: decompress + parse headers (entry bodies are drained, never written to disk), then system tar decompresses a second time to extract. Roughly a 2× decompress cost on the archive stream, no extra disk I/O at the destination. The extra pass — and the allow-list — are skipped entirely when validation is `'off'`.

## Parser-differential & PAX header defense (new)

Beyond the allow-list, this revision hardens the validator itself against attacks that exploit the gap between node-tar's parsing and that of real extractors:

- **Length-correct PAX re-parser (pax-reparse.ts).** node-tar v7 parses PAX extended-header bodies with a naive `split('\n')` (its own source notes `XXX Values with \n in them will fail this`). A PAX record is actually length-prefixed (`"<len> <key>=<value>\n"`), and GNU tar / libarchive consume exactly `<len>` bytes — so a value containing an embedded `"\n<len> path=<other>\n"` desynchronizes node-tar from every real extractor. The re-parser re-reads each captured PAX body byte-accurately and cross-checks `path` / `linkpath` against node-tar's view. Disagreement → `PAX_DESYNC`; a body node-tar treated as PAX that the re-parser can't fully account for → `PAX_PARSE_FAIL`.
- **Fail-closed PAX key allow-list.** Any PAX key outside a known-good set (POSIX.1-2017 §3.4.3 plus the `SCHILY.` / `GNU.` / `LIBARCHIVE.` / `POSIX.` namespaces) → `PAX_UNKNOWN_KEY`, so a future placement-affecting PAX extension can't silently slip past.
- **Unsupported / silently-ignored entries.** Entry types node-tar would silently drop are now surfaced as `UNSUPPORTED_TYPE` via an `ignoredEntry` listener, closing a class of "the validator never saw it, but tar might" bypasses.
- **Unsafe characters & glob metacharacters.** Entry paths containing `\n` / `\0` → `UNSAFE_CHAR`; paths containing `* ? [ ]` → `GLOB_METACHAR`. The glob check is also what lets the bsdtar `-T` allow-list degrade safely to exact-name matching.
- **Meta-body flood guard.** Captured PAX/global metadata bodies are bounded (`maxMetaEntrySize` of 1 MiB, `MAX_PENDING_META` cap) so a malicious archive can't exhaust memory via the validation pass.

## Other design notes

- **String enum, not boolean.** Three discrete modes map 1:1 onto behavior and leave room for future modes (e.g. `'quarantine'`).
- **Allowed roots derived from `paths`.** `deriveAllowedRoots` mirrors `cacheUtils.resolvePaths`: tilde / env-var expansion, glob-prefix stripping, etc.
- **`prepareAllowedRoots()`** pre-bakes per-root normalize/lowercase/trailing-sep forms once per archive, dropping the per-entry hot loop from O(N · R) to O(N + R).
- **Zstd lifecycle hardening:** 64 KiB stderr cap, `.catch(() => undefined)` on the pipeline / exit promises to suppress late rejections, `try/finally` `SIGTERM` if the child is still running on early exit.

## Violation codes

`ABSOLUTE_PATH`, `UNC_PATH`, `NUL_BYTE`, `OUTSIDE_ROOTS`, `LINK_OUTSIDE_ROOTS`, `UNSUPPORTED_TYPE`, `PAX_DESYNC`, `PAX_PARSE_FAIL`, `PAX_UNKNOWN_KEY`, `UNSAFE_CHAR`, `GLOB_METACHAR`.

## Known limitations

- **TOCTOU between validation and extraction.** The validator never touches the filesystem. A pre-existing symlink in the workspace that the archive then writes "through", or a concurrent process creating one between passes, won't be caught. (The allow-list constrains *which members* extract, not whether the destination tree is mutated under us.)
- **Case-sensitivity heuristic** (`caseInsensitiveFs()`) is a platform-level guess. On case-sensitive APFS or per-dir-case-sensitive NTFS it loosens (fewer violations) rather than tightens.
- **bsdtar `-T` defense-in-depth flags** (`--no-wildcards` etc.) are GNU-only; on BSD/Windows tar the allow-list relies on the validator's prior rejection of glob metacharacters to keep `-T` matching exact.

## Tests

All path-validation suites pass: **176 passed, 3 skipped** (zstd suites `describe.skip` when no `zstd` binary is on `PATH`). New/updated coverage in this revision:
- `__tests__/pax-reparse.test.ts` — unit tests for the length-correct PAX re-parser and cross-check (`PAX_DESYNC` / `PAX_PARSE_FAIL` / `PAX_UNKNOWN_KEY`), 100% line coverage of pax-reparse.ts.
- `__tests__/tarPathValidationAttacks.test.ts` — raw-tar attack builders covering the F1–F5 parser-differential bypass classes, `UNSAFE_CHAR` / `GLOB_METACHAR` / clean-PAX / meta-flood cases, plus end-to-end extraction tests against the real system `tar`.
- `__tests__/tarPathValidation.test.ts` — extended for the `'error'`-mode allow-list: asserts the `--null --no-recursion -T` args, the NUL-separated file contents, and temp-file cleanup.
- `__tests__/listAndValidate.test.ts` — updated for the new `{violations, approvedNames}` return shape.

See path-validation-test-plan.md for the full test-coverage description.

## Out of scope

- Replacing system `tar` for extraction (see above).
- Entry-count / total-bytes caps (cache backend enforces archive size at upload).
- Save-path validation (restore-only here).

## `actions/cache` update

[actions/cache#1761](https://github.com/actions/cache/pull/1761) is a related PR to actually use this path validation when restoring from cache. The path validation mode will be a new input to the action; initially the default will be `'warn'`, then we can consider changing the default to `'error'` in a major version update.

### DIFF
--- a/packages/cache/src/internal/listAndValidate.ts
+++ b/packages/cache/src/internal/listAndValidate.ts
@@ -172,9 +172,7 @@ async function streamArchiveTo(
         const tail = stderrTruncated ? ' (stderr truncated)' : ''
         reject(
           new Error(
-            `zstd ${cause}${
-              zstdStderr ? `: ${zstdStderr.trim()}` : ''
-            }${tail}`
+            `zstd ${cause}${zstdStderr ? `: ${zstdStderr.trim()}` : ''}${tail}`
           )
         )
       }
--- a/packages/cache/src/internal/listAndValidate.ts
+++ b/packages/cache/src/internal/listAndValidate.ts
@@ -1,0 +1,208 @@
+import {spawn} from 'child_process'
+import {createReadStream} from 'fs'
+import {Readable} from 'stream'
+import {pipeline} from 'stream/promises'
+import {Parser, ReadEntry} from 'tar'
+import {CompressionMethod} from './constants.js'
+import {
+  PathValidationViolation,
+  prepareAllowedRoots,
+  validateEntry
+} from './pathValidation.js'
+
+/**
+ * Stream the entries of a (possibly compressed) tar archive and validate each
+ * against the allowed roots. Does NOT extract any files — entries are read
+ * for header inspection only. Returns the list of violations (empty if the
+ * archive is clean).
+ *
+ * Throws an Error if the archive cannot be parsed (corrupt header,
+ * decompression failure, truncated stream, etc.). The caller is responsible
+ * for translating that into a CacheIntegrityError with code `PARSE_ERROR`.
+ */
+export async function listAndValidate(
+  archivePath: string,
+  compressionMethod: CompressionMethod,
+  allowedRoots: string[],
+  extractCwd: string
+): Promise<PathValidationViolation[]> {
+  const violations: PathValidationViolation[] = []
+
+  // Precompute the normalized / case-folded form of each allowed root once,
+  // so the per-entry containment check is a handful of string compares
+  // rather than an O(roots) re-normalization on every entry.
+  const preparedRoots = prepareAllowedRoots(allowedRoots)
+
+  // Captured parse error (if any). We don't throw synchronously from inside
+  // the parser's onwarn callback because doing so leaves the parser's
+  // internal stream in a half-broken state that hangs the surrounding
+  // pipeline. Instead we record the first critical warning and throw it
+  // after the pipeline completes.
+  let firstParseError: Error | undefined
+
+  const recordParseError = (code: string, message: string): void => {
+    if (firstParseError) return
+    firstParseError = new Error(`tar parse error (${code}): ${message}`)
+  }
+
+  // For gzip we let node-tar handle decompression internally (its built-in
+  // gzip support is mature). For zstd we spawn the system `zstd` binary so
+  // we get the same `--long=30` window-size handling as the existing
+  // extract codepath in tar.ts and avoid relying on Node's experimental
+  // zstd support.
+  const useNativeGzip = compressionMethod === CompressionMethod.Gzip
+  const parser = new Parser({
+    gzip: useNativeGzip,
+    // Disable strict mode so recoverable warnings (e.g. unknown extended
+    // headers) don't abort parsing. Real corruption is surfaced explicitly
+    // via the captured error below.
+    strict: false,
+    // Treat structural problems (bad archive, bad header, bad chksum) as
+    // hard parse errors — silently ignoring them would let a corrupt
+    // archive sail through validation. We DO NOT throw on softer warnings
+    // (extended headers, unknown PAX keys, etc.).
+    onwarn: (code, message) => {
+      if (
+        code === 'TAR_BAD_ARCHIVE' ||
+        code === 'TAR_ENTRY_INVALID' ||
+        code === 'TAR_ENTRY_ERROR' ||
+        code === 'TAR_ABORT'
+      ) {
+        recordParseError(code, message)
+      }
+    },
+    onReadEntry: (entry: ReadEntry) => {
+      try {
+        const result = validateEntry(
+          entry.path,
+          entry.linkpath || undefined,
+          entry.type,
+          preparedRoots,
+          extractCwd
+        )
+        if (!result.ok) {
+          violations.push({
+            path: entry.path,
+            linkpath: entry.linkpath || undefined,
+            resolved: result.resolved,
+            entryType: entry.type,
+            code: result.code,
+            reason: result.reason
+          })
+        }
+      } finally {
+        // Drain the entry so the parser advances. Without this the stream
+        // stalls waiting for the consumer to read the entry body.
+        entry.resume()
+      }
+    }
+  })
+
+  await streamArchiveTo(archivePath, compressionMethod, parser)
+
+  if (firstParseError) {
+    throw firstParseError
+  }
+  return violations
+}
+
+async function streamArchiveTo(
+  archivePath: string,
+  compressionMethod: CompressionMethod,
+  destination: NodeJS.WritableStream
+): Promise<void> {
+  const fileStream = createReadStream(archivePath)
+
+  if (compressionMethod === CompressionMethod.Gzip) {
+    // node-tar's Parser was constructed with `gzip: true`, so it handles
+    // decompression internally — just pipe the raw file stream in.
+    await pipeline(fileStream, destination)
+    return
+  }
+
+  // zstd-compressed archive. node-tar does not natively decompress zstd, so
+  // we shell out to the `zstd` binary the same way tar.ts does for the
+  // existing extract codepath. This adds one extra decompression vs. the
+  // existing extract step (which runs its own zstd), but only on archives
+  // where path validation is enabled.
+  const zstdArgs: string[] = ['-d', '-c']
+  if (compressionMethod === CompressionMethod.Zstd) {
+    zstdArgs.unshift('--long=30')
+  }
+
+  const zstd = spawn('zstd', zstdArgs, {
+    stdio: ['pipe', 'pipe', 'pipe']
+  })
+
+  // Cap stderr capture so a chatty/malicious zstd invocation can't grow
+  // memory without bound. 64 KiB is plenty for any realistic error message.
+  const STDERR_CAP_BYTES = 64 * 1024
+  let zstdStderr = ''
+  let stderrBytes = 0
+  let stderrTruncated = false
+  zstd.stderr.on('data', (chunk: Buffer) => {
+    if (stderrBytes >= STDERR_CAP_BYTES) {
+      stderrTruncated = true
+      return
+    }
+    const remaining = STDERR_CAP_BYTES - stderrBytes
+    if (chunk.length > remaining) {
+      zstdStderr += chunk.subarray(0, remaining).toString()
+      stderrBytes += remaining
+      stderrTruncated = true
+    } else {
+      zstdStderr += chunk.toString()
+      stderrBytes += chunk.length
+    }
+  })
+
+  const zstdExited = new Promise<void>((resolve, reject) => {
+    zstd.on('error', reject)
+    zstd.on('exit', (code, signal) => {
+      if (code === 0) {
+        resolve()
+      } else {
+        // A SIGTERM here means we killed the child ourselves during cleanup
+        // after an upstream failure — surface a clearer message in that
+        // case rather than the bare exit code.
+        const cause =
+          signal !== null
+            ? `terminated by signal ${signal}`
+            : `exited with code ${code}`
+        const tail = stderrTruncated ? ' (stderr truncated)' : ''
+        reject(
+          new Error(
+            `zstd ${cause}${
+              zstdStderr ? `: ${zstdStderr.trim()}` : ''
+            }${tail}`
+          )
+        )
+      }
+    })
+  })
+
+  const stdin = zstd.stdin as unknown as NodeJS.WritableStream
+  const stdout = zstd.stdout as unknown as Readable
+
+  // Run both sides of the pipeline concurrently, plus wait for the zstd
+  // process itself to exit cleanly. If decompression produces non-tar bytes
+  // the destination parser will reject; if zstd exits non-zero the exit
+  // promise will reject. Either way we surface the error.
+  const inPromise = pipeline(fileStream, stdin)
+  const outPromise = pipeline(stdout, destination)
+  // Suppress unhandled-rejection warnings on the individual promises. The
+  // first rejection is propagated via the Promise.all below; any later
+  // rejection (e.g. zstd reporting a non-zero exit after the parser
+  // already errored) would otherwise crash the process.
+  inPromise.catch(() => undefined)
+  outPromise.catch(() => undefined)
+  zstdExited.catch(() => undefined)
+
+  try {
+    await Promise.all([inPromise, outPromise, zstdExited])
+  } finally {
+    if (zstd.exitCode === null && zstd.signalCode === null && !zstd.killed) {
+      zstd.kill('SIGTERM')
+    }
+  }
+}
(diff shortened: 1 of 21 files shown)


packages/cache/src/options.ts
```diff
@@ -80,6 +80,32 @@ export interface DownloadOptions {
...
+   * Parse / decompression failures encountered while validating are
+   * handled per-mode:
+   * - 'warn': logged via `core.warning()`; validation is skipped and
+   *           extraction proceeds (the system `tar` invoked for the actual
+   *           extraction is more lenient than the validator and may still
+   *           succeed).
+   * - 'error': surfaced as `CacheIntegrityError` with `code: 'PARSE_ERROR'`
+   *            and extraction does not run.
+   *
+   * @default 'off'
```
ericsciple, line 106: This feels like a security bug fix.

I'm wondering if we should default to error. We could major-version the package if we're worried about a breaking change.

jasongin: There are consumers of this package other than `actions/cache` (primarily `actions/setup-*`), so I didn't want to risk impacting them automatically. Even with a major version update, they might not realize the implications of taking the new major version. So I think it's safer to require consumers to opt in explicitly. Or at best `'warn'` could be the default with a major version update.

The linked PR https://github.com/actions/cache/pull/1761 defaults to `warn` and I proposed updating the default to `error` in a major version update of the _action_ (not the toolkit package). We should consider similar updates to `actions/setup-*` repos.

packages/cache/src/internal/pathValidation.ts
```diff
@@ -0,0 +1,522 @@
...
+}
+
+/**
+ * True if `seg` contains a glob metacharacter that isn't part of an
+ * env-var reference. Strips `${VAR}`, `$VAR`, and `%VAR%` first so the
+ * curly braces in `${VAR}` aren't misread as a brace-glob.
+ */
+function segmentHasGlob(seg: string): boolean {
+  const stripped = seg
+    .replace(/\$\{[^}]+\}/g, '')
```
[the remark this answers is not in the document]
jasongin: Not a valid concern. Added suppression with justifying comments.

packages/cache/src/internal/pathValidation.ts
```diff
@@ -0,0 +1,522 @@
...
+    .replace(/\$[A-Za-z_][A-Za-z0-9_]*/g, '')
+    .replace(/%[^%]+%/g, '')
+  return GLOB_CHAR_REGEX.test(stripped)
+}
+
+function expandEnvVars(input: string): string {
+  let result = input
+  // ${VAR}
+  result = result.replace(/\$\{([^}]+)\}/g, (_, name) => process.env[name] ?? '')
```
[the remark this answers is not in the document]
jasongin: Not a valid concern. Added suppression with justifying comments.

packages/cache/src/internal/pathValidation.ts
```diff
@@ -0,0 +1,530 @@
...
+ * input (e.g. a long run of `${` with no closing `}`). That is not a
+ * security concern here: `seg` originates from the calling workflow's
+ * own declared cache `paths:`, so a hostile value would only DoS the
+ * same workflow that supplied it — it doesn't cross a trust boundary.
+ * The `lgtm` comments below suppress CodeQL's `js/polynomial-redos`
+ * alert on that basis.
+ */
+function segmentHasGlob(seg: string): boolean {
+  const stripped = seg
+    .replace(/\$\{[^}]+\}/g, '') // lgtm[js/polynomial-redos]
```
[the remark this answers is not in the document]
jasongin: Not a valid concern. Added suppression with justifying comments.

packages/cache/src/internal/pathValidation.ts
```diff
@@ -0,0 +1,522 @@
...
+  'OldGnuLongPath',
+  'OldGnuLongLink'
+])
+
+/**
+ * Characters that signal a glob portion of a declared cache path. Anything to
+ * the right of the first segment containing one of these is stripped when
+ * deriving the longest non-glob prefix.
+ */
+const GLOB_CHAR_REGEX = /[*?[\]{}!]/
```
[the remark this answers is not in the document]
jasongin: Fixed.

packages/cache/src/internal/pathValidation.ts
```diff
@@ -0,0 +1,522 @@
...
+      reason: `path resolves outside allowed roots: ${entryPath} -> ${resolvedEntry}`
+    }
+  }
+
+  // Validate link targets for symlinks and hardlinks.
+  if (
+    linkPath !== undefined &&
+    linkPath !== '' &&
+    (entryType === 'SymbolicLink' || entryType === 'Link')
+  ) {
```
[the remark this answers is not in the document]
jasongin: Fixed.

packages/cache/src/internal/pathValidation.ts
```diff
@@ -0,0 +1,572 @@
...
+  return GLOB_CHAR_REGEX.test(stripped)
+}
+
+function expandEnvVars(input: string): string {
+  let result = input
+  // ${VAR} — see segmentHasGlob for the polynomial-redos rationale.
+  result = result.replace(
+    /\$\{([^}]+)\}/g,
+    (_, name) => process.env[name] ?? ''
+  )
```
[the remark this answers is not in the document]
jasongin: Not a valid concern. Explained in comments in the code.

jasongin:
I am considering switching to path validation using system tar, which would avoid some kinds of attacks that could bypass node-tar's path validation. However there's a compatibility challenge: BSD tar (macOS without gtar, Windows tar.exe, some self-hosted runners) lacks sufficient options to list files with quoted/escaped paths and full link targets. We could consider requiring GNU tar - it is already installed on hosted runners but would be a compatibility problem for some self-hosted runners.

packages/cache/src/options.ts
ericsciple: > Even with a major version update, they might not realize the implications of taking the new major version             
                                                                                                                        
Breaking changes is what semver communicates.                                                                           
                                                                                                                        
We should have safe defaults. Is there a legitimate scenario where an archive entry should escape the declared cache paths?

jasongin: If this was a simple validation then I might agree. But given the complexity of the implementation, I think there is some risk of edge cases with certain platforms or archive contents. So I'm proposing to leave the validation disabled by default (opt-in) until we have more data proving it is solid in many real-world scenarios.

jasongin:
I pushed a major update that implements a revised approach to path validation. The main changes are the PAX header defense and use of an allow-list generated by node-tar passed to system-tar. They are explained in the updated PR description.

packages/cache/src/options.ts
Link-: If only we had a `beta` release cycle for these, this could have been warn in the beta and error in the GA. I'm fine with `warn` and optional for now, we don't have the capacity to deal with a deluge of breaking reports, even with a major version bump.

packages/cache/src/internal/tar.ts
```diff
@@ -55,7 +67,8 @@ async function getTarArgs(
   tarPath: ArchiveTool,
   compressionMethod: CompressionMethod,
   type: string,
-  archivePath = ''
+  archivePath = '',
+  allowListPath = ''
```
Link-, line 71: Hmm.. in the backlog we had plans to implement streamed extractions like we do in Artifacts to speed up the restoration process. How is this going to play out in the streaming scenario? Are they gonna have to be mutually exclusive features?

https://github.com/actions/toolkit/blob/0786132e6a1a4451c2d392bbbf481c2b172d4312/packages/artifact/src/internal/download/download-artifact.ts#L68

packages/cache/src/internal/tar.ts
```diff
@@ -272,13 +311,172 @@ export async function listTar(
...
+  // logged and extraction proceeds.
+  if (pathValidation !== 'off') {
+    const declaredPaths = options?.declaredPaths ?? []
+    let allowedRoots = deriveAllowedRoots(declaredPaths, workingDirectory)
+    // Fail-safe: if the caller didn't supply any declared paths (or all
+    // supplied paths were empty/negations), fall back to the workspace
+    // root as the sole allowed root. This still catches archives that try
+    // to escape the workspace via `..` or absolute paths.
+    if (allowedRoots.length === 0) {
+      allowedRoots = [workingDirectory]
```
Link-, line 340: Doesn't this mean that the `workingDirectory` is the new exploitation target? Especially with this definition: `return process.env['GITHUB_WORKSPACE'] ?? process.cwd()`

packages/cache/src/internal/pax-reparse.ts
```diff
@@ -0,0 +1,297 @@
+/**
```
Link-, line 1: This is a compatibility nightmare in the making. We don't have the bandwidth to maintain this. Making sure this is compatible across ALL the platforms this library runs on is going to be very difficult to put it mildly.

packages/cache/src/internal/listAndValidate.ts
```diff
@@ -0,0 +1,341 @@
...
+ * Stream the entries of a (possibly compressed) tar archive and validate each
+ * against the allowed roots. Does NOT extract any files — entries are read
+ * for header inspection only. Returns the list of violations (empty if the
+ * archive is clean) and the names approved for extraction.
+ *
+ * Throws an Error if the archive cannot be parsed (corrupt header,
+ * decompression failure, truncated stream, etc.). The caller is responsible
+ * for translating that into a CacheIntegrityError with code `PARSE_ERROR`.
+ */
+export async function listAndValidate(
```
Link-, line 63: Did we benchmark this? What's the cost of running this?

I'd be curious about the numbers on the various runners we already offer (before thinking of others)

PR #2421, opened 2026-05-30
fix(cache): make error classification resilient to minification

fengmk2:
## Problem

`packages/cache/src/cache.ts` classifies caught errors with
`typedError.name === SomeError.name`. The error classes set `this.name` to a
string literal, but the comparison reads the **class binding's** `.name`. When
an action bundles `@actions/cache` with a name-mangling minifier (esbuild /
terser / rolldown without keep-names), the binding is renamed, so
`SomeError.name` no longer equals the literal and the check silently fails.

Most visibly, a benign `ReserveCacheError` (a reserve race when matrix jobs
share a cache key) is then logged via `core.warning` instead of `core.info`,
producing noisy warning annotations even though caching succeeded.

## Fix

Replace the `=== SomeError.name` comparisons (7 sites across `restoreCacheV1/V2`
and `saveCacheV1/V2`) with a helper:

```ts
function isErrorOfType<T extends Error>(
  error: Error,
  ctor: new (...args: never[]) => T,
  name: string
): boolean {
  return error instanceof ctor || error.name === name
}
```

`instanceof` handles the normal case and survives minification (the classes
already set their prototype via `Object.setPrototypeOf`); the `name`
string-literal fallback handles a duplicate / cross-realm copy of
`@actions/cache` where `instanceof` fails.

## Tests

Regression tests simulate a minifier by mangling an error class's runtime name,
then assert the reserve race (v1) and a finalize failure (v2) are still
classified correctly. Both fail on `main` and pass with this change.
`npm test`, `npm run lint`, and `npm run build` all pass.

### DIFF
--- a/packages/cache/__tests__/saveCache.test.ts
+++ b/packages/cache/__tests__/saveCache.test.ts
@@ -1,6 +1,6 @@
 import * as core from '@actions/core'
 import * as path from 'path'
-import {saveCache} from '../src/cache'
+import {saveCache, ReserveCacheError} from '../src/cache'
 import * as cacheHttpClient from '../src/internal/cacheHttpClient'
 import * as cacheUtils from '../src/internal/cacheUtils'
 import * as config from '../src/internal/config'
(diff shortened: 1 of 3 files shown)


joe345-str:
function isErrorOfType<T extends Error>(
  error: Error,
  ctor: new (...args: never[]) => T,
  name: string
): boolean {
  return error instanceof ctor || error.name === name
}

joe345-str:
function isErrorOfType<T extends Error>(
  error: Error,
  ctor: new (...args: never[]) => T,
  name: string
): boolean {
  return error instanceof ctor || error.name === name
}

PR #2438, opened 2026-06-19
feat(artifact): unblock v4+ on self-hosted GHES when ACTIONS_RESULTS_URL is set

### LINKED ISSUE #2439: artifact: v4+ blocked on self-hosted GHES even on releases that ship the v2 backend (3.13+)
**Describe the enhancement**

`isGhes()` in `@actions/artifact` (packages/artifact/src/internal/shared/config.ts) returns true for any `GITHUB_SERVER_URL` whose hostname is not `github.com`, `*.ghe.com`, or `*.localhost`. When `isGhes()` returns true, `client.ts` throws `GHESNotSupportedError` on every upload, download, and list call, blocking the v4+ artifact family on every self-hosted GHES instance regardless of release.

GHES 3.13 added the artifact-storage-v2 backend that v4+ clients require. 3.16+ has full production support. Despite that, the hostname-only gate continues to throw on these releases, so v4+ users on supported GHES see:

```
@actions/artifact v2.0.0+, upload-artifact@v4+ and download-artifact@v4+
are not currently supported on GHES.
```

The runtime already exposes a clean signal for whether v2 is reachable: the runner sets `ACTIONS_RESULTS_URL` exactly when the storage backend is available. The artifact client's own `getResultsServiceUrl()` (in the same file) already throws if `ACTIONS_RESULTS_URL` is unset, so it's already a precondition for v4+ paths to function.

Treating its presence as the authority on "is the v2 backend reachable" would unblock GHES 3.13+ without any new env-var contract or operator opt-in.

**Code Snippet**

Proposed change to `packages/artifact/src/internal/shared/config.ts`:

```typescript
const hostnameLooksLikeGhes =
  !isGitHubHost && !isGheHost && !isLocalHost

if (hostnameLooksLikeGhes && process.env['ACTIONS_RESULTS_URL']) {
  return false
}

return hostnameLooksLikeGhes
```

**Additional information**

This is narrower in scope than #2123, which takes a more general vendor-aware approach (an explicit `ACTIONS_VENDOR` env var that operators set to identify their environment, supporting third-party servers like Gitea and Forgejo as well as GHES). That PR addresses the broader compatibility need; this issue covers the specific case of GHES instances on releases that already ship the v2 backend, where the runtime's existing `ACTIONS_RESULTS_URL` signal lets the action detect v2 availability automatically without operator action.

The two approaches are complementary and could both land. If they do, the check proposed here would short-circuit before vendor detection on v2-supporting GHES.

camjay:
Closes #2439
Related to #2123 (more general vendor-aware approach for non-GitHub servers; this PR addresses the narrower GHES-3.13+ case via existing runtime signals)

`isGhes()` in `packages/artifact/src/internal/shared/config.ts` currently throws `GHESNotSupportedError` on any GHES, regardless of release. GHES 3.13 added the artifact-storage-v2 backend that v4+ requires, but the gate is hostname-only and ignores the release. `ACTIONS_RESULTS_URL` is the runtime signal for v2 availability; the same package's `getResultsServiceUrl()` already requires it. This change makes `isGhes()` return `false` when the hostname looks like GHES and `ACTIONS_RESULTS_URL` is set.

Behaviour matrix:

- `github.com`, `*.ghe.com`, `*.localhost`: hostname check still returns `false`. No change.
- GHES <3.13 (no v2 backend, `ACTIONS_RESULTS_URL` unset): `isGhes()` returns `true`, throws as before. No change.
- GHES 3.13+ (v2 backend present, `ACTIONS_RESULTS_URL` set): `isGhes()` returns `false`, upload/download/list proceed.

Tests in `packages/artifact/__tests__/config.test.ts` cover both new branches. Also added a `beforeEach` cleanup of `GITHUB_SERVER_URL` and `ACTIONS_RESULTS_URL` inside the `isGhes` describe block; the previous setup relied on `jest.resetModules()` alone and could leak `ACTIONS_RESULTS_URL` from sibling tests. The previously-misleadingly-titled "should return false ... is specific to an enterprise" test (which expected `true`) is renamed to match its actual assertion.

Verified on a GHES 3.19.5 instance where the v2 backend is operational but v4+ artifact actions were blocked solely by the hostname gate.

Relationship to #2123: that PR takes a more general approach with an `ACTIONS_VENDOR` env var as an explicit opt-in, supporting non-GitHub vendor servers (Gitea, Forgejo, etc.) in addition to GHES. This PR is narrower: it consults an existing runtime signal (`ACTIONS_RESULTS_URL`) that GHES 3.13+ runners already emit, so no operator action is required and no new env-var contract is introduced. The two approaches are complementary; if both land, the check here would short-circuit before vendor detection on v2-supporting GHES.

### DIFF
--- a/packages/artifact/__tests__/config.test.ts
+++ b/packages/artifact/__tests__/config.test.ts
@@ -15,6 +15,11 @@ beforeEach(() => {
 })
 
 describe('isGhes', () => {
+  beforeEach(() => {
+    delete process.env.GITHUB_SERVER_URL
+    delete process.env.ACTIONS_RESULTS_URL
+  })
+
   it('should return false when the request domain is github.com', () => {
     process.env.GITHUB_SERVER_URL = 'https://github.com'
     expect(config.isGhes()).toBe(false)
@@ -35,9 +40,21 @@ describe('isGhes', () => {
     expect(config.isGhes()).toBe(false)
   })
 
-  it('should return false when the request domain is specific to an enterprise', () => {
+  it('should return true on a self-hosted enterprise hostname when ACTIONS_RESULTS_URL is unset', () => {
     process.env.GITHUB_SERVER_URL = 'https://my-enterprise.github.com'
     expect(config.isGhes()).toBe(true)
+  })
+
+  it('should return false on a self-hosted enterprise hostname when ACTIONS_RESULTS_URL is set (v2 backend available)', () => {
+    process.env.GITHUB_SERVER_URL = 'https://my-enterprise.github.com'
+    process.env.ACTIONS_RESULTS_URL = 'https://results.example.com/'
+    expect(config.isGhes()).toBe(false)
+  })
+
+  it('should ignore ACTIONS_RESULTS_URL on github.com (already returns false there)', () => {
+    process.env.GITHUB_SERVER_URL = 'https://github.com'
+    process.env.ACTIONS_RESULTS_URL = 'https://results.example.com/'
+    expect(config.isGhes()).toBe(false)
   })
 })
 
--- a/packages/artifact/src/internal/shared/config.ts
+++ b/packages/artifact/src/internal/shared/config.ts
@@ -34,7 +34,21 @@ export function isGhes(): boolean {
   const isGheHost = hostname.endsWith('.GHE.COM')
   const isLocalHost = hostname.endsWith('.LOCALHOST')
 
-  return !isGitHubHost && !isGheHost && !isLocalHost
+  const hostnameLooksLikeGhes =
+    !isGitHubHost && !isGheHost && !isLocalHost
+
+  // GHES 3.13+ ships the artifact-storage-v2 backend that the v4+
+  // artifact client requires. The runner exposes the backend's
+  // presence via ACTIONS_RESULTS_URL (which getResultsServiceUrl()
+  // already requires for upload/download/list to function). When
+  // that env var is set we know the backend is reachable, so an
+  // unrecognised hostname does not need to short-circuit into a
+  // GHESNotSupportedError.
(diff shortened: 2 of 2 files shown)


camjay:
Hey @actions/artifacts-actions, @danwkennedy, @jasongin: bumping this on the back of the artifact README's own note that v4 is "not currently supported on GHES yet."

GHES 3.13+ ships the v2 artifact backend; on those releases the runner already exports `ACTIONS_RESULTS_URL`. The current `isGhes()` check in `packages/artifact/src/internal/shared/config.ts` is purely hostname-based, so it short-circuits before that signal is consulted and v4 stays blocked even where the backend is fully present. This PR is a one-line gate change plus tests: keep the hostname heuristic, but skip blocking when `ACTIONS_RESULTS_URL` is set, which is the same signal the upload/download flow uses downstream.

Related: #2439 (issue) and the older #2123 (different approach to the same problem, open since early 2026).

Happy to adjust scope/tests/style. Just looking for a triage decision so we know whether to keep iterating here or pursue an ADR per CONTRIBUTING.md.

philip-gai:
@camjay Thanks for the PR! The v2 (Results) backend that `@actions/artifact` v4+ requires isn't available on GHES, so this should stay blocked there.

> GHES 3.13 added the artifact-storage-v2 backend that v4+ requires

This isn't the case, it's still unsupported on GHES (the README calls this out too). GHES runners also don't set `ACTIONS_RESULTS_URL`, so the check here wouldn't fire on a real GHES instance anyway, and if it did get set for some other reason, it'd swap the clear `GHESNotSupportedError` for confusing downstream failures.

Let me know if I'm misunderstanding something! 😄

kbsoso9999-maker (approved):

PR #2447, opened 2026-07-01
feat(cache): add cache-mode client behavior (read-denied warning + ACTIONS_CACHE_MODE skip)

philip-gai:
## Description

This PR adds client-side cache-mode behavior to `@actions/cache`. It combines two related pieces of work:

- Surface the service-side read-denied policy as a non-fatal warning.
- Honor a new `ACTIONS_CACHE_MODE` environment variable to skip cache operations the effective mode does not permit.

Cache-mode values are `none | read | write | write-only` (hyphenated). It is a partial lattice, not linear: readable modes = {read, write}, writable modes = {write, write-only}, none = neither. The `ACTIONS_CACHE_MODE` env var is provided by the Actions runner, and is gated on the service side. When it is unset or unrecognized, behavior is identical to today (regression-safe).

## Read-denied warning

When the cache service denies a download because the token has no readable scopes, the run should continue with a clear warning instead of a confusing failure.

- Added `CacheReadDeniedError` and the shared `cache read denied:` prefix constant.
- On the v2 restore path, the denial arrives as a wrapped 403; the restore dispatch detects the prefix and surfaces it as a `core.warning`, then returns a cache miss.
- Matches the existing write-denied handling (`cache write denied:`) for consistency.

## ACTIONS_CACHE_MODE skip

Skip operations the effective cache-mode does not permit, before any tar or network work.

- Skip restore when the mode is not readable (`none`, `write-only`).
- Skip save when the mode is not writable (`none`, `read`).
- Restore skip returns a cache miss (`undefined`); save skip returns the existing not-saved sentinel (`-1`). Neither throws.
- Exactly one `core.info` line per skip; extra detail (paths, key) is behind `ACTIONS_STEP_DEBUG`.
- Unset or unrecognized modes are permissive, so behavior is unchanged.

Rationale for skipping restore on `write-only`: write-only tokens have no read scope, so a restore would be denied service-side and trip the read-denied warning. Skipping client-side avoids the wasted round-trip and gives a clearer message.

## Companion changes

- actions/runner exports `ACTIONS_CACHE_MODE` into the step environment.

## Notes

- `packages/cache/RELEASES.md` updated under the 6.2.0 entry.
- Test coverage: full cache suite passing, including the read-denied dispatch and a cache-mode truth table (none/read/write/write-only plus unset and unknown regression cases).

https://github.com/github/actions-persistence/issues/1168

### DIFF
--- a/packages/cache/__tests__/config.test.ts
+++ b/packages/cache/__tests__/config.test.ts
@@ -23,3 +23,37 @@ test('isGhes returns false for ghe.localhost', () => {
   process.env.GITHUB_SERVER_URL = 'https://my.domain.ghe.localhost'
   expect(config.isGhes()).toBe(false)
 })
+
+describe('cache-mode helpers', () => {
+  const original = process.env.ACTIONS_CACHE_MODE
+
+  afterEach(() => {
+    if (original === undefined) {
+      delete process.env.ACTIONS_CACHE_MODE
+    } else {
+      process.env.ACTIONS_CACHE_MODE = original
+    }
+  })
+
+  test('getCacheMode normalizes whitespace and case', () => {
+    process.env.ACTIONS_CACHE_MODE = '  Write-Only  '
+    expect(config.getCacheMode()).toBe('write-only')
+  })
+
+  test('getCacheMode returns empty string when unset', () => {
+    delete process.env.ACTIONS_CACHE_MODE
+    expect(config.getCacheMode()).toBe('')
+  })
+
+  test.each([
+    ['', true, true],
+    ['read', true, false],
+    ['write', true, true],
+    ['write-only', false, true],
+    ['none', false, false],
+    ['garbage', true, true]
+  ])("mode '%s' -> readable=%s writable=%s", (mode, readable, writable) => {
+    expect(config.isCacheReadable(mode)).toBe(readable)
+    expect(config.isCacheWritable(mode)).toBe(writable)
+  })
+})
--- a/packages/cache/__tests__/restoreCache.test.ts
+++ b/packages/cache/__tests__/restoreCache.test.ts
@@ -146,6 +146,57 @@ test('restore surfaces a non-read-denied getCacheEntry error as a normal warning
   expect(logWarningMock).toHaveBeenCalledTimes(1)
 })
 
+describe('restore cache-mode gating', () => {
+  const original = process.env.ACTIONS_CACHE_MODE
+
+  afterEach(() => {
+    if (original === undefined) {
+      delete process.env.ACTIONS_CACHE_MODE
+    } else {
+      process.env.ACTIONS_CACHE_MODE = original
+    }
+  })
+
+  test.each(['none', 'write-only'])(
+    "mode '%s' skips restore without touching the cache service",
+    async mode => {
+      process.env.ACTIONS_CACHE_MODE = mode
+      const logInfoMock = jest.spyOn(core, 'info')
+      const getCacheEntryMock = jest.spyOn(cacheHttpClient, 'getCacheEntry')
+
+      const cacheKey = await restoreCache(['node_modules'], 'node-test')
+
+      expect(cacheKey).toBe(undefined)
+      expect(getCacheEntryMock).not.toHaveBeenCalled()
+      expect(logInfoMock).toHaveBeenCalledWith(
+        `Cache restore skipped: the effective cache-mode '${mode}' does not permit reads.`
+      )
+    }
+  )
+
+  test.each(['read', 'write', ''])(
+    "mode '%s' does not skip restore",
+    async mode => {
+      if (mode === '') {
+        delete process.env.ACTIONS_CACHE_MODE
+      } else {
+        process.env.ACTIONS_CACHE_MODE = mode
+      }
+      const logInfoMock = jest.spyOn(core, 'info')
+      const getCacheEntryMock = jest
+        .spyOn(cacheHttpClient, 'getCacheEntry')
+        .mockResolvedValue(null as never)
+
+      await restoreCache(['node_modules'], 'node-test')
+
+      expect(getCacheEntryMock).toHaveBeenCalledTimes(1)
+      expect(logInfoMock).not.toHaveBeenCalledWith(
+        expect.stringContaining('Cache restore skipped')
+      )
+    }
+  )
+})
+
 test('restore with restore keys and no cache found', async () => {
   const paths = ['node_modules']
   const key = 'node-test'
--- a/packages/cache/__tests__/restoreCache.test.ts
+++ b/packages/cache/__tests__/restoreCache.test.ts
@@ -123,6 +123,29 @@ test('restore denied by read-only token logs warning and reports cache miss', as
   expect(logWarningMock).toHaveBeenCalledTimes(1)
 })
 
+test('restore surfaces a non-read-denied getCacheEntry error as a normal warning', async () => {
+  // Guards the inner catch in restoreCacheV1: only `cache read denied:` errors
+  // are re-classified; every other getCacheEntry failure must pass through and
+  // be logged normally (not swallowed or mislabeled as a read denial).
+  const paths = ['node_modules']
+  const key = 'node-test'
+  const logErrorMock = jest.spyOn(core, 'error')
+  const logWarningMock = jest.spyOn(core, 'warning')
+  const genericMessage = 'Cache service responded with 400'
+
+  jest.spyOn(cacheHttpClient, 'getCacheEntry').mockImplementation(async () => {
+    throw new Error(genericMessage)
+  })
+
+  const cacheKey = await restoreCache(paths, key)
+  expect(cacheKey).toBe(undefined)
+  expect(logErrorMock).not.toHaveBeenCalled()
+  expect(logWarningMock).toHaveBeenCalledWith(
+    `Failed to restore: ${genericMessage}`
+  )
+  expect(logWarningMock).toHaveBeenCalledTimes(1)
+})
+
 test('restore with restore keys and no cache found', async () => {
   const paths = ['node_modules']
   const key = 'node-test'
--- a/packages/cache/__tests__/restoreCache.test.ts
+++ b/packages/cache/__tests__/restoreCache.test.ts
@@ -99,50 +99,24 @@ test('restore with server error should fail', async () => {
   delete process.env['ACTIONS_RESULTS_URL']
 })
 
-test('restore denied by read-only token logs warning and reports cache miss', async () => {
-  // The GHES v1 artifact cache service returns HTTP 403 with a
-  // `cache read denied:` body when the token has no readable scopes.
-  // getCacheEntry surfaces that message; expect a single warning (not error)
-  // and a cache miss.
+test('restore surfaces a getCacheEntry failure as a warning and reports a cache miss', async () => {
+  // restoreCache treats any getCacheEntry failure (read-denied or otherwise)
+  // as a non-fatal warning and a cache miss so the workflow continues. The
+  // read-denied prefix detection itself is covered in cacheHttpClient.test.ts.
   const paths = ['node_modules']
   const key = 'node-test'
   const logErrorMock = jest.spyOn(core, 'error')
   const logWarningMock = jest.spyOn(core, 'warning')
-  const deniedMessage = 'cache read denied: token has no readable scopes'
+  const message = 'cache read denied: token has no readable scopes'
 
   jest.spyOn(cacheHttpClient, 'getCacheEntry').mockImplementation(async () => {
-    throw new Error(deniedMessage)
+    throw new Error(message)
   })
 
   const cacheKey = await restoreCache(paths, key)
   expect(cacheKey).toBe(undefined)
   expect(logErrorMock).not.toHaveBeenCalled()
-  expect(logWarningMock).toHaveBeenCalledWith(
-    `Failed to restore: ${deniedMessage}`
-  )
-  expect(logWarningMock).toHaveBeenCalledTimes(1)
-})
-
-test('restore surfaces a non-read-denied getCacheEntry error as a normal warning', async () => {
-  // Guards the inner catch in restoreCacheV1: only `cache read denied:` errors
-  // are re-classified; every other getCacheEntry failure must pass through and
-  // be logged normally (not swallowed or mislabeled as a read denial).
-  const paths = ['node_modules']
-  const key = 'node-test'
-  const logErrorMock = jest.spyOn(core, 'error')
-  const logWarningMock = jest.spyOn(core, 'warning')
-  const genericMessage = 'Cache service responded with 400'
-
-  jest.spyOn(cacheHttpClient, 'getCacheEntry').mockImplementation(async () => {
-    throw new Error(genericMessage)
-  })
-
-  const cacheKey = await restoreCache(paths, key)
-  expect(cacheKey).toBe(undefined)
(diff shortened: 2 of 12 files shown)


packages/cache/src/cache.ts
```diff
@@ -62,6 +62,23 @@ export class CacheWriteDeniedError extends ReserveCacheError {
   }
 }
 
+// Prefix the receiver embeds in a cache read denial: the v2 twirp
+// GetCacheEntryDownloadURL error, or the GHES v1 `_apis/artifactcache` 403
+// body. Match with `includes`, not `startsWith`: the message is wrapped by
+// the transport, so the prefix ends up embedded rather than leading.
+export const CACHE_READ_DENIED_PREFIX = 'cache read denied:'
```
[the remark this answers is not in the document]
philip-gai: Good catch. Moved the prefix to the dependency-free `src/internal/constants.ts` as `CacheReadDeniedMessagePrefix`, imported it in both `cache.ts` and `cacheHttpClient.ts`, and re-exported it from `cache.ts` as `CACHE_READ_DENIED_PREFIX` so consumers are unaffected. No more duplicated literal. Fixed in b504f4ff.

packages/cache/src/internal/config.ts
```diff
@@ -19,6 +19,26 @@ export function getCacheServiceVersion(): string {
   return process.env['ACTIONS_CACHE_SERVICE_V2'] ? 'v2' : 'v1'
 }
 
+// The cache-mode lattice (ADR c2c-actions#10264): readable = {read, write},
```
jasongin: nit: avoid references to internal issues and documents in public source code.

jasongin (dismissed):

packages/cache/src/internal/config.ts
philip-gai: Oops good call, pushed an update removing that.

jasongin (dismissed):

packages/cache/__tests__/config.test.ts
```diff
@@ -23,3 +23,37 @@ test('isGhes returns false for ghe.localhost', () => {
...
+  test('getCacheMode returns empty string when unset', () => {
+    delete process.env.ACTIONS_CACHE_MODE
+    expect(config.getCacheMode()).toBe('')
+  })
+
+  test.each([
+    ['', true, true],
+    ['read', true, false],
+    ['write', true, true],
+    ['write-only', false, true],
```
boxofyellow, line 52: the write-only seems very strange.

Normally the workflow is like this

1. Try to get the cache
   1. If can't go do stuff locally to replace what would be in the cache
1. Go do stuff with the value in the cache
1. If you got a cache miss try and save the locally created cache

So if you set this to read only, you will always get a cache miss, so you will alway try to write the cache at the end.

So besides being silly, I bet we will return 409 - conflict when you try to overwrite the cache that you were not allowed to read

packages/cache/__tests__/config.test.ts
```diff
@@ -23,3 +23,37 @@ test('isGhes returns false for ghe.localhost', () => {
...
+    ['none', false, false],
+    ['garbage', true, true]
```
boxofyellow, line 54: Do we really want this case to effectively "fail open", it seems like if put something there and we can't tell what it is we should not assume your wanted write (after all if you wanted write, you could just leave it blank).

packages/cache/__tests__/restoreCache.test.ts
```diff
@@ -99,6 +99,132 @@ test('restore with server error should fail', async () => {
...
+  const cacheKey = await restoreCache(paths, key)
+  expect(cacheKey).toBe(undefined)
+  expect(logErrorMock).not.toHaveBeenCalled()
+  expect(logWarningMock).toHaveBeenCalledWith(
+    `Failed to restore: ${deniedMessage}`
+  )
+  expect(logWarningMock).toHaveBeenCalledTimes(1)
+})
+
+test('restore surfaces a non-read-denied getCacheEntry error as a normal warning', async () => {
```
boxofyellow: How is this test functionally any different than `'restore denied by read-only token logs warning and reports cache miss'`?  It basically is setting up the same mocks and asserting the same things.

packages/cache/__tests__/restoreCache.test.ts
```diff
@@ -99,6 +99,132 @@ test('restore with server error should fail', async () => {
...
+      process.env.ACTIONS_CACHE_MODE = mode
+      restoreEnv('ACTIONS_CACHE_SERVICE_V2', v2)
+      const getCacheEntryMock = jest.spyOn(cacheHttpClient, 'getCacheEntry')
+
+      const cacheKey = await restoreCache(['node_modules'], 'node-test')
+
+      expect(cacheKey).toBe(undefined)
+      expect(getCacheEntryMock).not.toHaveBeenCalled()
+    }
+  )
```
boxofyellow: It seems like we don't need both test test


```suggestion
  // The skip short-circuits before v1/v2 dispatch, so it applies regardless of
  // the ACTIONS_CACHE_SERVICE_V2 feature flag.
  test.each([
    ['none', undefined],
    ['none', 'true'],
    ['write-only', undefined],
    ['write-only', 'true']
  ])(
    "mode '%s' skips restore with ACTIONS_CACHE_SERVICE_V2=%s",
    async (mode, v2) => {
      process.env.ACTIONS_CACHE_MODE = mode
      restoreEnv('ACTIONS_CACHE_SERVICE_V2', v2)
      const logInfoMock = jest.spyOn(core, 'info')
      const getCacheEntryMock = jest.spyOn(cacheHttpClient, 'getCacheEntry')

      const cacheKey = await restoreCache(['node_modules'], 'node-test')

      expect(cacheKey).toBe(undefined)
      expect(getCacheEntryMock).not.toHaveBeenCalled()
      expect(logInfoMock).toHaveBeenCalledTimes(1)
      expect(logInfoMock).toHaveBeenCalledWith(
        `Cache restore skipped: the effective cache-mode '${mode}' does not permit reads.`
      )
    }
  )
```

packages/cache/__tests__/config.test.ts
boxofyellow: One sec... in this case we are going to ask the service and if it does not like `garbage` (which it should not) then it will block them.

If that is the case, I'm on board here, if the client get something it does not know what the server will do with it then it should throw it at the server to see what sticks.

packages/cache/__tests__/saveCache.test.ts
```diff
@@ -45,6 +57,88 @@ test('save with missing input should fail', async () => {
...
+      process.env.ACTIONS_CACHE_MODE = mode
+      restoreEnv('ACTIONS_CACHE_SERVICE_V2', v2)
+      const resolvePathsMock = jest.spyOn(cacheUtils, 'resolvePaths')
+
+      const cacheId = await saveCache(['node_modules'], 'node-test')
+
+      expect(cacheId).toBe(-1)
+      expect(resolvePathsMock).not.toHaveBeenCalled()
+    }
+  )
```
boxofyellow: just like the others these look combinable


```suggestion
  // The skip short-circuits before v1/v2 dispatch, so it applies regardless of
  // the ACTIONS_CACHE_SERVICE_V2 feature flag.
  test.each([
    ['read', undefined],
    ['read', 'true'],
    ['none', undefined],
    ['none', 'true']
  ])(
    "mode '%s' skips save with ACTIONS_CACHE_SERVICE_V2=%s",
    async (mode, v2) => {
      process.env.ACTIONS_CACHE_MODE = mode
      restoreEnv('ACTIONS_CACHE_SERVICE_V2', v2)
      const logInfoMock = jest.spyOn(core, 'info')
      const resolvePathsMock = jest.spyOn(cacheUtils, 'resolvePaths')

      const cacheId = await saveCache(['node_modules'], 'node-test')

      expect(cacheId).toBe(-1)
      expect(resolvePathsMock).not.toHaveBeenCalled()
      expect(logInfoMock).toHaveBeenCalledTimes(1)
      expect(logInfoMock).toHaveBeenCalledWith(
        `Cache save skipped: the effective cache-mode '${mode}' does not permit writes.`
      )
    }
  )
```

packages/cache/src/internal/cacheHttpClient.ts
```diff
@@ -101,6 +102,12 @@ export async function getCacheEntry(
     return null
   }
   if (!isSuccessStatusCode(response.statusCode)) {
+    // Only surface the receiver's body for a `cache read denied:` policy denial
+    // so callers can dispatch on it; keep the generic message otherwise.
+    const errorMessage = response.error?.message
```
boxofyellow, line 107: it seems odd to me that only have this "fix up the error" message info in the read path.  It seams like the write path benefit from it.  And a write read only setup seems far more helpful than a write only one in my eyes.

packages/cache/src/cache.ts
```diff
@@ -191,10 +224,23 @@ async function restoreCacheV1(
...
-      enableCrossOsArchive
-    })
+    let cacheEntry
+    try {
+      cacheEntry = await cacheHttpClient.getCacheEntry(keys, paths, {
+        compressionMethod,
+        enableCrossOsArchive
+      })
+    } catch (error) {
+      // The GHES v1 artifact cache service returns HTTP 403 with a
```
boxofyellow: Why is the error fixing up logic here and in `getCacheEntry`?

packages/cache/src/cache.ts
```diff
@@ -237,6 +283,10 @@ async function restoreCacheV1(
     const typedError = error as Error
     if (typedError.name === ValidationError.name) {
       throw error
+    } else if (typedError.name === CacheReadDeniedError.name) {
+      // Read denied by policy (token has no readable cache scopes). Warn and
+      // treat as a cache miss so the workflow continues.
+      core.warning(`Failed to restore: ${typedError.message}`)
```
boxofyellow: I'm ok being explicit here but I just want to make sure I understand.

It does not look like this `else if` is strictly needed, we could just let it fall in the `else` block and since it won't look like 5xx error, it would still just warn.

packages/cache/src/cache.ts
```diff
@@ -314,7 +364,19 @@ async function restoreCacheV2(
...
+      response = await twirpClient.GetCacheEntryDownloadURL(request)
+    } catch (error) {
+      // The receiver returns twirp PermissionDenied (403) when the run's token
+      // has no readable cache scopes. The client wraps that 403, so the stable
+      // prefix is embedded in the message rather than leading it.
+      const errorMessage = (error as Error)?.message ?? ''
+      if (errorMessage.includes(CACHE_READ_DENIED_PREFIX)) {
+        throw new CacheReadDeniedError(errorMessage)
+      }
+      throw error
```
boxofyellow, line 379: Right below we don't throw on a miss, we return `undefined` when we can't find it the cache, the equivalent to a 404.  Do we want to be throwing on 403?

packages/cache/__tests__/restoreCache.test.ts
philip-gai: This makes sure that other generic errors are not reclassified as denials to avoid regressions

packages/cache/__tests__/config.test.ts
philip-gai: Yeah and we don't want to break current clients by treating unset `cache-mode` as `none`, it needs to keep `write` (which is what it is today) so we don't cause a big regression.

packages/cache/__tests__/config.test.ts
philip-gai: A lot of customers want write-only so that they can have workflows that cannot read from a poisoned cache but can write fresh entries into a cache on `main`

packages/cache/src/internal/cacheHttpClient.ts
philip-gai: We already have a block for `CacheWriteDeniedError`, this is just adding it for the read path whenever the scope is `write-only` or `none` and reads are denied

packages/cache/__tests__/config.test.ts
boxofyellow: I see, they can be locked down, and in turn can't be poisoned them self, which prevents them from spreading the poison.

packages/cache/__tests__/config.test.ts
boxofyellow: Yeah it took me minute to get there, I'm on board.

packages/cache/__tests__/restoreCache.test.ts
boxofyellow: > This makes sure that other generic errors are not reclassified as denials to avoid regressions

I don't think that is true 🙃, as the two tests are structurally identical

They both do this

1. set `path` (to the same value)
2. set `key` (to the same value)
3. set up a mock for the error log
4. set up a mock for the warning log
5. set message (one calls the variable `deniedMessage` and other calls it `genericMessage`, the content of the message is different but interchangeable)
6. set up a mock on `getCacheEntry` that throws the previosusly setup message.
7. call `restoreCache`
8. assert the result is `undefined`
9. assert nothing was written to the error log
10. assert `Failed to restore:` followed by the setup message was written as a warning
11. assert that only one thing was written as a warning.

So the only change is that step 5, and I believe you could replace any text, and get the same result.

packages/cache/src/internal/cacheHttpClient.ts
boxofyellow: I think some of my confusion that we have the error fixing up code in both `packages/cache/src/internal/cacheHttpClient.ts` and `packages/cache/src/cache.ts` for the read case, while for write we only have it in `packages/cache/src/cache.ts` (it also does not seen to call out that likely your scope is blocking you.  I don't know if there are other cases in the write path that could yield a 403)

packages/cache/__tests__/saveCache.test.ts
philip-gai: Combined

packages/cache/src/cache.ts
philip-gai: Yeah this is redundant but I was originally following what we had for write already in place too. But now I removed them both to keep this less redundant 😄

packages/cache/__tests__/restoreCache.test.ts
philip-gai: ah true, consolidated

packages/cache/src/cache.ts
philip-gai: Yeah, this will just route as to the outer catch the same way it did before I added a catch here, and we want to be able to throw our `CacheReadDeniedError` and rethrow other errors we may get

packages/cache/src/internal/cacheHttpClient.ts
philip-gai: Oh, the write path's `reserveCache` method already returns a response of `statusCode` + `error` to clients, but read's `getCacheEntry` only throws a plain string `Error`. I didn't want clients to have to update to a new response structure, so we just throw with the `cache read denied:` prefix and `cache.ts` then inspects it and throws it as a `CacheReadDeniedError`

And we do call it out, with the message `cache read denied: token has no readable scopes`

packages/cache/__tests__/restoreCache.test.ts
philip-gai: Fixed!

packages/cache/src/cache.ts
philip-gai: I _think_ I responded to this in https://github.com/actions/toolkit/pull/2447#discussion_r3561930583 😄

boxofyellow (approved):
Looks good to me.

PR #2435, opened 2026-06-17
Backport: Handle cache write error due to RO token

jasongin:
- Backport #2429 to v5 so that it can be consumed by downstream packages that have not yet migrated to ESM.
 - Bump `@actions/cache` package version from `5.0.1` to `5.1.0` and update `RELEASES.md`.

PR #2451, opened 2026-07-13
feat: backport cache read-denied + ACTIONS_CACHE_MODE handling (5.2.0)

philip-gai:
Backports the cache read-denied handling and `ACTIONS_CACHE_MODE` gating from the ESM v6.2.0 line (#2447) to the CommonJS v5 line, to ship as `@actions/cache` 5.2.0. This mirrors the earlier write-denied backport (#2435, shipped as 5.1.0) and targets the `cache-v5.2.0-release` branch.

## Changes

- Detect the `cache read denied:` prefix on cache download failures on both the v2 twirp path (`GetCacheEntryDownloadURL`) and the v1 `_apis/artifactcache` path, and surface it as a `core.warning` without failing the run (treated as a cache miss).
- Add `CacheReadDeniedError` and re-export `CACHE_READ_DENIED_PREFIX` from the shared `CacheReadDeniedMessagePrefix` constant. Read-denied matching uses `includes(prefix)` (the twirp 403 wraps the prefix mid-message); write-denied continues to use `startsWith`.
- Honor the `ACTIONS_CACHE_MODE` environment variable: skip restore when the effective cache-mode does not permit reads (`none`, `write-only`) and skip save when it does not permit writes (`none`, `read`), logging a single non-fatal `core.info` line. Unset or unrecognized modes are unchanged.
- Add `getCacheMode()`, `isCacheReadable()`, and `isCacheWritable()` helpers to `config.ts`.
- Surface the receiver's body message from `getCacheEntry` (v1 path) only for read-denied policy denials, keeping the generic status-code error otherwise.
- Mirror the 6.2.0 read-denied and cache-mode tests across the cache test suites.
- Bump `@actions/cache` to `5.2.0` and add the corresponding `RELEASES.md` entry.

## Notes

- Logic is identical to the v6.2.0 implementation; the only adaptation is CommonJS import specifiers (no ESM `.js` extensions), since the v5 line predates the ESM refactor.
- Scoped entirely to `packages/cache`.

## Validation

- `tsc`: clean
- `jest`: 150/150 passing
- `prettier --check`: clean
- `eslint`: clean

### DIFF
--- a/packages/cache/RELEASES.md
+++ b/packages/cache/RELEASES.md
@@ -1,5 +1,10 @@
 # @actions/cache Releases
 
+### 5.2.0
+
+- Handle cache read error due to read-only token: detect the `cache read denied:` prefix on cache download failures (both the v2 twirp path and the v1 `_apis/artifactcache` path) and surface it as a `core.warning` (without failing the run).
+- Honor the `ACTIONS_CACHE_MODE` environment variable: skip restore when the effective cache-mode does not permit reads (`none`, `write-only`) and skip save when it does not permit writes (`none`, `read`), logging a single non-fatal `core.info` line. When `ACTIONS_CACHE_MODE` is unset or unrecognized, behavior is unchanged.
+
 ### 5.1.0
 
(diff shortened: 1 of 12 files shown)


boxofyellow (approved):
LGTM: :shipit:

Issue #427, opened 2020-04-23
downloadTool should accept and verify an optional checksum


hashtagchris:
This could be used to verify tool downloads aren't altered or corrupt.

PR #2476, opened 2026-08-06
feat(tool-cache): optionally specify expected checksum

raboof:
Implements #427

A bit ugly with the `undefined` at the call site, though otoh having the download and the verification in the same invocation is somewhat satisfying.


raboof:
or should we refactor the signature to take an object? LMK what you prefer.

